Issue #2, opened 2019-08-21
ethereumAPI looking for feedback


terencechain:
One of the most important pieces of eth2 in phase 0 is access to data. We have collected service schema definitions that we find valuable to validator client operators, block explorers, and anyone interested in the data of eth2. We collected data API use cases in this [products requirement document](https://docs.google.com/document/d/1LmlF0meuzx7LN8fSkuZWtGSZy1K0ZgyTIwL0zdR8Or0/edit#heading=h.4d4hnezaletr) and received feedback from Danny, Jannik, Zak and many others.

We turned that feedback into a self-describing schema utilizing protocol buffers with the hope that we can start eth 2.0 with a consistent and easy to use API. The schema is defined in this repo: https://github.com/prysmaticlabs/ethereumapis

I'd love to hear more feedback on the products requirement document and the schema repo. I'd like to help migrate ethereumapis repo to here

Issue #1, opened 2019-08-21
Swagger or MD?


paulhauner:
**TL;DR: Should we use loosely-defined markdown instead of Swagger?**

We've been working to bring our [Beacon Node REST API](https://app.swaggerhub.com/apis/spble/lighthouse_rest_api/0.1.0) up-to-date to submit here, however I get the feeling the Swagger format isn't right for a `eth2.0-xxxx` repo. It seems to me that this repo should contain easily-maintainable markdown docs.

Before we go and submit more Swagger YAML to this repo, I wanted to reach out and get feedback about **should this repo use Swagger YAML or markdown to outline eth2.0 APIs?**

I've listed some pros and cons for Swagger vs Markdown below. Keen to hear your thoughts.

## Option 1. Swagger

Example: we have the [Minimal Beacon Node API for Validator](https://app.swaggerhub.com/apis/spble/beacon_node_api_for_validator/0.2.0) on SwaggerHub.

### Pros

- Super well-defined.
- Has a web interface with examples, etc.
- Does code generation.
- Additional features, etc..

#### Maintenance

Working in Swagger YAML is not necessarily ergonomic. Consider this `200` response:

```yaml
'200':
          description: Success response
          content:
            application/json:
              schema:
                type: object
                properties:
                  epoch:
                    type: integer
                    format: uint64
                    description: >-
                      The epoch for which the returned active validator changes
                      are provided.
                  balances:
                    type: array
                    items:
                      title: ValidatorBalances
                      type: object
                      properties:
                        pubkey:
                          $ref: '#/components/schemas/pubkey'
                        index:
                          type: integer
                          format: uint64
                          description: The global ValidatorIndex of the validator.
                        balance:
                          type: integer
                          format: uint64
                          description: >-
                            The balance of the validator at the specified epoch,
                            expressed in Gwei
```

I suspect maintaining this will become frustrating, especially if it's a collaborative document.

You can modify this YAML file in your favourite IDE, however you need to copy-paste it into SwaggerHub to compile it and render the UI (AFAIK).

#### Readability/hosting/CI

The YAML file itself is not accessible to developers who want to figure out how to consume the API, in the same way that markdown is auto-rendered by Github. In order to view the API, they need to either:

- Find a URL that hosts an interactive HTML version (e.g., SwaggerHub)
- Copy-paste the yaml file into SwaggerHub themselves

[...]

PR #3, opened 2019-08-22
Basic Beacon API

paulhauner:
## What

Adds a markdown document describing a basic beacon API that provides block, state, head and network information.

## Why

In the short term, clients need to interoperate and it would be helpful if they implemented standard, human-readable APIs so we don't need to go scrounging through logs to find simple things. Also, so people can build tools that can assess our interoperability foo.

In the long term, it'd be great if clients standardize upon API functionality to help make a unified developer ecosystem. Maybe this document can be the basis for this standardization effort.

## Details

What this is:

- Testing the waters to try and find a good format that is:
  - Easily maintainable (i.e., easy to read and modify the source).
  - Not going to require lots of work whenever the spec data structures change.
  - Well defined enough to allow someone to build a HTTP-wrapper library from it.
  - Not completely HTTP-centric; sets a standard that can be used for other transports (e.g., gRPC).
  - Acknowledges that HTTP seems to be the preferred API among clients.
  - Provides easy, `curl`-able examples.
  - Small enough to go from nothing to something with minimal contention.

What this is not:

- Optimized for data transmission (e.g., if you want the finalized checkpoint you need to download the whole state).
- The full suite of useful API endpoints (e.g., there's no shuffling endpoints).

### Use cases

Using this API, we can perform these tasks:

- `$ curl /network/enr` and copy-paste it into a `--bootnodes` CLI flag on another node.
- `$ curl /network/peer_id` to get the node's libp2p PeerId.
- `$ curl /network/peers` to see if a node is peered with some other node (comparing peer ids).
- Check that some nodes agree on the same chain at some slot using  `/beacon/block?slot=42` and comparing the returned `root` fields.
- Check that two nodes share the same finalized root by using `/beacon/head` to learn the state root at the head of the chain, then use `/beacon/state` to read the finalized checkpoints.

Additionally, I hope this API can be used during network testing/analysis with tools that detect network connectivity (connected peers, etc) and canonical chain info (head, blocks at slots, etc).

### Notes

- This API does not allow for requesting multiple blocks or states. I left it out for simplicity, thinking we can add it later. I could be persuaded to add it now, if it's widely desired.
- This API allows a useful method we've implemented in Lighthouse, where you "bootstrap" a node by pointing it at another node's HTTP interface and download the genesis state and peer connection info.

### DIFF
--- a/apis/beacon/basic.md
+++ b/apis/beacon/basic.md
@@ -28,14 +28,14 @@ port is not part of the specification.
 
 Table of endpoints:
 
-| Title | HTTP Path |
-| --- | --- |
-[Beacon Head](#beacon-head) | `/beacon/head`
-[Beacon Block](#beacon-block) | `/beacon/block`
-[Beacon State](#beacon-state) | `/beacon/state`
-[Network Peer Id](#network-peerid) | `/network/peer_id`
-[Network Peers](#network-peers) | `/network/peers`
-[Network ENR Address](#network-enr-address) | `/network/enr`
+| Title | HTTP Path | Description |
+| --- | --- | -- |
+[Beacon Head](#beacon-head) | `/beacon/head` | Get the head slot, block root and state root of the canonical chain.
+[Beacon Block](#beacon-block) | `/beacon/block` | Get a `BeaconBlock` by slot or root.
+[Beacon State](#beacon-state) | `/beacon/state` | Get a `BeaconState` by slot or root
+[Network Peer Id](#network-peerid) | `/network/peer_id` | Get a node's libp2p `PeerId`.
+[Network Peers](#network-peers) | `/network/peers` | List a node's libp2p peers (as `PeerIds`).
+[Network ENR Address](#network-enr-address) | `/network/enr` | Get a node's discovery `ENR` address.
 
 ## Beacon Head
 
@@ -65,7 +65,7 @@ Typical Responses | 200
 #### HTTP Example
 
 ```
-$ curl "localhost:5051/node/head"
+$ curl "localhost:5051/beacon/head"
 {"slot":546,"block_root":"0xee0973130905bfdf1deeed88ac0c09623c2cc071b03db68097f2e3b496258c17","state_root":"0xc87889ad8c760b1ec14746271372f9fb53d9f16463dcd224f86a6207799ad702"}
 ```
 
@@ -175,7 +175,7 @@ Typical Responses | 200, 404
 #### HTTP Example
 
 ```
-$ curl localhost:5051/beacon/state?slot=0
+$ curl "localhost:5051/beacon/state?slot=0"
 {"root":"0x3c06d45011320bc5a340811898e27c427547ee79e1e0f29892b4c5235d0c8c8e","beacon_state":{"genesis_time":1566448200,"slot":0,
 ```
 
--- a/apis/beacon/basic.md
+++ b/apis/beacon/basic.md
@@ -1,0 +1,266 @@
+# Beacon Node Basic API
+
+## Introduction
+
+This document provides a list of abstract API definitions, each accompanied by
+a specification of that endpoint when served via HTTP.
+
+Clients may adhere to this specification by:
+
+- Implementing the HTTP specifications exactly (or a implementing a superset).
+- Using their own transport with respect to the abstract definitions (e.g.
+	gRPC, JSON RPC).
+  - Optionally, basing their resource access paths upon the
+	HTTP paths.
+
+### Notes
+
+- The examples use `localhost:5051` purely for example purposes. The HTTP API
+port is not part of the specification.
+- This document seeks to be loosely coupled to the Eth2 spec, reducing the
+	overhead in maintaining this document as the specification progresses.
+  - Eth2 types (e.g., `BeaconBlock`) are loosely defined
+	and their version (e.g., `v0.8.3`) is open to interpretation.
+  - The JSON encoding of these types is analogous to the YAML serialization
+	  format (each endpoint should include a JSON-encoded example).
+
+## Endpoints
+
+Table of endpoints:
+
+| Title | HTTP Path |
+| --- | --- |
+[Beacon Head](#beacon-head) | `/beacon/head`
+[Beacon Block](#beacon-block) | `/beacon/block`
+[Beacon State](#beacon-state) | `/beacon/state`
+[Network Peer Id](#network-peerid) | `/network/peer_id`
+[Network Peers](#network-peers) | `/network/peers`
+[Network ENR Address](#network-enr-address) | `/network/enr`
+
+## Beacon Head
+
+Requests information about the head of the beacon chain, from the node's
+perspective.
+
+#### Example Response
+
+```
+{
+	slot: 1,
+	block_root: "0xe7bb65da065e8ea1f6d5adde378348ed",
+	state_root: "0x981fbdb550d2b5a36370f471fc16ddcf"
+}
+```
+
+### HTTP Specification
+
+| Property | Specification |
+| --- |--- |
+Path | `/beacon/head`
+Method | GET
+JSON Encoding | Object
+Query Parameters | None
+Typical Responses | 200
+
+#### HTTP Example
+
+```
+$ curl "localhost:5051/node/head"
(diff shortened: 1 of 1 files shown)


paulhauner:
As mentioned by @terencechain in #2, Prysm have been doing lots of work on the API too. This is great and I've been using it as a reference for this.

After chatting on Discord, I thought it might be useful if I listed other place's where this PR differs from the Prysm doc. That way we can see if I've made some errors in my decision making (likely).

I reference mainly [api.prylabs.net](https://api.prylabs.net/) instead of [ethereumapis](https://github.com/prysmaticlabs/ethereumapis) because the former is in HTTP-style and the latter is more gRPC focused.

#### Get blocks/states by epoch

I think it's ambiguous at to whether or not querying by `epoch` should return `slots_per_epoch` blocks, or the first block in the epoch (e.g., a finalized checkpoint).

I don't think this query is strictly necessary and that the typing in our API would be clearer without it.

#### Can't get multiple blocks/states

Trying to keep this PR small. There's a bit of complexity in getting multiple blocks (e.g., handling skip slots, insufficient history, etc). If it's a must have, I'd propose we do that in a separate PR following this one.

#### `/eth/v1aplha1` prefix

I'm not sure of the value of the `/eth/` prefix. It seems to me that having `/beacon` is enough without an `/eth/` prefix. I might be missing something though.

WRT to the `v1alpha1`, versioning makes sense. I figure we can define versioning in a later PR, however I'm happy to do it now if desired.

#### `chainhead` method replaced by `head`

My idea was to keep the `head` method small for now and say that reading the finalized checkpoints is a state-reading optimization for a future PR (e.g., provide a `/state/finalized_checkpoint` method so you don't need to download a whole state).

paulhauner:
Also, #1 is still open so we could decide that this markdown format is a bad idea.

prestonvanloon:
**Get blocks/states by epoch**

Filed that note here: https://github.com/prysmaticlabs/ethereumapis/issues/23
Replace the epoch filter and slot filter with a slot-range filter sounds like a great idea to me.

**Can't get multiple blocks/states**

If you are looking for a long lasting API, I urge you to reconsider. What is the use case where users would only want one block/state, aside from debugging?

We removed state fetches from our data API as it was going to be far too expensive to serve on average. Maybe the methods you are suggesting are under a `/debug` service which can be conditionally enabled in some deployments.

**/eth/v1aplha1 prefix**

The decision for this was largely influenced by the idea what github.com/prysmaticlabs/ethereumapis could be a large collection of any API that applied in the ethereum ecosystem. Layer 2 solutions, public dApps, anything with an API. So, we prefix with the general domain `eth` and the versioning `v1alpha`.

In the context of this repository where it is only Ethereum 2.0 APIs, it isn't necessary. 
The versioning prefix, we took inspiration from this guide: https://cloud.google.com/apis/design/versioning
We haven't written out the version migration pattern, but likely `v1beta1` would launch with phase 0 then we finalize it at some point after running in production to `v1` 

**chainhead method replaced by head**

I don't see how this is better to make multiple requests for the head information. I suppose if you only wanted a subset of the payload then you could use GraphQL or some response filter.

paulhauner:
Thanks for the response!

> If you are looking for a long lasting API, I urge you to reconsider. What is the use case where users would only want one block/state, aside from debugging?

I'm totally happy to make a new PR straight after this that describes that.

Are your urging me to include it in this PR?

> The versioning prefix, we took inspiration from this guide: https://cloud.google.com/apis/design/versioning

Cool. Do you think we should version the `/beacon` and `/network` endpoints individually, or group them all under one version?

> I don't see how this is better to make multiple requests for the head information.

Yeah, I'm down to add more info to `/beacon/head`. You've changed my mind.

cemozerr:
Great work @paulhauner! I'm going through, implementing this API for Artemis right now. For the "/beacon/head" path, do you think it's necessary to return beacon state root? IMO, since beacon blocks already include the state root, that might be a bit redundant. Also, the Store uses beacon block roots to index the corresponding state in time.

djrtwo:
I'm keen to get some basics merged here. Where do things stand @paulhauner. Any thing to update/modify before an initial merge?

cemozerr:
To add to my previous comment, in the "/beacon/state" path, would it not be better to index the beacon state using a beacon block root hash tree root, since that how we index beacon state's in the Store object?

jrhea:
@paulhauner is it worth adding `epoch` in addition to slot considering that someone could be running minimal constants?

paulhauner:
> @cemozerr: Great work @paulhauner! I'm going through, implementing this API for Artemis right now. For the "/beacon/head" path, do you think it's necessary to return beacon state root? IMO, since beacon blocks already include the state root, that might be a bit redundant. Also, the Store uses beacon block roots to index the corresponding state in time.

Thank you! I agree that it's not maximally minimal (lol) to return the beacon state root, however it is convenient for users. @prestonvanloon touched upon this in the "chainhead method replaced by head" part of his [comment](https://github.com/ethereum/eth2.0-APIs/pull/3#issuecomment-524127137). I ended up capitulating to Preston's idea that including things like `state_root` in the `/head` route will make life easier for API consumers and it seems to come at little cost for us developers.

> @cemozerr: To add to my previous comment, in the "/beacon/state" path, would it not be better to index the beacon state using a beacon block root hash tree root, since that how we index beacon state's in the Store object?

I assume the `Store` object is the one referenced in [fork choice](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/core/0_fork-choice.md#store)?

I think it may be confusing to users if we index `BeaconState` by the root of a `BeaconBlock`, as it deviates from the easy-to-comprehend concept of _"index `Thing` by `root(Thing)`"_.

My gut feeling is that keying `BeaconState` by a `BeaconBlock` root would be shaping the API to suit a particular description of the fork choice algorithm, instead of shaping it to make life easier for API consumers. I would understand if the specification was heavily biased towards keying by block root (instead of state root) but as far as I can tell, the `Store` is only used to describe a fork choice implementation and that implementation is not "production ready" (it's slow) and therefore unlikely to be used in the long term.

> @jrhea: is it worth adding epoch in addition to slot (for each response) considering that someone could be running minimal constants?

I could be wrong, but I think you're suggesting to permit querying for items via `epoch`, as well as `slot` and `root`? Sorry if I misunderstood.

If so, I remember chatting to @prestonvanloon about this. I noted that when querying by `epoch` it's not clear if you should get `SLOTS_PER_EPOCH` items or the item at the first slot of the epoch. IIRC we agreed that the benefits from querying by epoch don't outweigh the potential confusion.

> @djrtwo: I'm keen to get some basics merged here. Where do things stand @paulhauner. Any thing to update/modify before an initial merge?

I think the main thing to consider is #1. Since @mpetrunic is pushing forward with the Swagger API, perhaps this markdown version is just duplicate effort?

If we do want to go ahead with this, I think all I need to do is add more info to the `/head` endpoint (as per @prestonvanloon's comments, but perhaps not if @cemozerr disagrees).

jrhea:
@paulhauner thanks for the reply. 

> I could be wrong, but I think you're suggesting to permit querying for items via epoch, as well as slot and root? Sorry if I misunderstood.

Actually, I was suggesting adding epoch to the response.  Agreed that it would be confusing to query by epoch, but there is some value to knowing the epoch of the response bc slots_per_epoch is different for minimal and main net configs

mpetrunic:
Do we wanna follow more restful endpoints like:
- beacon/blocks/head
- beacon/slots/{slot}
- beacon/blocks/{root}
?

https://restfulapi.net/resource-naming/

paulhauner:
> Actually, I was suggesting adding epoch to the response. Agreed that it would be confusing to query by epoch, but there is some value to knowing the epoch of the response bc slots_per_epoch is different for minimal and main net configs

Oh yeah I'm with you now. We had this problem on Lighthouse, we had to expose a `spec/slots_per_epoch` endpoint to resolve it.

I think my preference would be to define the [`config.yaml`](https://github.com/ethereum/eth2.0-specs/blob/dev/configs/mainnet.yaml) as a struct which can be passed over the wire. We do something along these lines (same information, different struct) for Lighthouse, which ensures that the validator client works on the same constants as the beacon node. Thoughts?

mkalinin:
I am in favor of conformance with validator API and using RESTful endpoints like @mpetrunic mentioned. This is very common practice in the web. Usually, query string contains optional parameters of the request while parameter in the path makes it obligatory which is straightforward when resources are requested by their identifiers.

djrtwo:
Merging to get the base in. We can work from here.

PR #5, opened 2019-10-16
Add endpoint to get validator details

mpetrunic:
I'm primarily adding this endpoint as it extremely useful for validator clients to check their details and see if they are no longer active validators (slashed).

### DIFF
--- a/apis/validator/beacon-node-oapi.yaml
+++ b/apis/validator/beacon-node-oapi.yaml
@@ -87,6 +87,32 @@ paths:
         500:
           $ref: '#/components/responses/InternalError'
 
+  /validator:
+    get:
+      tags:
+        - MinimalSet
+      summary: "Get validator details."
+      description: "Requests the beacon node to provide details about validator or 404 if validator is not present in validator registry"
+      parameters:
+        - name: validator_pubkey
+          in: query
+          required: true
+          schema:
+            $ref: '#/components/schemas/pubkey'
+      responses:
+        200:
+          description: Success response
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Validator'
+        400:
+          $ref: '#/components/responses/InvalidRequest'
+        404:
+          description: "Validator not found in registry"
+        500:
+          $ref: '#/components/responses/InternalError'
+
   /validator/duties:
     get:
       tags:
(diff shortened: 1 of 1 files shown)


apis/validator/beacon-node-oapi.yaml
```diff
@@ -290,6 +316,40 @@ components:
           format: uint64
           nullable: true
           description: "The slot in which a validator must propose a block, or `null` if block production is not required."
+    Validator:
+      type: object
+      properties:
+        validator_pubkey:
+          $ref: '#/components/schemas/pubkey'
+        withrawal_credentials:
```
djrtwo:
```suggestion
        withdrawal_credentials:
```

apis/validator/beacon-node-oapi.yaml
```diff
@@ -290,6 +316,40 @@ components:
...
+          type: string
+          format: bytes
+          pattern: "^0x[a-fA-F0-9]{64}$"
+          description: "Hash of withdrawal credentials"
+        effective_balance:
+          type: integer
+          format: int64
```
djrtwo: why `int` instead of `uint`?

```suggestion
          format: uint64
```

apis/validator/beacon-node-oapi.yaml
```diff
@@ -290,6 +316,40 @@ components:
...
+          type: boolean
+          description: "Was validator slashed (not longer active)."
+        activation_eligibility_epoch:
+          type: integer
+          format: uint64
+          description: "When criteria for activation were met."
+        activation_epoch:
+          type: integer
+          format: uint64
+          description: "Epoch when validator activated. Null if not activated"
```
djrtwo: Why Null instead of `FAR_FUTURE_EPOCH` as the field is in state?

apis/validator/beacon-node-oapi.yaml
mpetrunic: oApi defines only int64 type: https://swagger.io/docs/specification/data-models/data-types/#numbers

Do we wan't to define uin64 as string type to avoid rounding in various tools?

apis/validator/beacon-node-oapi.yaml
mpetrunic: should be far future epoch^^

apis/validator/beacon-node-oapi.yaml
djrtwo: hmmm uint64 is defined in a number of fields above. 
It'd be a nice to have. What do you recommend?

mpetrunic: We can leave like this for now. I've opened [issue](https://github.com/ethereum/eth2.0-APIs/issues/9) to gather more feedback.

PR #15, opened 2019-12-24
0.10.1 Validator api

### LINKED ISSUE #12: Validator api - spec 0.9.2
I would propose new validator API as current API is causing troubles in this aggregation strategy.
Problems are:
- we are suppose to call `duties` for current epoch to check if we are proposer and for future epoch so we can subscribe to attestations (call for proposer checking will subscribe us (beacon node) to unnecessary topics)
- every time we check for future attesting assignments we are calculating proposers unnecessary
- there is no API to fetch attestations to aggregate

I would propose following:
- GET `/eth/validator/duties/proposer/{epoch}` - returns `{slot: publicKey}` mapping for every slot in given epoch
- GET `/eth/validator/duties/attester/{epoch}?publicKeys=public1, public2` - returns `[{publicKey, slot, committeeIndex}]` for given publicKeys and subscribes to topics related to returned commiteeIndices
- GET `/eth/validator/attestations/{commiteeIndex}` - returns all attestations collected from that committee index topic 
- POST `/eth/validator/aggregate`

Other apis will remain the same (attestation type needs change).

After we have agreed api, I'll open PR on repo.

mpetrunic:
This PR is mostly based on issue #12 
I made few additional changes:
- subscribing to committee attestation subnet requires validator signature so malicious use cannot spam beacon nodes and subscribe them to all subnet topics
- validator can produce aggreagate in two ways:
   - obtain all wire attestations for given committee and slot, aggregate and publish AggregateAndProof which allows validator to collect attestation from multiple BN
   - request aggregated attestations from BN where BN does the heavy work and VC is required just to add signature and call endpoint to publish aggregate

You can checkout new api visually in  my repo: https://mpetrunic.github.io/eth2.0-APIs/

resolves #12


mpetrunic:
@mkalinin and @paulhauner I would like your reviews here as well. It's probably easier to iterate on PR

paulhauner:
Oh sorry, I did https://github.com/ethereum/eth2.0-APIs/issues/12#issuecomment-570431686 before I saw this!

AgeManning:
I have a few comments and questions:

The splitting of attestation and block production duties would imply that you would be calling the `/attester` endpoint more regularly than the `/proposer`. In principle, there could be a re-org, which is why we currently call both each slot. In this new API how often would you call each, and how are re-orgs handled? 

Not sure I follow the need for the `beacon_committee_subscription/{committee_index}/attestations` and `/validator/aggregate_and_proof`. 

I would imagine the logic flow to go something like:
Validator client gets it's attestation duties. After it realises the slot to attest to, it sends `beacon_committee_subscription` informing the beacon node to join the necessary subnet. As usual, the validator client will sign its singular attestation and send to the beacon node. The beacon node, which now has the slot signature and pubkey of the validator for this committee can work out if it needs to aggregate and publish or do nothing. The beacon node then aggregates and publishes if the validator is required to do so. If there are many validators in the committee it may only publish one rather than duplicates. 

If we are concerned with multiple BN's per single validator, then the method above will cause beacon nodes to duplicate messages, but they should be the same aggregate, and we now filter those at the gossipsub layer so there's no real harm or extra bandwidth with multiple beacon nodes publishing the same message.

AgeManning:
Also, the subscription endpoint currently only takes a single validator as an input. For large number of validators, we would need to do a large number of http requests. 

Perhaps we should allow the subscription endpoint to to pass a list of validators?

mpetrunic:
Hey @AgeManning, my huge reply disappeared, luckily you've made another comment :laughing: 

> The splitting of attestation and block production duties would imply that you would be calling the /attester endpoint more regularly than the /proposer. In principle, there could be a re-org, which is why we currently call both each slot. In this new API how often would you call each, and how are re-orgs handled?

That's because of attestation duty lookahead. You are suppose to check duties for next epoch and block proposers in that requests are useless. Since attestation are stable for +1 epoch, you don't really need to check attestation duties in current epoch as you've already recorded at which slot you are suppose to attest. As for reorgs, I don't think we will see reorgs 5 epochs behind (it takes at least 5 epochs to exit validator) anytime soon. In any case, it would make much more sense to add some server side event api which would notify validator on reorg instead of pinging duties API every slot.

> Not sure I follow the need for the beacon_committee_subscription/{committee_index}/attestations and /validator/aggregate_and_proof.

Reasons:
- allows easier BN redundancy - publish same aggregate on N different BNs
- protection against malicious BNs
- ensured better aggregation - you have ~4s to collect up to 64 attestations
   - in your env where everything is running on 2 nodes it isn't a problem to propagate it but when we have more larger network with far away nodes and latencies, 3 nodes with different peers have better chance to collect all attestations. 
- it's simpler API to implement
  
How did you implement single api for aggregation and publish?
I assume you receive request from validator, subscribe to topic, find nodes, wait for validator to publish attestation, collect attestation for 4 seconds and then publish? How does validator know if request is completed successfully? Does it wait for response entire time?

> Trust in the BN ( had a few debates about this, and given the current API, I think the BN has to be trusted regardless of whether we aggregate in the VC or BN).

Out of curiosity, what are the reasons to trust BN?

> Also, the subscription endpoint currently only takes a single validator as an input. For large number of validators, we would need to do a large number of http requests.
Perhaps we should allow the subscription endpoint to to pass a list of validators?

Yeah that's cool!

AgeManning:
Thanks for the detailed response. 

>  I don't think we will see reorgs 5 epochs behind (it takes at least 5 epochs to exit validator) anytime soon. In any case, it would make much more sense to add some server side event api which would notify validator on reorg instead of pinging duties API every slot.

Wouldn't it be possible that a re-org of 1 epoch (or a few slots on an epoch boundary) changes the shuffling and hence validator duties?
I agree it would be nice to add a notification to notify the VC, however we would then have to add some mechanism, i.e http server on the VC to allow the BN to notify it. (I guess this can be left to client implementations). 

To add some clarification, I think it's sensible to allow the validator client to choose which beacon node to publish on and hence requires the extra http endpoints. I think the logic I was suggesting aligns with your points also:
> allows easier BN redundancy - publish same aggregate on N different BNs
> ensured better aggregation - you have ~4s to collect up to 64 attestations
> it's simpler API to implement

All connected BN's would publish the aggregate if the validator required to do so and there are less API endpoints.

> Out of curiosity, what are the reasons to trust BN?

Mainly the complexity that would have to be added to the VC. If a BN is malicious it would be quite difficult to counter as the VC has no knowledge of the blockchain, consensus, validator pool etc. A beacon node in principle could feed malicious blocks etc to the VC and as long as there is no obvious double voting the VC will happily sign and publish. Therefore, we've assumed a trust model were the VC simply trusts the BN. 

In terms of the API and logic, I'm thinking of something like this (not entirely thought out, will report back): 

1. VC subscribes it's validators to shards/committees, which notifies the BN in-advance (potentially 1 epoch) which val client is connected and wants to receive attestations for which epoch and slot. The BN then has time to run discovery and find peers of the required subnet and subscribe to the topic.

2. VC publishes unaggregated attestations whilst the BN collects and aggregates all attestations on the subscribed channels and slots.

3. VC sends something like `publish_and_proof` to whichever BN it wants to publish on. The BN checks the proof, if it's valid, publishes the aggregated attestations it's collected on behalf of the VC.

mpetrunic:
> Wouldn't it be possible that a re-org of 1 epoch (or a few slots on an epoch boundary) changes the shuffling and hence validator duties?

Validator exit is only effective after 5 epoch after block in which it's announce and since only active validator indices have effect on getting proposer and committee it seems that reorg needs to be at least 5 epochs before. I might be missing something.

> I agree it would be nice to add a notification to notify the VC, however we would then have to add some mechanism, i.e http server on the VC to allow the BN to notify it. (I guess this can be left to client implementations).

Solution is server-side-events, quite simple to implement and consume. I think prysmatic already has similar endpoints. Might be good idea to standardize (way better than pinging BN every slot) :slightly_smiling_face: 

> All connected BN's would publish the aggregate if the validator required to do so and there are less API endpoints.

But this could be potentially bad if there aggregate is different, right? (like they missed to detect some aggregation).

> Mainly the complexity that would have to be added to the VC. If a BN is malicious it would be quite difficult to counter as the VC has no knowledge of the blockchain, consensus, validator pool etc. A beacon node in principle could feed malicious blocks etc to the VC and as long as there is no obvious double voting the VC will happily sign and publish. Therefore, we've assumed a trust model were the VC simply trusts the BN.

But you could be connected to multiple nodes and compare outputs?

> VC publishes unaggregated attestations whilst the BN collects and aggregates all attestations on the subscribed channels and slots.

Seems a bit redundant to aggregate all attestations as you only need to aggregate attestations which match aggregator vote though.

AgeManning:
> Validator exit is only effective after 5 epoch after block in which it's announce and since only active validator indices have effect on getting proposer and committee it seems that reorg needs to be at least 5 epochs before. I might be missing something.

I think you are missing the fact that the shuffling can change on a re-org. 

> Solution is server-side-events, quite simple to implement and consume. I think prysmatic already has similar endpoints. Might be good idea to standardize (way better than pinging BN every slot)

Yeah, I think they are using gRPC. Standardising this would be ideal.

> But this could be potentially bad if there aggregate is different, right? (like they missed to detect some aggregation).

This is would be fine. There is potentially a loss of income as if we got attestations from all nodes, potentially we could build an aggregate that has more attestations, if such an attestation could be formed. There is a bit of difficulty in optimising for finding the best aggregate. 

> But you could be connected to multiple nodes and compare outputs?

This would be quite difficult. If you connected to 2 nodes. And the blocks are different, which do you trust? If you connect to lots of nodes, and you employ a simple majority, each node could be on a different fork, have a different view of the network and essentially you'd have your validator client making a judgement on which fork (based on majority of your connected nodes). For attestations, nodes will also have different views and will provide different attestations to sign, how do you tell which is malicious and which are not? In principle, I think it could be possible to build some logic that attempts all of this, but I think its non-trivial. 

> Seems a bit redundant to aggregate all attestations as you only need to aggregate attestations which match aggregator vote though.

Yeah. Preventing this will require knowledge of the validator proof. I guess the validator should send this in advance during step 1. 
Although, attestations can be included in future blocks, so it may be beneficial for clients to aggregate and potentially use in future blocks for profit maximisation. In either case, the more general case for the API is to allow clients to choose, which involves sending the proof with the subscription I think.

mpetrunic:
Sorry for the delay, I was on holiday. I will add streaming api for fork choice and allow subscription endpoint to allow multiple validators to subscribe. 

I would opt-in to merge in with both options for aggregations until [#1570](https://github.com/ethereum/eth2.0-specs/issues/1570) is resolved? cc @AgeManning

hwwhww:
Hi all, does it make sense if we bump to v0.10.x APIs? What's the latest target spec version teams are working now? In Trinity, we are targeting on v0.10.x, so we are making some improvisation with it. 😄

mpetrunic:
Well current types are 0.9.4. I might be wrong but it seems everyone is targeting 0.9.4. Would be nice if we merge something so we can add other stuff 

cc @djrtwo

AgeManning:
Ok, sorry for the delay. This is what lighthouse will be merging to master soon:

`ValidatorDuty` now has an `aggregator_modulo` field. This allows the validator client given a duty to work out if it's supposed to be an aggregator or not and then subsequently send subscriptions. 

`/validator/subscribe` - POST sends a list of subscriptions to the beacon node. A subscription is:
```
Subscription {
   validator_index: u64,
   attestation_committee_index: CommitteeIndex,
   slot: Slot,
   slot_signature: Signature
}
```
`/validator/attestation` - GET - gets a raw attestation for the slot, committee
`/validator/aggregate_attestation` - GET - gets the aggregate attestation for the slot, committee
`/validator/attestations` - POST - publishes a list of raw attestations
`/validator/aggregate_and_proofs` - POST - publishes a list of signed aggregate and proofs

mpetrunic:
Hey no problem, let's work something out :slightly_smiling_face: 

> `ValidatorDuty` now has an `aggregator_modulo` field. This allows the validator client given a duty to work out if it's supposed to be an aggregator or not and then subsequently send subscriptions.

I like this very much.

> Subscription {
>    validator_index: u64,
>    attestation_committee_index: CommitteeIndex,
>    slot: Slot,
>    slot_signature: Signature
> }

I wish we would stick to pubkey as validator identifier and not mix both. My reasoning is that validator knows his pubkey while his index, he either has to find out trough duties or separate api and than it has to store it.

> `/validator/aggregate_attestation` - GET - gets the aggregate attestation for the slot, committee

Ain't this suppose to have hash of validator `AttestationData` so you don't get aggregate without your attestation?

> `/validator/attestations` - POST - publishes a list of raw attestations
> `/validator/aggregate_and_proofs` - POST - publishes a list of signed aggregate and proofs

Did you had issues with requests when this api accepted only single attestation/aggregate?

AgeManning:
> I wish we would stick to pubkey as validator identifier and not mix both. My reasoning is that validator knows his pubkey while his index, he either has to find out trough duties or separate api and than it has to store it.

I'm fine with standardising this. I used index specifically because I was trying to minimise state lookup in certain places. The validator client needs to know it's duties before sending a subscription. In which case it has found its validator index and when getting duties, the state lookup has already happened to get this data. I opted to use the index to avoid a state lookup a second time, which can be expensive. But I'm not super attached to this. 

> Ain't this suppose to have hash of validator AttestationData so you don't get aggregate without your attestation?

Umm not sure about this. In my impl each slot, the validator will publish it's raw attestation, so the beacon node has it. The beacon node then aggregates everything it sees (including the validator(s) that published via the API). At the appropriate time the validator asks for the aggregate and the beacon node will return the total aggregate (including any local validators attestations that published in that slot)

> Did you had issues with requests when this api accepted only single attestation/aggregate?

Yes. We want to support large validators. We were running a single validator client with 100,000 validators. It's not great doing 100,000 requests each time, much better to group them. Although this isn't realistic, even with 20 validators or so, I think its better to group into single requests.

mpetrunic:
> Umm not sure about this. In my impl each slot, the validator will publish it's raw attestation, so the beacon node has it. The beacon node then aggregates everything it sees (including the validator(s) that published via the API). At the appropriate time the validator asks for the aggregate and the beacon node will return the total aggregate (including any local validators attestations that published in that slot)

That's not really stateless, it would be issue if you are running couple of nodes behind load balancer (which I assume some will do to ensure redundancy). 

> Yes. We want to support large validators. We were running a single validator client with 100,000 validators. It's not great doing 100,000 requests each time, much better to group them. Although this isn't realistic, even with 20 validators or so, I think its better to group into single requests.

I'm fine either way, just asking because in real case scenario, your validators won't be aggregators at the same time (most likely) and it makes it a bit weird to propagate error if some request failed.

djrtwo:
What are the last things to still figure out here?

* pubkey for identifier seems reasonable for the subscription. I assume subscriptions are sent very infrequently so the cost of the state lookup is negligible 
* I think `/validator/aggregate_attestation - GET` does need a particular attestation-data hash or validator pubkey in the request. The BN might have multiple subscriptions from different validators at same slot/committee with different attestation data published from each. In this case, it would have multiple local aggregates for the same slot/committee
* As for publishing, I think the list POST makes sense for raw attestations, and I like the list structure mirrored in aggregates. Agreed that the likelihood of aggregating at the same slot as other local validators is low, it is certainly something that will happen

AgeManning:
Sorry for the late reply. Have been off making sure this implementation works. 

> pubkey for identifier seems reasonable for the subscription. I assume subscriptions are sent very infrequently so the cost of the state lookup is negligible

Agreed. We've since cached this, so its no longer an issue for us. 

> I think /validator/aggregate_attestation - GET does need a particular attestation-data hash or validator pubkey in the request. The BN might have multiple subscriptions from different validators at same slot/committee with different attestation data published from each. In this case, it would have multiple local aggregates for the same slot/committee

> with different attestation data published from each

Of course, I missed this (@mpetrunic sorry for missing this point). So I guess the hash seems to be the most efficient way to track the attestation. The validator client, should then keep a hash of the attestation it published (if the validator is an aggregator) and re-send it with the `aggregate_attestation`. 

> As for publishing, I think the list POST makes sense for raw attestations, and I like the list structure mirrored in aggregates. Agreed that the likelihood of aggregating at the same slot as other local validators is low, it is certainly something that will happen

Agreed. 

I'll modify our `aggregate_attestation` to include the hash of an `Aggregate`.

mpetrunic:
@djrtwo @AgeManning I think I've addressed all issues now.

djrtwo:
merge conflicts 😩

djrtwo:
Thanks for the merge. Changed title of PR to "0.10.1 Validator api"

types/validator.yaml
```diff
@@ -38,6 +38,10 @@ schemas:
     properties:
       validator_pubkey:
         $ref: './misc.yaml#/schemas/PublicKey'
+      aggregator_modulo:
+        allOf:
+          - $ref: './misc.yaml#/schemas/Uint64'
+          - description: "Modulo used in [spec](https://github.com/ethereum/eth2.0-specs/blob/v0.9.4/specs/validator/0_beacon-chain-validator.md#aggregation-selection) to determine if validator is aggregator"
```
djrtwo: I had to chase down exactly what was being passed around here.

Might make sense to put a clarifying comment, that the validator can then locally calculate `is_aggregator` (aka `bytes_to_int(hash(slot_signature)[0:8]) % aggregator_modulo == 0`)

djrtwo:
One minor comment. Looks good.

Want another look by @AgeManning before merge

djrtwo:
Discussed with @paulhauner and @AgeManning  that it might make sense to just send the `attestation_data` instead of the root of it with `/validator/aggregate_attestation`. The overhead is low for a local API and provides the node with more information

@paulhauner had some additional reasoning if he wants to chime in

mpetrunic:
> aggregate_attestation



> Discussed with @paulhauner and @AgeManning that it might make sense to just send the `attestation_data` instead of the root of it with `/validator/aggregate_attestation`. The overhead is low for a local API and provides the node with more information
> 
> @paulhauner had some additional reasoning if he wants to chime in
Ssz serialized?
Because that is GET request where we cannot send body, rather is has to be url encoded.
I would love if @paulhauner would share his reasoning :)

paulhauner:
There’s nothing particularly interesting 🙂

I just couldn’t see a convincing reason to hash it. What we gain by saving 100 bytes or so doesn’t seem worth what we lose by making the request opaque (harder for troubleshooting) and slightly more expensive/complex to match with an exisiting attestation.

If we hash we also have to send through the slot and index too. Just seems odd to hash something that doesn’t _need_ to hashed. Especially when we do require access to some of the preimage.

mpetrunic:
> There’s nothing particularly interesting 
> 
> I just couldn’t see a convincing reason to hash it. What we gain by saving 100 bytes or so doesn’t seem worth what we lose by making the request opaque (harder for troubleshooting) and slightly more expensive/complex to match with an exisiting attestation.
> 
> If we hash we also have to send through the slot and index too. Just seems odd to hash something that doesn’t _need_ to hashed. Especially when we do require access to some of the preimage.

So are you suggesting to ssz serialize it in request than?

paulhauner:
> So are you suggesting to ssz serialize it in request than?

That would make it no-longer human-readable, which would be a bit of a shame. Is it reasonable to use  `POST` or a `GET` with a body to achieve sending the `AttestationData`?

mpetrunic:
> > So are you suggesting to ssz serialize it in request than?
> 
> That would make it no-longer human-readable, which would be a bit of a shame. Is it reasonable to use  `POST` or a `GET` with a body to achieve sending the `AttestationData`?

POST is already taken, and GET body [should be ignored by servers] (https://stackoverflow.com/questions/978061/http-get-with-request-body).

You could use simpleserialize.com to make it human readable

paulhauner:
> POST is already taken

I believe we have the freedom to name it as we see fit, therefore resolving conflicts.

> GET body should be ignored by servers.

According to the approved answer in that post, the SHOULD was made obsolete in 2014.

mpetrunic:
> I believe we have the freedom to name it as we see fit, therefore resolving conflicts.

I think we should stick with [rest naming and conventions](https://restfulapi.net/resource-naming/).

> According to the approved answer in that post, the SHOULD was made obsolete in 2014.

Yeah but most tools/client libraries/server implementation/cache layers don't support it and I don't we should be ice breakers for those things.

I don't see problem with ssz in this case, users debugging will be perfectly able to deserialize it while eth2 clients can log it out in human readable format.

I'll update PR to whatever most people agrees.

paulhauner:
> I don't see problem with ssz in this case

Works for me!

ajsutton:
Couple of questions about the APIs to create aggregate attestations:

 - In `GET /validator/aggregate_and_proof` the validator provides `attestation_data`, `committee_index` and optionally `slot` but `attestation_data` already includes slot and committee index. Am I missing some detail or should we just require `attestation_data`?

- The response from `GET /validator/aggregate_and_proof` is an array of `Aggregate` where `Aggregate` contains the `aggregator_index` but to calculate the index of the aggregator I'd need the slot number signed by the validator which the beacon node can't create.

- The description for the OK response to `GET /validator/aggregate_and_proof` says it "Returns only unaggregated attestations for given committee" but the API description says it aggregates all attestations in current slot. 

- Having `GET /validator_aggregate_and_proof` return an array of aggregates seems slightly strange given that https://github.com/ethereum/eth2.0-specs/blob/v0.10.1/specs/phase0/validator.md#attestation-aggregation requires the aggregator to publish only their best aggregation.

AgeManning:
We have GET `/validator/aggregate_attestation` It takes `attestation_data` (so the beacon node knows which aggregate that is associated with the validator to return. It returns the aggregate attestation.  There is no need for slot, committee the beacon node can aggregate the attestation for all votes of the same attestation data.

We are almost finished testing out our API and will publish on our docs and post an issue here once complete

mpetrunic:
> * In `GET /validator/aggregate_and_proof` the validator provides `attestation_data`, `committee_index` and optionally `slot` but `attestation_data` already includes slot and committee index. Am I missing some detail or should we just require `attestation_data`?

Nice catch, originally it was (slot, index, attestation_data_hash) but we switched to attestation data so slot and index are indeed no longer necessary. I'll update.

> * The response from `GET /validator/aggregate_and_proof` is an array of `Aggregate` where `Aggregate` contains the `aggregator_index` but to calculate the index of the aggregator I'd need the slot number signed by the validator which the beacon node can't create.

You are suppose to call [POST beacon_committee_subscription api](https://mpetrunic.github.io/eth2.0-APIs/#/MinimalSet/post_validator_beacon_committee_subscription) before calling this api so you are already aggregating attestations for that aggregator. We could add some flow diagram in readme.

@AgeManning @paulhauner Should we change it to aggregator_pubkey to be consistent with pubkey as identifier?

> * The description for the OK response to `GET /validator/aggregate_and_proof` says it "Returns only unaggregated attestations for given committee" but the API description says it aggregates all attestations in current slot.

Typo.

> * Having `GET /validator_aggregate_and_proof` return an array of aggregates seems slightly strange given that https://github.com/ethereum/eth2.0-specs/blob/v0.10.1/specs/phase0/validator.md#attestation-aggregation requires the aggregator to publish only their best aggregation.

It returns array in case you have multiple validators so you can batch requests. As already discussed in this PR, it will rarely happen in production setup but it's useful for testnet so 100k validators aren't ddos-ing BN api.

ajsutton:
> You are suppose to call POST beacon_committee_subscription api before calling this api so you are already aggregating attestations for that aggregator. We could add some flow diagram in readme.

Why would the beacon node remember the aggregator index so that it can give it back to the validator which originally asked it to subscribe though?  Wouldn't the validator just remember it?

> It returns array in case you have multiple validators so you can batch requests. As already discussed in this PR, it will rarely happen in production setup but it's useful for testnet so 100k validators aren't ddos-ing BN api.

I suspect I'm missing something, but if there's only one `AttestationData` supplied to the request, why would the number of validators matter? Or is the `attestation_data` param actually an SSZ list of `AttestationData`? Worst case you'd still only have one `AttestationData` and resulting aggregate per-committee not per validator you're running but I can see the desire to batch those requests.

mpetrunic:
@ajsutton You are right, I'm just looking trough this PR discussion, I've probably made a mistake at some iteration. Hopefully, in future, we will update this in smaller increments.

GET aggregate_and_proof should return single aggregate for given attestation_data

djrtwo:
@mpetrunic Are we going to make the adjustment for GET aggregate_and_proof before merge? 

After that, ready when you are. Excellent work. Now time to iterate in smaller chunks!

PR #18, opened 2020-02-22
Fix broken links to eth2.0-specs

snario:


types/block.yaml
```diff
@@ -1,14 +1,14 @@
 schemas:
   BeaconBlock:
-    description: "The [`BeaconBlock`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beaconblock) object from the Eth2.0 spec."
+    description: "The [`BeaconBlock`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/phase0/beacon-chain.md#beaconblock) object from the Eth2.0 spec."
```
mpetrunic: Might be better if links would target spec version (github tag)

djrtwo: agreed

PR #19, opened 2020-03-16
0.10.1 types

mpetrunic:
- updated existing types
- added BeaconState object

I will try to migrate current beacon spec into openapi schema based on this branch


types/block.yaml
```diff
@@ -1,52 +1,69 @@
 schemas:
+  BeaconBlockCommon:
```
djrtwo: My bad on bumping to v0.11.0 links without fixing params.

Is the plan here to get v0.10.x cleaned up and release. Then update to v0.11?

types/block.yaml
```diff
@@ -1,52 +1,69 @@
...
+    # An abstract object to collect the common fields between the BeaconBlockHeader and the BeaconBlock objects
```
djrtwo: My bad on bumping to v0.11.0 links without fixing params.

Is the plan here to get v0.10.x cleaned up and release. Then update to v0.11?

types/misc.yaml
```diff
@@ -1,33 +1,19 @@
...
+      epoch:
+        $ref: "./primitive.yaml#/schemas/Uint64"
+  Checkpoint:
+    type: object
+    description: "The [`Checkpoint`](https://github.com/ethereum/eth2.0-specs/blob/v0.10.1/specs/phase0/beacon-chain.md#checkpoint"
+    properties:
+      epoch:
+        $ref: "./primitive.yaml#/schemas/Uint64"
+      root:
+        $ref: "./primitive.yaml#/schemas/Root"
```
djrtwo: github suggesting a newline here

types/operations/attestation.yaml
```diff
@@ -1,94 +1,66 @@
...
-              - $ref: '../misc.yaml#/schemas/Hash'
-              - description: "Root of the previous crosslink."
-          data_root:
-            allOf:
-              - $ref: '../misc.yaml#/schemas/Hash'
-              - description: "Root of the crosslinked shard data since the previous crosslink."
+      source:
+        $ref: "../misc.yaml#/schemas/Checkpoint"
+      target:
+        $ref: "../misc.yaml#/schemas/Checkpoint"
```
djrtwo: github suggesting newline

types/operations/deposit.yaml
```diff
@@ -1,37 +1,34 @@
...
         allOf:
-          - $ref: '../misc.yaml#/schemas/Hash'
+          - $ref: '../primitive.yaml#/schemas/Root'
           - description: "The withdrawal credentials."
       amount:
-        type: integer
-        format: uint64
-        description: "Amount in Gwei."
+        allOf:
+          - $ref: "../primitive.yaml#/schemas/Uint64"
```
djrtwo: Should we define a Gwei primitive?

types/primitive.yaml
```diff
@@ -0,0 +1,39 @@
...
+    description: "A string which uniquely identifies the client implementation and its version; similar to [HTTP User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3)."
+    example: "Lighthouse / v0.1.5 (Linux x86_64)"
+
+  GenesisTime:
+    type: integer
+    format: uint64
+    description: "The genesis_time configured for the beacon node, which is the unix time at which the Eth2.0 chain began."
+    example: 1557716289
+
+  Uint64:
```
djrtwo: why `Uint64` defined as a string? whereas `GenesisTime` is defined as an integer

types/state.yaml
```diff
@@ -0,0 +1,90 @@
...
+        type: string
+        format: byte
+        pattern: "^0x[a-fA-F0-9]+$"
+        description: "Bit set for every recent justified epoch"
+      previous_justified_checkpoint:
+        $ref: "./misc.yaml#/schemas/Checkpoint"
+      current_justified_checkpoint:
+        $ref: "./misc.yaml#/schemas/Checkpoint"
+      finalized_checkpoint:
+        $ref: "./misc.yaml#/schemas/Checkpoint"
```
djrtwo: github suggesting newline

types/block.yaml
mpetrunic: I though 0.10.x was going to be targeted version, I was wrong :smile: 
So yeah, I think it makes sense to update incrementally

types/operations/deposit.yaml
mpetrunic: yeah, we should!

types/primitive.yaml
mpetrunic: It was left from initial version of api, will update to uint64

PR #23, opened 2020-04-09
0.11.1 types

mpetrunic:
Notable changes:
- genesis endpoint returns "genesis_validator_root" so validator can construct domains for signatures
- removed "block_proposal_slot" from `AttesterDuty` (not sure how that was forgot there)

### DIFF
--- a/apis/validator/beacon-node-oapi.yaml
+++ b/apis/validator/beacon-node-oapi.yaml
@@ -11,7 +11,7 @@ servers:
   - url: "{server_url}"
     variables:
       server_url:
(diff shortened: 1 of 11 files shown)


mpetrunic:
@djrtwo I think someone is ticking of "Build forked pull requests" option on project settings :smile:

mpetrunic:
I will followup on weekend with PR to allow adding beacon apis since currently all definitions and endpoints are in validator directory.

djrtwo:
You're right.. the box was flipped.

it's back on now!

PR #33, opened 2020-04-20
Make proposer duties ssz serializable

mpetrunic:
We cannot ssz serialize map with dynamic keys hence return proposer duties as array of "containers" with static keys


djrtwo:
What's the status on this PR? Still want review?

mpetrunic:
> What's the status on this PR? Still want review?

I'll update it, that is format in new proposal anyways.

mpetrunic:
This will be inside PR to update validator duties so closing for now

djrtwo:
was just about to ask. thanks!

PR #36, opened 2020-05-14
Create main.yml

decanus:
possible method to remove external check dependencies by using github actions.


djrtwo:
cc @mpetrunic 

proposal to replace circleCI with a github action.
I'm kind of in the "if it ain't broke, don't fix it camp", but don't really care too much

.github/workflows/main.yml
```diff
@@ -0,0 +1,30 @@
...
+
+name: CI
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
mpetrunic, line 12: I wouldn't restrict it, I would let it run on every push and PR

decanus:
@mpetrunic fixed, @djrtwo ready to merge!

djrtwo:
okay, just disabled circleci build

PR #44, opened 2020-05-30
[WG proposal] /beacon/state endpoints

mpetrunic:


types/api.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - **pending_queued** - When validator is waiting to get activated, and have enough funds etc. while in the queue, validator activation epoch keeps changing until it gets to the front and make it through (finalization is a requirement here too).
+    - **active_ongoing** - When validator must be attesting, and have not initiated any exit.
+    - **active_exiting** - When validator is still active, but filed a voluntary request to exit.
+    - **active_slashed** - When validator is still active, but have a slashed status and is scheduled to exit.
+    - **exited_unslashed** - When validator has reached reguler exit epoch, not being slashed, and doesn't have to attest any more, but cannot withdraw yet.
+    - **exited_slashed** - When validator has reached reguler exit epoch, but was slashed, have to wait for a longer withdrawal period.
+    - **withdrawal_possible** - After validator has exited, a while later is permitted to move funds, and is truly out of the system.
+    - **withdrawal_done** - (not possible in phase0, except slashing full balance) - actually having moved funds away
```
protolambda, line 31: Can we filter on `pending`, `active`, `exited` and `withdrawal` too? To match all the respective sub-statuses?

types/state.yaml
```diff
@@ -1,92 +1,91 @@
...
+        allOf:
+          - $ref: './eth1.yaml#/schemas/Eth1Data'
+      maxItems: 1024
+    eth1_deposit_index:
+      $ref: "./primitive.yaml#/schemas/Uint64"
+    validators:
+      type: array
+      items:
+        allOf:
+          - $ref: './validator.yaml#/Validator'
```
protolambda, line 49: Why not specify max-items here, but for historical roots there is a max? While both are lists with large limits?

params/index.yaml
```diff
@@ -0,0 +1,9 @@
+StateId:
+  name: state_id
+  in: path
+  required: true
+  schema:
+    type: string
+  description: |
+    State identifier.
+    Can be one of: "head" (canonical head in node's view), "genesis", "finalized", "justified", \<slot\>, \<stateRoot\>.
```
protolambda, line 9: Slot and state root were going to be `/slot/<number>` and `/state_root/<hex>` to avoid duck typing the variable in the path, right? Or does that not fit in the API definition? Should we change it again?

apis/beacon/states/validators.yaml
```diff
@@ -0,0 +1,47 @@
...
+      description: "[Validator status specification](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ)"
+      in: query
+      required: false
+      schema:
+        type: array
+        uniqueItems: true
+        items:
+          allOf:
+            - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorStatus'
+            - enum: ["active", "pending", "exited", "withdrawal"]
```
protolambda, line 31: The other one uses the detailed sub statuses, but this one does not?

apis/beacon/states/validators.yaml
```diff
@@ -0,0 +1,47 @@
...
+  operationId: "getStateValidators"
+  summary: "Get validators from state"
+  description: "Returns filterable list of validators with their balance, status and index."
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+    - name: id
+      description: "Either public key or validator index"
```
protolambda, line 11: And what happened to the duck typing discussion for query params? Is it ok to do it like this?

apis/beacon/states/validator.yaml
```diff
@@ -0,0 +1,31 @@
...
+  summary: "Get validator from state by id"
+  description: "Returns validator specified by state and id or public key along with status and balance."
+  tags:
+    - Beacon
+    - MinimalValidatorApi
+  parameters:
+    - name: state_id
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+    - name: validator_id
+      description: "Either public key or validator index"
```
protolambda, line 12: ditto

types/api.yaml
mpetrunic: Those are just statuses that's returned in validator response. Those are exented with your suggestions in query param in actual endpoint

params/index.yaml
mpetrunic: Sorry, forgot about that. Ill update with those, it will dsplay better in swagger anyways

apis/beacon/states/validators.yaml
mpetrunic: allOf key combines those two so its effectively extends statuses that can be returned in ValidatorResponse

apis/beacon/states/validators.yaml
mpetrunic: Seems like nobody complained about that but I can separate it so we avoid duck typing here

mpetrunic:
@protolambda I've removed duck typed params. 
Pros:
- we can remove bySlot and byRoot endpoints from MinimalValidatorApi
- state id endpoints provides nice dropdown with allowed statuses

Cons:
- we now have bunch of endpoints with similar duplicated spec which is a bit harder to maintain

I'm fine with how it is now, but it could be issue if we try to avoid duck typing on second level like
`/beacon/states/slots/2/validators/pubkey/0xabe23ba...` for which we would have to spec 6 endpoints

mcdee:
Have to admit the lack of duck typing does make the spec verbose and somewhat confusing.  Endpoints like `/v1/beacon/states/{state_id}/validators` and `/v1/beacon/states/slots/{slot}/validators`, for example, no longer fit REST principles.

mpetrunic:
Reverted back to duck-typing and added "data" envelop to single item responses as per WG proposal

types/validator.yaml
```diff
@@ -1,80 +1,76 @@
...
-      committee_index:
-        allOf:
-          - $ref: "./primitive.yaml#/schemas/Uint64"
-          - description: "The committee index"
+AttesterDuty:
+  type: object
+  properties:
+    validator_pubkey:
+      $ref: './primitive.yaml#/schemas/PublicKey'
+    aggregator_modulo:
```
mcdee, line 40: This doesn't match the return info on the spreadsheet.

beacon-node-oapi.yaml
```diff
@@ -50,8 +58,6 @@ paths:
   /v1/node/health:
     $ref: "./apis/node/health.yaml"
 
-  /validator/{pubkey}:
-    $ref: "./apis/validator/validators.yaml"
   /validator/duties/{epoch}/attester:
     $ref: "./apis/validator/duties.yaml#/attester"
   /validator/duties/{epoch}/proposer:
```
mcdee, line 63: The return values here don't match the spreadsheet.  That said, this could just be an array of 32 public keys rather than the map of slot->pubkey as in the current repository if we want to go minimal.

mpetrunic: I haven't updated /validator endpoints. It will be separate pr.

 That said array won't work if you have small set of validators (common in dev/testnet chains) there can be slots without proposer duties

types/validator.yaml
mpetrunic: Haven't actually update anything. Just removed one level of tabs. Will update this in pr for /validator endpoints

beacon-node-oapi.yaml
mcdee: Ah okay; saw it in the PR so figured it was being updated.  Please ignore this and above :)

I thought that we'd always have a proposer for a slot; maybe not, then.

mpetrunic: I might be wrong, we had some wierd things with just 8 validators but it might have been some bug.

djrtwo: You always have a proposer assignment. Proposer's are sampled from the whole active set at each slot _with_ replacement, and proposer slashing is wrt slot rather than epoch

If just one validator, they will be assigned to propose at all slots

apis/beacon/genesis.yaml
```diff
@@ -13,9 +13,16 @@
...
+                  type: object
+                  properties:
+                    genesis_time:
+                      $ref: '../../beacon-node-oapi.yaml#/components/schemas/GenesisTime'
+                    genesis_validators_root:
+                      $ref: '../../beacon-node-oapi.yaml#/components/schemas/Root'
+                    genesis_fork_version:
+                      $ref: '../../beacon-node-oapi.yaml#/components/schemas/ForkVersion'
+      "404":
+        description: "Chain genesis has not yet occurred"
```
djrtwo, line 26: Maybe the following is a bit more accurate because there is a time in which you know the genesis data before genesis actually happens

```suggestion
        description: "Chain genesis info is not yet known"
```

apis/beacon/states/committee.yaml
```diff
@@ -0,0 +1,43 @@
+get:
+  operationId: "getEpochCommittees"
+  summary: "Get all committees for epoch"
+  description: "Retrieves the committees for the given state at the given epoch."
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+    - name: epoch
```
djrtwo, line 10: Should `epoch` be required? 

The call without `epoch` can default to `get_current_epoch(state)` of the `state_id`. It might be pretty normal to do something like `/beacon/states/head/committees` and having to get the head epoch adds an extra step here

apis/beacon/states/committee.yaml
```diff
@@ -0,0 +1,43 @@
...
+      required: true
+      schema:
+        $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+    - name: index
+      description: "Committee index"
+      in: query
+      required: false
+      schema:
+        $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+    - name: slot
```
djrtwo, line 22: Must the `slot` be from the correct `epoch`?
If so, is there a good way to note those? If not, can we note how these interact (e.g. pull all from epoch, then filter by slot -- so if slot is not within epoch, will return empty)

apis/beacon/states/fork.yaml
```diff
@@ -0,0 +1,24 @@
+get:
+  operationId: "getStateFork"
+  summary: "Get Fork object for requested state"
+  description: "Returns [Fork](https://github.com/ethereum/eth2.0-specs/blob/v0.11.1/specs/phase0/beacon-chain.md#fork) object for state with given 'stateId'."
```
djrtwo, line 4: Do we intend to hard-code `v0.11.1` for spec references? Will we update all after this initial release?

params/index.yaml
djrtwo, line 9: Should we specific how `stateRoot` needs to be encoded? (0x prefixed hex)

types/api.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - description: "Current validator balance in gwei."
+    status:
+      $ref: "#/ValidatorStatus"
+    validator:
+      $ref: "./validator.yaml#/Validator"
+
+
+ValidatorStatus:
+  description: |
+    Possible statuses:
```
djrtwo, line 22: Can I just specify a primary status (e.g. `pending`) rather than a full status with substatus (e.g. `pending_initialized`)?

If so, we should note that

djrtwo: Ah, i see  that `enum` does not list primary statuses individually. 

It might be useful to be able to just say `active` and get all the active's without having to list all active sub-statuses

types/api.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - **withdrawal_possible** - After validator has exited, a while later is permitted to move funds, and is truly out of the system.
+    - **withdrawal_done** - (not possible in phase0, except slashing full balance) - actually having moved funds away
+
+    [Validator status specification](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ)
+  enum: ["pending_initialized", "pending_queued", "active_ongoing", "active_exiting", "active_slashed", "exited_unslashed", "exited_slashed", "withdrawal_possible", "withdrawal_done"]
+  example: "active_ongoing"
+
+
+Committee:
+  description: Group of validators assigned to attest at specific slot
```
djrtwo, line 39: And that have the same committee index (shard in phase 1). There can be multiple committees per slot so a sub-distinction is probably warranted

apis/beacon/states/committee.yaml
mpetrunic: I will check, I think path params are required by default but I can make it explicit as well

apis/beacon/states/committee.yaml
mpetrunic: we cannot specify min/max to depend on another param. I think is fine to return empty array along with description that is expected behavior

apis/beacon/states/fork.yaml
mpetrunic: I'm keeping them all the same, so I can FindAndReplace to update everything to 0.12 once we have all the endpoints :smile:

types/api.yaml
mpetrunic: So in response, we have these details statuses, which in query param are extended with "active". Yeah we have inheritance here :rofl:

mpetrunic: https://github.com/ethereum/eth2.0-APIs/pull/44/files#diff-564ddc01cfd0c4c456b0e2b545375f0eR31

apis/beacon/states/committee.yaml
mcdee: Can we state that we return a 400 if the slot is not in the epoch?

PR #45, opened 2020-06-03
[WG Proposal] Pool endpoints

mpetrunic:
- up till now attestation submission endpoint accepted multiple attestation
   - it was to allow batching if someone is running a lot of validators against single node
   - I've switched it to single as per proposal, if we wan't to accept multiple we need to rethink responses in case some attestation fails validation and should be rejected or some attestation aren't broadcast because some criteria's aren't fulfilled

- any suggestions how to better link to validations node should provide? Maybe we can create some .md file with explanation when to reject operation, when to store for later broadcast etc


apis/beacon/pool/attestations.yaml
```diff
@@ -0,0 +1,54 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
+  responses:
+    "200":
+      description: Attestation is stored in pool and broadcast on appropriate subnet
+    "204":
+      description: Attestation is stored in pool and it will be broadcast once all requirements are fulfilled.
```
djrtwo, line 50: which requirements not be filled but could become filled?

apis/beacon/pool/attester_slashings.yaml
```diff
@@ -0,0 +1,40 @@
...
+                type: array
+                items:
+                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/AttesterSlashing'
+    "500":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
+
+post:
+  operationId: submitPoolAttesterSlashings
+  summary: Submit AttesterSlashing object to node's pool
+  description: Submits AttesterSlashing object to node's pool and node MUST broadcast it to network.
```
djrtwo, line 25: if passes validations

apis/beacon/pool/voluntary_exists.yaml
```diff
@@ -0,0 +1,42 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: '../../../beacon-node-oapi.yaml#/components/schemas/SignedVoluntaryExit'
+  responses:
+    "200":
+      description: Voluntary exit is stored in node and broadcasted to network
+    "204":
+      description: Voluntary exit is stored in node and it will be broadcast once all requirements are fulfilled. [See voluntary_exit topic validation](https://github.com/ethereum/eth2.0-specs/blob/v0.11.3/specs/phase0/p2p-interface.md#global-topics)
```
djrtwo, line 38: So this 204 puts a new requirement on nodes -- to store messages locally and broadcast them on certainly conditions. I'm pretty sure nodes don't hvae this functionality today and I'm not certain it's worth putting it in there.

apis/beacon/pool/attestations.yaml
```diff
@@ -0,0 +1,54 @@
...
+                type: array
+                items:
+                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
+    "500":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
+
+post:
+  operationId: submitPoolAttestations
+  summary: Submit Attestation object to node
+  description: Submits Attestation object to node. If attestation passes all validation constraint, node MUST publish attestation on appropriate subnet.
```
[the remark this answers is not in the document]
mpetrunic: yes. I wanted to link validation from spec but they are a bit hard to link since there is no anchor for each topic

apis/beacon/pool/attestations.yaml
mpetrunic:
```
attestation.data.slot is within the last ATTESTATION_PROPAGATION_SLOT_RANGE slots (within a MAXIMUM_GOSSIP_CLOCK_DISPARITY allowance) -- i.e. attestation.data.slot + ATTESTATION_PROPAGATION_SLOT_RANGE >= current_slot >= attestation.data.slot (a client MAY queue future attestations for processing at the appropriate slot).
```
Also if you received notification that block is received from one node and you submit to another (which hasn't seen that block yet) it will have to wait for that block before validating.


I was reluctant to copy constraints here because it would be hard to keep them up to date if spec changes. Any suggestions how to handle those?

apis/beacon/pool/voluntary_exists.yaml
mpetrunic: Yeah,makes sense. It's easy to validate epoch constraint against node

apis/beacon/pool/attestations.yaml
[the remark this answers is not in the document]
djrtwo: We can split them up but it won't be released until a week or more from now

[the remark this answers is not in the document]
mpetrunic: Should I just leave it like that for now or put general topic validations link?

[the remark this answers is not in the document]
djrtwo: just leave for now

apis/beacon/pool/attestations.yaml
djrtwo: should we allow a 204 here? Still seems to put the burden of new functionality on a node (store in pool and query if/when good). It could return an error and allow the message provider to decide what to do

mpetrunic: I guess it's fine for now. Haven't seen any VC that submits attestation ahead of time

apis/beacon/pool/attestations.yaml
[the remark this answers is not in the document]
rolfyone: nit: 'all validation constraints'

apis/beacon/pool/voluntary_exists.yaml
rolfyone: nit: 'and broadcast to network'

PR #46, opened 2020-06-07
Add data envelop to node endpoints

mpetrunic:
I thought conclusion was to have data envelop only on array responses but seems like there is majority to have "data" envelop on every endpoint to allow passing various metadata.

### DIFF
--- a/apis/node/identity.yaml
+++ b/apis/node/identity.yaml
@@ -10,6 +10,9 @@ get:
       content:
         application/json:
           schema:
-            $ref: "../../beacon-node-oapi.yaml#/components/schemas/NetworkIdentity"
+            type: object
+            properties:
+              data:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/NetworkIdentity"
     "500":
       $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
--- a/apis/node/syncing.yaml
+++ b/apis/node/syncing.yaml
@@ -13,13 +13,16 @@
             schema:
               type: object
               properties:
-                head_slot:
-                  allOf:
-                    - description: "Head slot node is trying to reach"
-                    - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
-                sync_distance:
-                  allOf:
-                    - description: "How much slots node needs to process to reach head. 0 if synced."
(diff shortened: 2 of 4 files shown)


rolfyone:
looks good. I don't have the build tools to check in swagger or something, so I didn't approve, but this is in line with what I was expecting.

mpetrunic:
> looks good. I don't have the build tools to check in swagger or something, so I didn't approve, but this is in line with what I was expecting.

Added instruction for rendering docs:
https://github.com/ethereum/eth2.0-APIs#render

apis/node/version.yaml
```diff
@@ -10,6 +10,9 @@ get:
       content:
         application/json:
           schema:
-            $ref: '../../beacon-node-oapi.yaml#/components/schemas/Version'
+            type: object
+            properties:
+              data:
+                $ref: '../../beacon-node-oapi.yaml#/components/schemas/Version'
```
rolfyone, line 16: I read the doc as this should return
``` 
{
  "data": {
   "version": String
  }
}
```
I could be mis-understanding, thoughts? currently, this returns
```
{
  "data": "Lighthouse / v0.1.5 (Linux x86_64)"
}
```
and the spreadsheet indicates:
```
data {
  version string
}
```
in the response...

apis/node/identity.yaml
```diff
@@ -10,6 +10,9 @@ get:
       content:
         application/json:
           schema:
-            $ref: "../../beacon-node-oapi.yaml#/components/schemas/NetworkIdentity"
+            type: object
+            properties:
+              data:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/NetworkIdentity"
```
rolfyone, line 16: This looks as expected.
there's another PR with example value and type questions, but that's just FYI, and should be kept outside this change, since it's well defined.

apis/node/syncing.yaml
```diff
@@ -13,13 +13,16 @@
...
+                data:
+                  type: object
+                  properties:
+                    head_slot:
+                      allOf:
+                        - description: "Head slot node is trying to reach"
+                        - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+                    sync_distance:
+                      allOf:
+                        - description: "How much slots node needs to process to reach head. 0 if synced."
```
rolfyone, line 25: nit: how many slots
not your words, I realise.

PR #47, opened 2020-06-12
Fix Uint64 example

Magicking:
The maximum value for uint64 is « 2^64–1 = 18446744073709551615 », previous example was not correct.

### DIFF
--- a/types/primitive.yaml
+++ b/types/primitive.yaml
@@ -26,7 +26,7 @@ schemas:
 
   Uint64:
     type: string
-    example: "18446744073709551615"
+    example: "1"
 
   Root:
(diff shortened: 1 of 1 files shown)


types/primitive.yaml
```diff
@@ -26,7 +26,7 @@ schemas:
 
   Uint64:
     type: string
-    example: "1213123123123123123123123122"
+    example: "18446744073709551615"
```
mpetrunic, line 29:
```suggestion
    example: "1"
```
I don't think we should put max value as example as it is suggested in query and path params. I also should not put random value, tnx for noticing

PR #51, opened 2020-06-16
[WG Proposal] Block endpoints

mpetrunic:


apis/beacon/blocks/attestations.yaml
```diff
@@ -0,0 +1,24 @@
+get:
+  operationId: getBlockAttestations
+  summary: Get block attestations
```
djrtwo, line 3: Why this accessor specific for attestations? Is `/beacon/blocks/{block_id}` not sufficient?
And if this accessor, why not for the rest of the operations?

mpetrunic: Afaik there is a lot of use cases with just block attestation without rest of block data so it would reduce network traffic.

cc @mcdee @mkalinin

djrtwo: I fine to merge if this was previously discussed

PR #60, opened 2020-07-01
Remove genesis_fork_version from getGenesis

arnetheduck:
The genesis fork version is not part of a state, so keeping it around would require keeping an archive / registry of fork ancient fork versions even though in theory, a client can start from an arbitrary (trusted) state. 

Publishing the current and previous fork versions could provide some information here.

### DIFF
--- a/apis/beacon/genesis.yaml
+++ b/apis/beacon/genesis.yaml
@@ -20,8 +20,6 @@
                       $ref: '../../beacon-node-oapi.yaml#/components/schemas/GenesisTime'
                     genesis_validators_root:
                       $ref: '../../beacon-node-oapi.yaml#/components/schemas/Root'
-                    genesis_fork_version:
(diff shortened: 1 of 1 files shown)


mpetrunic:
genesis_fork_version is part of spec config so you should have it always, no?

djrtwo:
Yes, genesis fork version will always stick around

copied from discord:

> There really is an implicit input to the state transition (the configuration). Note that the upgrade path for configuration is to never overwrite but only to add -- e.g. `PHASE_1_MAX_DEPOSITS` if we wanted to change max deposits per block in phase 1 -- so you'll always have access to that un-tainted genesis fork param

djrtwo:
I think this is part of a larger debate as to where configuration values really live.
Again, genesis fork version is a persistent vlaue that will always be known about to process deposits so I see no issue in leaving it here

PR #61, opened 2020-07-11
Release actions (CD)

### LINKED ISSUE #54: Api docs deployment
Current status:
- github pages hosted on master branch
- every file is downloaded on it's own
- slow initialization speed
- once we release we would have to use develop branch and we won't be able to see progress

Ideas:
- add to ci task to bundle all files into single blob (this would be useful for converting other formats as well)
- publish current development render from gh-pages branch and using bundle instead of separate files (faster loading speed)
- have some hosting service/server where we would host released api docs (are we gonna have some stable releases?)
- publish every PR somewhere and use some github bot to comment url of build

mpetrunic:
- bundles spec and ui and deploy to gh-pages branch (@djrtwo after merge and ci is passed, you should switch github pages to use gh-pages branch instead of master)
   - this will make spec loading instantly in UI as it will download single file instead
- add release action
   - when you push tag v* it will create spec bundle (json and yaml), create github release with commit changelog and upload bundles as release asset
- added section in readme on how to add released bundle in dropdown on deployed swagger-ui
   - to avoid github cors issue, it's using proxy server to download api spec (happy to hear better workaround if any)

resolves #54


djrtwo:
Github pages updated to `gh-pages`

PR #73, opened 2020-08-16
Add eventstream api

### LINKED ISSUE #25: BeaconNode APIs
Thus far, this repo only consists of "validator" APIs. To aid in application developers, we aim to conform upon a set of user-level "beacon node" APIs.

Currently, Prysmatic has a number of user-level APIs defined in [prysmaticlabs/ethereumapis](https://github.com/prysmaticlabs/ethereumapis) with some decent traction with block explorers and other applications, and Lighthouse has a number of APIs that have also begun to be used by various applications (link?).

I propose that we make a list of the obvious items first (and especially any that have overlap between prysmatic and lighthouse). We can list them out, get a feel for the general categories, structure of arguments, etc and move from there. Explicit notes on how/why users are using them today will be helpful in better understanding the API in general.

After that, we can debate anything that seemed "non-obvious" or maybe client specific

### LINKED ISSUE #31: Stream standards
It is desirable in many contexts to support streaming as a first class citizen in this API. Streaming/event type functionality is already seen to be [useful in eth1](https://geth.ethereum.org/docs/rpc/pubsub) and is also currently being used by consumers of both the prysm API and lighthouse websockets.

[Actual protocol aside] The two methods discussed for supporting this in the API are specific stream endpoints _or_ a single event stream.

* stream endpoints
    * designates `/stream` versions of a resource endpoint -- e.g. `/resources/stream`
    * pro: allows for explicit, granular control in the API to designate which resources are valuable to stream
    * con: adds overhead in maintaining as there are now poll and stream versions of (potentially) many endpoints
* event stream
    * an alternative brought  up by @mcdee is to have one stream endpoint that sends events to the client. The client can then decide if/how to query info related to the events (e.g. client receives `EVENT new_block 0xfffffff` and then calls `/beacon/chain/block/0xffffff` to retrieve the alerted about block.
    * pro: isolates stream functionality to one very specific place. arguably easier to maintain for client than several separate stream endpoints
    * pro: keeps core API simpler by not having duplicate endpoints for stream vs not
    * con: must define/maintain a separate list of events to be supported
    * does stream allow for subscription for subsets of events? if so, server must maintain routing table.

### LINKED ISSUE #37: [WG] proposal
Please take a look at proposed API endpoints (sheet `“[WG] Proposal"):
https://docs.google.com/spreadsheets/d/1kVIx6GvzVLwNYbcd-Fj8YUlPf4qGrWUlS35uaTnIAVg/edit#gid=1802603696

Feel free to comment here or in sheet if you have questions. I would like you to checkout "Oustanding questions" as well.

API endpoints that aren't being contested will be opened here in form of PR where further things like descriptions, validation, response codes etc will be discussed.

cc-ing interested parties @paulhauner @hwwhww @arnetheduck @mratsim @terencechain @prestonvanloon @AgeManning @wemeetagain @mkalinin @ajsutton @rolfyone @moles1 @skmgoldin

mpetrunic:
Assumed implementation is server sent events.

Since this is last missing method from WG proposal, this resolves #37 and resolves #25 and resolves #31

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -71,3 +71,4 @@ get:
                 \n
     "500":
       $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
+      
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -1,0 +1,73 @@
+get:
+  operationId: "eventstream"
+  summary: Subscribe to beacon node events
+  tags:
+    - Events
+  description: |
+    Provides endpoint to subscribe to beacon node Server-Sent-Events stream.
+    Consumers should use [eventsource](https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface)
+    implementation to listen on those events.
+  parameters:
+    - name: topics
+      in: query
+      required: true
+      description: Event types to subscribe to
+      schema:
+        type: array
+        items:
+          type: string
+          uniqueItems: true
+          enum:
+            - head
+            - block
+            - attestation
+            - voluntary_exit
+            - finalized_checkpoint
+            - chain_reorg
+  responses:
+    "200":
+      description: Opened SSE stream.
+      content:
+        text/event-stream:
+          schema:
+            type: string
+            description: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
+          examples:
+            head:
+              description: The node has finished processing, resulting in a new head
+              value: |
+                event: head\n
+                data: "{\"slot\": \"10\", \"block\": \"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"state\":\"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch_transition\": false}"\n
+                \n
+            block:
+              description: The node has received a block (from P2P or API)
+              value: |
+                event: block\n
+                data: "{\"slot\": \"10\", \"block\": \"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\"}"\n
+                \n
+            attestation:
+              description: The node has received an attestation (from P2P or API)
+              value: |
+                event: attestation\n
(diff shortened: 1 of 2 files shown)


apis/eventstream/index.yaml
```diff
@@ -0,0 +1,73 @@
...
+              description: Finalized checkpoint has been updated
+              value: |
+                event: finalized_checkpoint\n
+                data: "{\"block\":\"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"state\": \"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch\": \"2\" }"\n
+                \n
+            chain_reorg:
+              description: The node has reorganized its chain
+              value: |
+                event: chain_reorg\n
+                data: "{\"slot\":\"200\", \"depth\": \"50\", \"old_head_state\":\"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"state\": \"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch\": \"2\" }"\n
```
mcdee, line 70: You have `old_head_state` and `state` here, which seems a bit weird.  Would be better for them to either be `old_state` and `state`, or `old_head_state` and `head_state`.

mpetrunic: Didn't you define those fields in WG Proposal sheet? :smile: 
Maybe old_head_state, new_head_state?

Curious as to why state roots and not block roots?

mpetrunic: @mcdee ^^

mcdee: Sorry was going to look back to where I picked state root rather than block root, but I can't find it so either my detective skills are very poor or I made it up; either way I'll take the blame :laughing: 

State roots allow us to fetch items from /beacon/states directly, which seems to be a logical next step when the chain has reorganized.  But if you think that block roots are a good idea in addition we could add them in.

And yes, `old_head_state` and `new_head state` are nicely explicit so let's go with them.

mpetrunic: I'm gonna add block roots since full beacon states are debug endpoint and I don't see I way to determine block heads from current state endpoints?

apis/eventstream/index.yaml
```diff
@@ -0,0 +1,73 @@
+get:
+  operationId: "eventstream"
```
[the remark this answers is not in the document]
mpetrunic: You open stream with as many "topics"(item-types) you would like to follow.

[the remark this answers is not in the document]
djrtwo: got it

PR #74, opened 2020-08-21
Minimal SSV API

dankrad:
Suggested API for VC to communicated with SSV node (SSV node will emulate BN API to VC, to keep things simple)

### DIFF
--- a/apis/validator/ssv/stream.yaml
+++ b/apis/validator/ssv/stream.yaml
@@ -23,6 +23,7 @@ get:
           enum:
             - sign_block
             - sign_attestation
+            - sign_aggregation_proof
             - check_block
             - check_attestation
   responses:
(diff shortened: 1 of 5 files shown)


hsailern (dismissed):

apis/validator/ssv/stream.yaml
```diff
@@ -0,0 +1,86 @@
...
+      schema:
+        type: array
+        items:
+          type: string
+          uniqueItems: true
+          enum:
+            - sign_block
+            - sign_attestation
+            - check_block
+            - check_attestation
```
alonmuroch, line 27: @dankrad  should include sign aggregation as well

dankrad: What do you man signature aggregation? The reconstruction of the threshold signature should be done by the SSV node, the VC doesn't need to do anything about it

alonmuroch: if that validator has an aggregation duty for a particular epoch we should be able to sign the proof.

dankrad: Ah right, I forgot about attestation aggregation proofs. Yes, these need to be added then.

axljoker03 (approved):

rolfyone:
Closing as very stale, we can revisit if it comes back around...

PR #76, opened 2020-08-31
Add root field to /beacon/headers/{block_id}

paulhauner:
I noticed this field is missing in `beacon/headers/{block_id}` but present in `beacon/headers`. I assume it's supposed to be in both places, please correct me if I'm wrong :)


mpetrunic (approved):
Yeah, I think it makes sense.

PR #80, opened 2020-09-08
make ENR nullable in peers response

rolfyone:
in line with a discussion on discord, Teku is not currently able to supply the ENR for peers, because libp2p is kept quite separate from discovery.

There's still a lot of helpful context in the peers list without ENR, so making ENR nullable will keep everything 'correct'.

### DIFF
--- a/types/p2p.yaml
+++ b/types/p2p.yaml
@@ -39,7 +39,9 @@
       peer_id:
         $ref: "./p2p.yaml#/PeerId"
       enr:
-        $ref: "./p2p.yaml#/ENR"
+        allOf:
+          - $ref: "./p2p.yaml#/ENR"
+          - nullable: true
       address:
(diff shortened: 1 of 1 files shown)


types/p2p.yaml
```diff
@@ -58,6 +58,7 @@
 
   ENR:
     type: string
+    nullable: true
```
mpetrunic, line 61: Can you add this to
Peer.properties.peer_id.enr since this enr type might be used in some other api endpoints where it isn't necessary nullable.

Syntax is:
```
enr:
   allOf:
      - $ref: "./p2p.yaml#/ENR"
      - nullable: true
```

rolfyone: done. looked at the local server and it seems to mark both endpoints nullable, so looks ok.

mpetrunic (approved):

PR #82, opened 2020-09-09
Add validator balances endpoint

mcdee:
This patch adds the validator balances endpoint `/eth/v1/beacon/states/{state_id}/validator_balances`  Its purpose is to fetch validator balance at a given epoch separate from the entire validator structure.

Rationale behind this being a separate endpoint is that historical data on a validator is, in general, not a lot of use; the current state for validators provides all of the prior state of interest for most use cases.  Having to fetch or recreate the entire validator structure for multiple epochs, when all a user wants is to obtain a history of their balance, can lead to a lot of wasted CPU time (at least it does with the current non-standard APIs on beacon node implementations).

### DIFF
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -43,7 +43,7 @@ get:
               data:
                 type: array
                 items:
-                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorBalancesResponse'
+                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorBalanceResponse'
     "500":
       $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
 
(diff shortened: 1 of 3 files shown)


apis/beacon/states/validator_balances.yaml
```diff
@@ -0,0 +1,49 @@
...
+      in: query
+      required: false
+      schema:
+        type: array
+        maxItems: 30
+        uniqueItems: true
+        items:
+          description: "Either hex encoded public key (with 0x prefix) or validator index"
+          type: string
+    - name: status
```
mpetrunic, line 22: do we need this? It requires getting validator and we won't have it in response?

mcdee: I kept the same parameters as for `/validators`.  It's tricky: on the one hand, I can see that it has use (_e.g._ I want the balances of all active validators).  On the other, as you say, it would involve knowing more about the validator and possibly invalidating the whole purpose of the endpoint.

Shall we remove it for now?  We can always add it back at a later date if people need it (and we can find a way to do it sensibly).

mcdee: @mpetrunic I removed the status part of the request.

mpetrunic (approved):

djrtwo (approved):

PR #84, opened 2020-09-10
Change validator.slashed to validator.slashed_epoch

guybrush:
Currently there is `GetStateValidatorResponse.data.status` which indicates if a validator is slashed or not and also there is `GetStateValidatorResponse.data.validator.slashed` which is redundant information.

I think it is better to use `GetStateValidatorResponse.data.validator.slashed_epoch` just like `GetStateValidatorResponse.data.validator.{exit,withdrawable,activation,activation_eligibility}_epoch`.


guybrush:
Ok I guess it is better to follow the spec. If at all it would make more sense to change in spec first.

PR #86, opened 2020-09-15
Add validator index to proposer duties.

mcdee:
This adds the validator index to proposer duty return information, so that similar to attester duties they can be referenced by index rather than public key if desired.


rolfyone (approved):
LGTM

mpetrunic (approved):

PR #89, opened 2020-09-21
Add return code 400 for malformed input

mcdee:
This adds a response code 400 for all endpoints that have an input.  It allows it to be returned when the input data is in some way malformed.

Note that there is a difference between this and 404.  If you asked for block 5 and block 5 doesn't exist you would expect a 404.  If you asked for block foo you would expect a 400, as 'foo' is not a valid block number.

### DIFF
--- a/apis/beacon/blocks/attestations.yaml
+++ b/apis/beacon/blocks/attestations.yaml
@@ -28,7 +28,7 @@ get:
         application/json:
           schema:
             allOf:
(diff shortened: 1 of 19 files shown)


mpetrunic (approved):

rolfyone (approved):
LGTM

mpetrunic:
I think you have invalid reference somewhere. Probably to many `../` somewhere

mpetrunic (approved):

djrtwo (approved):

drumenbassik (approved):

PR #94, opened 2020-10-06
Update validator status

mpetrunic:
- update validator status according to new spec: https://hackmd.io/bQxMDRt1RbS1TLno8K4NPg
- removed validator status groups (it's simpler to just define all statuses)
- didn't include next_state_epoch but included description how to get it for each status

### DIFF
--- a/types/api.yaml
+++ b/types/api.yaml
@@ -31,19 +31,31 @@ ValidatorBalanceResponse:
 ValidatorStatus:
   description: |
     Possible statuses:
-    - **pending_initialized** - When the first deposit is processed, but not enough funds are available (or not yet the end of the first epoch) to get validator into the activation queue.
-    - **pending_queued** - When validator is waiting to get activated, and have enough funds etc. while in the queue, validator activation epoch keeps changing until it gets to the front and make it through (finalization is a requirement here too).
-    - **active_ongoing** - When validator must be attesting, and have not initiated any exit.
-    - **active_exiting** - When validator is still active, but filed a voluntary request to exit.
-    - **active_slashed** - When validator is still active, but have a slashed status and is scheduled to exit.
-    - **exited_unslashed** - When validator has reached reguler exit epoch, not being slashed, and doesn't have to attest any more, but cannot withdraw yet.
-    - **exited_slashed** - When validator has reached reguler exit epoch, but was slashed, have to wait for a longer withdrawal period.
-    - **withdrawal_possible** - After validator has exited, a while later is permitted to move funds, and is truly out of the system.
-    - **withdrawal_done** - (not possible in phase0, except slashing full balance) - actually having moved funds away
+    - **waiting_for_eligibility** - A `Deposit` is included in a `BeaconBlock`, causing the Beacon Chain to become aware of the validator.
+    - **waiting_for_finality** - The Beacon Chain has observed that the validator could become eligible and it is waiting for those conditions 
+    to become finalized.
+    - **waiting_in_queue** - The conditions which allow the validator to become active are finalized.
+    The validator is now waiting in the activation queue, awaiting its turn to be activated.
+    - **standby_for_active** - The conditions which allow the validator to become active are finalized.
+    The validator is now waiting in the activation queue, awaiting its turn to be activated.
+    Validator status will change to `active` on `validator.activation_epoch`.
+    - **active** - The validator is now active and, for the first time, is rewarded for including attestations and blocks in the beacon chain.
+     It will be penalized if it fails to have attestations included in the chain.
+    - **active_awaiting_slashed_exit** - The validator has produced conflicting messages and they have been included
+     in the chain via a ProposerSlashing or AttesterSlashing. It will not be rewarded for producing attestations but may be 
+     required to (and rewarded for) producing blocks. Validator will change status to `exited_slashed` on `validator.exit_epoch`.
+    - **active_awaiting_voluntary_exit** - The validator has voluntarily started exiting. As with the active state, the
+    validator will be rewarded for including attestations and blocks in the chain
+    and penalized for failing to include attestations. Validator will change status to `exited_voluntarily` on `validator.exit_epoch`.
+    - **exited_voluntarily** - The validator has exited and is no longer permitted to include attestations or
+    blocks in the beacon chain. Validator will change status to `withdrawable` on `validator.withdrawable_epoch`.
+    - **exited_slashed** - The validator has exited and is no longer permitted to include attestations or
+    blocks in the beacon chain. Validator will change status to `withdrawable` on `validator.withdrawable_epoch`.
+    - **withdrawable** - The validator has exited and waited long enough to be withdrawable.
 
-    [Validator status specification](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ)
-  enum: ["pending_initialized", "pending_queued", "active_ongoing", "active_exiting", "active_slashed", "exited_unslashed", "exited_slashed", "withdrawal_possible", "withdrawal_done"]
-  example: "active_ongoing"
+    [More indepth validator status specification](https://hackmd.io/bQxMDRt1RbS1TLno8K4NPg)
+  enum: ["waiting_for_eligibility", "waiting_for_finality", "waiting_in_queue", "standby_for_active", "active", "active_awaiting_slashed_exit", "active_awaiting_voluntary_exit", "exited_voluntarily", "exited_slashed", "withdrawable"]
+  example: "active"
 
 
 Committee:
--- a/apis/beacon/states/validators.yaml
+++ b/apis/beacon/states/validators.yaml
@@ -20,16 +20,14 @@ get:
           description: "Either hex encoded public key (with 0x prefix) or validator index"
           type: string
     - name: status
-      description: "[Validator status specification](https://hackmd.io/ofFJ5gOmQpu1jjHilHbdQQ)"
+      description: "[Validator status specification](https://hackmd.io/bQxMDRt1RbS1TLno8K4NPg)"
       in: query
       required: false
       schema:
         type: array
         uniqueItems: true
         items:
-          allOf:
-            - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorStatus'
-            - enum: ["active", "pending", "exited", "withdrawal"]
+          $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorStatus'
 
   responses:
     "200":


mpetrunic:
cc @protolambda @paulhauner

mcdee:
What happens if the validator's balance drops below 16 Ether?  I don't see either `ExitedVoluntarily` or `ExitedSlashed` describing this state.

I'm also a little concerned that the names quite verbose yet unclear regarding sequence.  Would you be able to order `WaitingForEligibility`, `WaitingForFinality` and `WaitingInQueue` without looking at their definitions?  Can you go from waiting for finality to waiting for eligibility in case of a reorg?

Providing the transition epoch is useful in terms of standardizing the information (although we should include this when waiting in the queue, as a common user question is "when will my validator start?"). 

The additional states are welcome, but the simplicity of the existing `group_state` definitions is useful.  For example, what does `ActiveAwaitingVoluntaryExit` gain over the existing `active_exiting` state?  I would prefer to see this format retained, as it is easier to obtain the high-level state of a validator (pending,active,exited,withdrawable) without having to know all of the substates.

guybrush:
Just for convenience here is an overview (changed `withdrawable` to `withdrawable_{voluntary,slashed}` and added `withdrawn_{voluntary,slashed}`)

```
proposal1            proposal2                       info
-------------------  ------------------------------  ------------------------------------------------------
pending_initialized  waiting_for_eligibility         The validator does not have enough funds.
                     waiting_for_finality            Once the chain finalizes, the validator will become eligible.
pending_queued       waiting_in_queue                The validator is eligible to be activated, however it must wait for other validators to activate first.
                     standby_for_active              The validator has reached the front of the activation queue and has been assigned an activation epoch (next_state_epoch).
active_ongoing       active                          The validator is active, unslashed and has not had an exit included in the chain. The validator is expected to perform duties until they exit voluntarily or are slashed.
active_exiting       active_awaiting_slashed_exit    The validator is active, but a SignedVoluntaryExit has been included on-chain. The validator is expected to perform duties until next_state_epoch.
active_slashed       active_awaiting_voluntary_exit  The validator is active, but it has been slashed. The validator is expected to perform duties until next_state_epoch.
exited_unslashed     exited_voluntarily              The validator exited voluntarily and is no longer required to perform duties. They will become withdrawable at next_state_epoch.
exited_slashed       exited_slashed                  The validator was slashed and exited by the protocol. They will become withdrawable at next_state_epoch.
withdrawal_possible  withdrawable_voluntarily        The validator is eligible to withdraw thier balance.
                     withdrawable_slashed            The slashed validator is eligible to withdraw thier balance.
withdrawal_done      withdrawn_voluntarily           The validator has withdrawn their ETH. This is not possible in phase 0.
                     withdrawn_slashed               The slashed validator has withdrawn their ETH. This is not possible in phase 0.
```

I will use proposal2 in our database for now

djrtwo:
@guybrush your descriptions of `active_awaiting_slashed_exit` and `active_awaiting_voluntary_exit` are flipped

guybrush:
> @guybrush your descriptions of `active_awaiting_slashed_exit` and `active_awaiting_voluntary_exit` are flipped

updated in my original comment

djrtwo (commented):
Seems like getting forcefully exited via an ejection isn't accounted for as @mcdee noted

mpetrunic:
@paulhauner would you like to chip in and/or update validator status document before we make changes here?

paulhauner:
### Responses

> What happens if the validator's balance drops below 16 Ether? I don't see either ExitedVoluntarily or ExitedSlashed describing this state.

This is a good point, I can't see any way to determine the difference between a voluntary exit and a low-balance exit without doing expensive historical searches.

It seems we need to group voluntary and low-balance exits.

>  the simplicity of the existing group_state definitions is useful

I find the distinction between "states" and "group states" to be unclear. There's nowhere I can see (except via bullet-point nesting in the HackMD) that states what's a "state" and what's a "group state". Here's two questions that I think identify the ambiguity:

- *"Why do I get `active_ongoing` when I query with `active`?"*
- *"Why are none of my validators `active`?"*

I think we need to be explicit about groups and have separate fields and queries for "states" and "group states". It is more verbose, but I think the API is currently ambiguous rather than concise.

### Suggestions

1. Add a `status_group` field to the `Validator` response that includes @mcdee's groups (not attached to name of field),
1. Remove the link to my HackMD, since it's unlikely to be maintained long term (not due to lack of caring, just being realistic).
1. Remove the "voluntarily" qualifications (see table below)

#### Updated status table

I've updated @guybrush's handy table as per my suggestions.

```
proposal1            proposal2                       status_group     info
-------------------  -----------------------------  ------            ------------------------------------------------
pending_initialized  waiting_for_eligibility        pending           The validator does not have enough funds.
                     waiting_for_finality           pending           Once the chain finalizes, the validator will become eligible.
pending_queued       waiting_in_queue               pending           The validator is eligible to be activated, however it must wait for other validators to activate first.
                     standby_for_active             pending           The validator has reached the front of the activation queue and has been assigned an activation epoch (next_state_epoch).
active_ongoing       active                         active            The validator is active, unslashed and has not had an exit included in the chain. The validator is expected to perform duties until they exit voluntarily or are slashed.
active_exiting       active_awaiting_exit           active            The validator is active, but a SignedVoluntaryExit has been included on-chain or the validator was ejected due to low balance. The validator is expected to perform duties until next_state_epoch.
active_slashed       active_awaiting_slashed_exit   active            The validator is active, but it has been slashed. The validator is expected to perform duties until next_state_epoch.
exited_unslashed     exited                         exited            The validator exited voluntarily or due to low balance and is no longer required to perform duties. They will become withdrawable at next_state_epoch.
exited_slashed       exited_slashed                 exited            The validator was slashed and exited by the protocol. They will become withdrawable at next_state_epoch.
withdrawal_possible  withdrawable                   withdrawal        The validator is eligible to withdraw thier balance.
                     withdrawable_slashed           withdrawal        The slashed validator is eligible to withdraw thier balance.
withdrawal_done      withdrawn                      withdrawal        The validator has withdrawn their ETH. This is not possible in phase 0.
                     withdrawn_slashed              withdrawal        The slashed validator has withdrawn their ETH. This is not possible in phase 0.

```

guybrush:
I put more thought into it and I think having the status in the api is redundant information. But it introduces names for state-conditions and putting these names into the api makes everyone use the same words for the same things.

Here is another proposal for status-names, this time with corresponding conditions of the state (against spec-variables):
```
validator_status          activation_epoch  withdrawable_epoch  exit_epoch            effective_balance      slashed
----                      ----              ----                ----                  ----                   ----
pending_depositing        >  CURRENT_EPOCH  >  CURRENT_EPOCH    == FAR_FUTURE_EPOCH   <  MIN_DEPOSIT_AMOUNT  f
pending_deposited         >  CURRENT_EPOCH  >  CURRENT_EPOCH    == FAR_FUTURE_EPOCH   >= MIN_DEPOSIT_AMOUNT  f
pending_initialized       >  CURRENT_EPOCH  >  CURRENT_EPOCH    == FAR_FUTURE_EPOCH   >= MIN_DEPOSIT_AMOUNT  f
pending_queued            >  CURRENT_EPOCH  >  CURRENT_EPOCH    == FAR_FUTURE_EPOCH   >= MIN_DEPOSIT_AMOUNT  f
active_ongoing            <= CURRENT_EPOCH  >  CURRENT_EPOCH    == FAR_FUTURE_EPOCH   >  EJECTION_BALANCE    f
active_exiting            <= CURRENT_EPOCH  >  CURRENT_EPOCH    <  FAR_FUTURE_EPOCH   >  EJECTION_BALANCE    f
active_slashed            <= CURRENT_EPOCH  >  CURRENT_EPOCH    <  FAR_FUTURE_EPOCH   >  EJECTION_BALANCE    t
active_ejected            <= CURRENT_EPOCH  >  CURRENT_EPOCH    <  FAR_FUTURE_EPOCH   <= EJECTION_BALANCE    f
exited_voluntarily        <= CURRENT_EPOCH  >  CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    f
exited_slashed            <= CURRENT_EPOCH  >  CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    t
exited_ejected            <= CURRENT_EPOCH  >  CURRENT_EPOCH    <= CURRENT_EPOCH      <= EJECTION_BALANCE    f
withdrawable_voluntarily  <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    f
withdrawable_slashed      <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    t
withdrawable_ejected      <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      <= EJECTION_BALANCE    f
withdrawn_voluntarily     <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    f
withdrawn_slashed         <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    t
withdrawn_ejected         <= CURRENT_EPOCH  <= CURRENT_EPOCH    <= CURRENT_EPOCH      <= EJECTION_BALANCE    f
```

paulhauner:
```
withdrawable_voluntarily  <= CURRENT_EPOCH  <= CURRENT_EPOCH              <= CURRENT_EPOCH    <= CURRENT_EPOCH      >  EJECTION_BALANCE    f        The validator is eligible to withdraw thier balance.
...
withdrawable_ejected      <= CURRENT_EPOCH  <= CURRENT_EPOCH              <= CURRENT_EPOCH    <= CURRENT_EPOCH      <= EJECTION_BALANCE    f        The ejected validator is eligible to withdraw thier balance.
```

This is not strictly true, since it's possible to be ejected due to low balance and then submit a deposit to increase your `effective_balance` above `EJECTION_BALANCE`.

guybrush:
Thank you very much! I was not aware of this.

paulhauner:
@guybrush I think I may have missed your point, but my point was that was cannot* distinguish between a voluntarily/ejected validator. As such, I would say that the API should not make a distinction between these two states.

*: It can be determined by searching back through the history of the chain, but this is too computationally expensive for this endpoint, in my opinion.

mpetrunic:
Whats the conclusion here, should I update statuses with @paulhauner last update: https://github.com/ethereum/eth2.0-APIs/pull/94#issuecomment-708110821 ?
I guess I should pull md file inside repo?
Should we get groups back or keep them removed? What's client implementation status on this endpoint?

mcdee:
Regarding the states, I'm fine with Paul's states except for the last three.  As we do not know how withdrawals are going to occur we don't know if there are extra states that will go in there (`withdrawal_pending`, that type of thing).  We also don't care if a validator is slashed or not at the point that it is withdrawable, as there is no difference in functionality.

Because of the above I would drop the last three states (`withdrawable_slashed`, `withdrawn` and `withdrawn_slashed`) and perhaps change the last group state to "done", "finished" or something else that gives us a bit of wriggle room when it comes to implementing withdrawals.

paulhauner:
> "done", "finished" or something else that gives us a bit of wriggle room when it comes to implementing withdrawals.

I'm happy to remove the `withdrawn` states, since we can't reach them. But I would advise against the "done" or "finished" states; they have no meaning beyond inferring that another state is impossible. Since we do not know how withdrawals are going to occur, I would avoid making any statements about it.

mcdee:
> ...I would advise against the "done" or "finished" states...

We can leave the group state at "withdrawal", then.

types/api.yaml
```diff
@@ -31,19 +31,31 @@ ValidatorBalanceResponse:
...
+    - **waiting_for_finality** - The Beacon Chain has observed that the validator could become eligible and it is waiting for those conditions 
+    to become finalized.
+    - **waiting_in_queue** - The conditions which allow the validator to become active are finalized.
+    The validator is now waiting in the activation queue, awaiting its turn to be activated.
+    - **standby_for_active** - The conditions which allow the validator to become active are finalized.
+    The validator is now waiting in the activation queue, awaiting its turn to be activated.
+    Validator status will change to `active` on `validator.activation_epoch`.
+    - **active** - The validator is now active and, for the first time, is rewarded for including attestations and blocks in the beacon chain.
+     It will be penalized if it fails to have attestations included in the chain.
+    - **active_awaiting_slashed_exit** - The validator has produced conflicting messages and they have been included
```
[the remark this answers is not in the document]
djrtwo: Note, should still remove the part about producing blocks. Slashed vals can't validly produce a block

types/api.yaml
```diff
@@ -31,19 +31,31 @@ ValidatorBalanceResponse:
...
+     in the chain via a ProposerSlashing or AttesterSlashing. It will not be rewarded for producing attestations but may be 
+     required to (and rewarded for) producing blocks. Validator will change status to `exited_slashed` on `validator.exit_epoch`.
+    - **active_awaiting_voluntary_exit** - The validator has voluntarily started exiting. As with the active state, the
```
djrtwo, line 47: @paulhauner One thing to note about your DAG is that you can *kind of* go from `active_awaiting_voluntary_exit` to `active_awaiting_slashed_exit`. Well I suppose you voluntarily exited but are now slashed..

Similarly, you can be `exited_voluntarily` and then slashed. Which what state does that put you in? `exited_slashed`? In such a case, the DAG should be updated to account for these scenarios

djrtwo: A couple of note about "8. `exited_volutarily` & `exited_slashed`" in your doc. 

> This withdrawable epoch was decided by the protocol when the exit was initiated.

This is true in many cases, but if the validator first sent a voluntary exit and then was slashed, the withdrawable epoch would be updated to be longer once the val is slashed. So this epoch is actually decided "when the exit is initiated or the validator slashed, whichever was the latest of the two actions"

> The validator will remain in this state for 256 epochs (~27 hours) before
progressing to the next state.

This is the case when a non-slashed exit. Otherwise, the validator is not withdrawable until `slashing_epoch  + EPOCHS_PER_SLASHINGS_VECTOR` (36 days).

djrtwo: I know those are a bit nit-picking but I want to better understand how you plan to represent these cases so we can try to conform on the edges

djrtwo (commented):
Awesome! 

@paulhauner I have a few questions about corner cases related to slashing. keen to hear your input

djrtwo:
I had a couple of lingering questions on slashed related states and  the sequence of transition
https://github.com/ethereum/eth2.0-APIs/pull/94#discussion_r520957680

mcdee:
Do we want to proceed with this?  The API is basically out of the door in its V1 state now that mainnet is launched, and although there are benefits to this it will no doubt cause confusion and problems as beacon nodes transition from one to the other at an uneven rate.

Perhaps push this to V2, whenever that may be?

mpetrunic:
That might be good idea. In V2 of api we can add that under /v2 prefix and gradually deprecate

mpetrunic:
don't seems this will move forward

PR #96, opened 2020-10-10
Add "state" event

mcdee:
This PR unifies the existing "head" and "chain_reorg" events in to a single "state" event.

The rationale behind changing "head" to "state" is that the state can change for multiple reasons other than a new head.  For example, an epoch transition can occur without a block being received and it is important that clients receive a notification that this has occurred.

The rationale behind integrating the "chain_reorg" event in to "state" is that any chain reorganization will, by definition, result in a new state so would trigger both a "chain_reorg" and a "state" event.  For a client to successfully understand what has happened it would need both events, but given they are asynchronous this would lead to complexity for client developers.  Providing a single removes that complexity.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -20,12 +20,10 @@ get:
           uniqueItems: true
           enum:
             - state
-            - head
             - block
             - attestation
             - voluntary_exit
             - finalized_checkpoint
-            - chain_reorg
   responses:
     "200":
       description: Opened SSE stream.
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -19,6 +19,7 @@ get:
           type: string
           uniqueItems: true
           enum:
+            - state
             - head
             - block
             - attestation
@@ -34,12 +35,30 @@ get:
             type: string
             description: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
           examples:
-            head:
-              description: The node has finished processing, resulting in a new head
+            state:
+              description: |
+                The node has updated its state.  This event will be sent in three situations:
+
+                - a node has received a block, and updated state accordingly
+                - a node has _not_ received a block four seconds after the start of the slot, and updated state accordingly
+                - a node has carried out a chain reorganization, and updated state accordingly
+
+                The fields returned in the data for this event are:
+
+                  - **slot** the current slot as understood by the node
+                  - **block** the block root of the chain head as understood by the node
+                  - **state** the state root as understood by the node
+                  - **distance** the number of slots between the slot of the chain head and the current slot as understood by the node
+                  - **epoch_transition** _true_ if the slot marks the first slot in an epoch, otherwise _false_
+                  - **reorg** _true_ if the state has updated due to a reorganization, otherwise _false_
+                  - **reorg_distance** the number of slots between the slot of the block that caused a reorganization and the current slot as understood by the node
+
+                On receipt of this event a client can query the node for information about the given state using the provided state root.
+
+                Note that it is possible to receive multiple events for the same slot, when multiple situations as above occur within a single slot.  It is also possible to receive events for slots that are not the current slot, for example if the node is syncing.  The user should compare information about the slot returned by this event with the expected slot
               value: |
-                event: head\n
-                data: "{\"slot\": \"10\", \"block\": \"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"state\":\"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch_transition\": false}"\n
-                \n
+                event: state
+                data: {"slot": "10", "block": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "distance: "0", "epoch_transition": false, "reorg": false, "reorg_distance:" "0"}
             block:
               description: The node has received a block (from P2P or API)
               value: |
@@ -63,12 +82,6 @@ get:
               value: |
                 event: finalized_checkpoint\n
                 data: "{\"block\":\"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"state\": \"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch\": \"2\" }"\n
-                \n
-            chain_reorg:
-              description: The node has reorganized its chain
-              value: |
-                event: chain_reorg\n
-                data: "{\"slot\":\"200\", \"depth\": \"50\", \"old_head_block\": \"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"new_head_block\": \"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76\", \"old_head_state\":\"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf\", \"new_head_state\": \"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9\", \"epoch\": \"2\" }"\n
                 \n
     "400":
       description: "The topics supplied could not be parsed"


apis/eventstream/index.yaml
```diff
@@ -34,12 +33,30 @@ get:
...
+                - a node has received a block, and updated state accordingly
+                - a node has _not_ received a block four seconds after the start of the slot, and updated state accordingly
+                - a node has carried out a chain reorganization, and updated state accordingly
+
+                The fields returned in the data for this event are:
+
+                  - **slot** the current slot as understood by the node
+                  - **block** the block root of the chain head as understood by the node
+                  - **state** the state root as understood by the node
+                  - **distance** the number of slots between the slot of the chain head and the current slot as understood by the node
```
ajsutton, line 49: Is this essentially the number of empty slots after the head block?

apis/eventstream/index.yaml
```diff
@@ -34,12 +33,30 @@ get:
...
+                  - **epoch_transition** _true_ if the slot marks the first slot in an epoch, otherwise _false_
+                  - **reorg** _true_ if the state has updated due to a reorganization, otherwise _false_
+                  - **reorg_distance** the number of slots between the slot of the block that caused a reorganization and the current slot as understood by the node
```
ajsutton, line 52: This description of reorg_distance doesn't make sense to me - the reorg isn't necessarily caused by a new block being received, it may be because new attestations were received. So I'm a bit unclear what this value would actually be.

Validator clients need to determine the slot of the common ancestor block so this would either need to be that directly or the distance from the `slot` value back to the common ancestor slot.  That allows the client to invalidate only duties that are actually affected by the change instead of all of them.

ajsutton (commented):
Combining head and reorg events makes a lot of sense to me. Just need to make sure the slot of the common ancestor can be calculated from a reorg.

apis/eventstream/index.yaml
mcdee: Yes.  I'm using "chain head" to mean the same thing as "head block"; can change the wording if you think it's unclear.

apis/eventstream/index.yaml
mcdee: The idea of the reorg distance is, as you say, the distance from the current block back to the common ancestor slot.  "slot of the block that caused the a reorganization" tries to explain it, but obviously I'm not doing a great job there.  Happy to take suggestions on how to clarify this.

apis/eventstream/index.yaml
ajsutton: I think chain head is fine here as long as we're consistent in that usage.  We may need a glossary at some point to define some key terms in one place.

apis/eventstream/index.yaml
ajsutton: I think the key term is common ancestor block so I'd probably describe it as something like:

```
   - **reorg_distance** the number of slots which were changed by this reorg. The common ancestor of the old and new forks is therefore at slot `slot - reorg_distance`.
```

Though it feels like it would be simpler to explain and use if it just included the ancestor slot directly.

apis/eventstream/index.yaml
```diff
@@ -34,12 +33,30 @@ get:
...
             description: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
           examples:
-            head:
-              description: The node has finished processing, resulting in a new head
+            state:
+              description: |
+                The node has updated its state.  This event will be sent in three situations:
+
+                - a node has received a block, and updated state accordingly
+                - a node has _not_ received a block four seconds after the start of the slot, and updated state accordingly
```
mpetrunic, line 41: Is this per spec? I don't think we should include this, as it depends on how each client handles this. 

It makes a lot more sense to introduce separate clock:slot/slot event so vc can listen on start of beacon chain slot.
We had some issues where VC clock is few microseconds ahead of Bn clock so it asks for duties and other actions a bit prematurely.

mpetrunic (commented):
TBH, I'm more in favour of separate chain_head and reorg event where we introduce number of skip slots to chain_head and emit reorg only when we change fork branch rather then emitting it every time we have skip slot

apis/eventstream/index.yaml
mcdee: There was quite a lengthy discussion on this in the Discord starting at https://discordapp.com/channels/595666850260713488/710831784991916072/760955856165797890 but to focus on the first point of `chain_head` Vs. `state`: the discussion was around the fact that state can change even though the chan head has not (_e.g._ in the case of skipped slots).  In this situation, clients need to be made aware that state has updated hence the move from head to state.

The other point of the `state` event is to subsume the requirement for clock events.  The clock event basically says "I'm ready to start serving for state _x_", so we're providing much the same information as we already do in the state event.  However, if an implementation is not ready to serve for state _x_ as soon as that slot starts (for example, if it is carrying out internal accounting for the epoch transition) then sending a clock event has no value.  I think we're talking about much the same thing with clock and state events, it's just that the state event covers instances where beacon nodes are not immediately ready at a given clock tick.

One thing:

> ...vc can listen on start of beacon chain slot and decide on it's own how long it wants to wait.

A major advantage of the state event is that the VC doesn't have to do this guess; as soon as it receives the state event it knows it is good to query the BN for information related to that state.

mpetrunic: > the discussion was around the fact that state can change even though the chan head has not (e.g. in the case of skipped slots)

But it cannot really. that's just client optimization that may or may or may not payoff? 
There could be block even though client didn't receive it so client might do useless epoch transition which not every client can afford :smile: 

> The clock event basically says "I'm ready to start serving for state x"

First chain_head is marking that not slot event, I don't think client should calculate states ahead of chain_head and returning it, that should be error 404.
Second, we can add doc saying that this just marks start of slot in BN clock.

mcdee: Perhaps worth going through a specific situation in more detail.

The first functional sentence of the validator flow doc is "On start of every epoch, validator should fetch proposer duties.".  Given we cannot assume that the validator client and beacon node clocks are in sync, the beacon node should indicate to the validator client when it thinks the epoch has started so that the validator client can fetch the proposer duties.  This is one of the clock events that have been discussed.

However, a simple clock event is not enough.  This is because the some beacon nodes do not update their state until the epoch starts.  This means that there is a time after the start of the epoch when the beacon node will return the previous epoch's information.  So we need some sort of event to say not just "the epoch has started" but "I'm ready to serve data for the epoch".  For some beacon node implementations this may be the same thing, but for others it may not.  Either way, though, we can use a single `state` event to notify clients that the beacon node is ready to serve.

Does this help to clarify where the `state` event would be used, and how it combines features of the current `head` and discussed `clock` events?

(We can discuss the other features of the `state` event separately, but I wanted to try to lay some common ground rather than attempt to tackle all of the different points at once).

paulhauner:
> The rationale behind integrating the "chain_reorg" event in to "state" is that any chain reorganization will, by definition, result in a new state so would trigger both a "chain_reorg" and a "state" event.

The `state` event is significantly more onerous to produce than the `chain_reorg` event, since `state` can require an unbounded amount of skip-slots transitions (and DB writes) whilst `chain_reorg` requires none.

> Providing a single removes that complexity.

I would argue that this endpoint *increases* complexity for implementer (that's the case for LH).

From a broad perspective, I think there's an assumption here that nodes have "a state" which is kept up-to-date with the wall clock slot. This is not the case for Lighthouse; it has a bunch of head states (where one is canonical) and then a bunch of caches to serve requests to those states. We skip slots *on-demand*, and reduce that demand through caching. This is a perfectly valid way to operate, since it reduces memory usage and reduces the cost of bouncing between heads (no need to skip the new state to the current slot if it's going to be reverted again before you use it). Also, it turns out that you can key your caches such that you can serve multiple states from disparate heads via the same cache entry, dramatically reducing the load on the client.

As such, this new event involves adding the following components/tasks to LH:

1. A timer to fire 4s after each slot starts.
1. A routine that can scan all the heads and determine if some block has been received for some slot.
1. Additional state transitions for each skip slot between the head and the current slot.
1. Additional DB writes to store the above states (I assume if we're going to report the state root to users, then they should also be able to get it from the DB).

(1) and (2) only really require additional development hours, whilst (3) and (4) require dev hours *and* incur a rather significant runtime cost. 

>  if an implementation is not ready to serve for state x as soon as that slot starts

As I understand it, this `state` event is catering for the client that is unable to serve data for some slot by the time that slot has arrived.

I would argue that such client should just compute the state inside the API call (priming their caches there, instead of on the start of a new slot) rather than adding complexity to all other clients. This improves the UX of their API and means that other clients don't have to act as if they also have such a limitation as well.

> However, a simple clock event is not enough. 

In my opinion, a simple wall-clock slot event is perfect. The user can wait 4s (or whatever they want, no need for the BN to make this decision for them) and then poll for duties if they want.

My preference aligns with @mpetrunic's suggestion: maintain the current `head` and `chain_reorg` methods and, if desired, add a `slot` and/or `epoch` event.

apis/eventstream/index.yaml
paulhauner: Using 4 seconds here makes two assumptions:

- That 4 seconds is relevant to the user (it's the case if they're producing attestations, but might not always be the case)
- The client is able to perform the required skip-slots within 4s.

paulhauner:
Alternatively, my concerns would be alleviated by removing:

```
                - a node has _not_ received a block four seconds after the start of the slot, and updated state accordingly
```

This is where we have to do an unbounded and potentially unnecessary state skip and store. I have no fundamental objections to combining `head` and `chain_reorg`, it's just the 4s condition which adds unnecessary assumptions (e.g., 4 seconds, internal caching methods) and unnecessary work (skipping and storing states).

ajsutton:
I don't overly mind whether `head` and `chain_reorg` are combined.  The key reasoning behind getting an event for empty slots is to provide a guarantee from the beacon node that it will notify the Validator Client if a later chain reorg affects that slot.  The vc may need to recalculate duties because of that switch from empty to filled block.  https://hackmd.io/kutQ7smJRZ-sJNSuY1WWVA#Why-head-must-be-published-for-empty-slots sets out the scenario.

In Teku's case we track the head states much like Lighthouse does and fire the `head` event for an empty slot without actually processing the empty slot, but we track that we sent it so if a block for that slot later arrives we know to fire a reorg event. 

An `epoch` event would also work as long as it also provided the guarantee that a `chain_reorg` event would fire if anything happened that would affect scheduling of duties (either switching to another fork or importing a block on the current canonical chain prior to the epoch start).

paulhauner:
> The key reasoning behind getting an event for empty slots is to provide a guarantee from the beacon node that it will notify the Validator Client if a later chain reorg affects that slot. 

As discussed offline with @ajsutton, this is how I propose we address this https://github.com/ethereum/eth2.0-APIs/issues/116

mcdee:
> As I understand it, this `state` event is catering for the client that is unable to serve data for some slot by the time that slot has arrived.
> 
> I would argue that such client should just compute the state inside the API call (priming their caches there, instead of on the start of a new slot) rather than adding complexity to all other clients. This improves the UX of their API and means that other clients don't have to act as if they have such a limitation as well.

I agree with this, however I'm not sure that this matches the state of current clients.  For example, in lighthouse there is:

```
/// Delay this period of time after the slot starts. This allows the node to process the new slot.   
const TIME_DELAY_FROM_SLOT: Duration = Duration::from_millis(100);                                   
```

and, anecdotally, if I hit prysm as soon as an epoch passes I can obtain incorrect attester duties, so assuming that something similar is going on there.

If all clients can respond for the current slot as soon as the beacon node's clock hits the start of that slot then it is true that all we need are clock events, and I'm fine with that.  The point of having a "ready to serve" event (doesn't have to be the state event, if we go with #116) is that it avoids validator client code that goes "receive event; wait a bit", as the "a bit" is always going to be problematic one way (wait too long, strain the network) or the other (wait too short, obtain incorrect duties).

djrtwo:
Where do we stand here? Is this conflicting with @paulhauner's proposal in #116?

mcdee:
> Where do we stand here? Is this conflicting with @paulhauner's proposal in #116?

Partly, yes.  The requirement for a `state` event every time state changes is no longer necessary with the addition of dependent roots in the `head` event.  The other purpose of the event, though, was to act as a "beacon node ready to serve for this state", which isn't addressed.  There has been debate on if this is necessary, which I think the current issues with proposals and attestations for slot 0 of epoch suggest there is, but we may want to hold off until we have some resolution to that issue and understand if this is a resultant requirement.

I'm fine if this PR is closed until the point at which we have more clarity on the issue.  If we do need an event it will be more focused, and I can reference the relevant pieces of the conversation here in the new PR.

PR #101, opened 2020-10-20
Update peer endpoints

### LINKED ISSUE #97: ​[Clarification] /eth​/v1​/node​/peers
This endpoint has a few interesting fields which would be good to clarify I think. 

I noticed that there wasn't a `count` endpoint. So for apps that just want the count, they have to obtain all the peer data?

For this specific endpoint, there is the `address` field which the examples show a multiaddr. Firstly, what exactly should this address be. Is it simply the multiaddr that can potentially be obtained from the ENR, in which case its likely redundant information, and also an ENR may not contain this information. 
Perhaps this is the "last seen address". I.e the Socket that we observe the connection on?. If so, this potentially could have multiple values for implementations that allow multiple values. Should this then be a comma separated list? 
Also the example shows the TCP port, but most nodes will have UDP also. I guess this is attempting to show just the libp2p listening address, or last seen socket addr on the TCP side?

Another point to clarify is the `state` and `direction` field. In Lighthouse these are grouped into a single enum. If a peer is disconnected, it doesn't have an `ingoing` or `outgoing` state. Can this be nullable in this instance, or do we need to keep track of past connected states for disconnected peers?
In general, if in the future we allow multiple connections per peer, the peer could have both ingoing and outgoing connections. In this case which should be returned?

mpetrunic:
Breaking:
peer.address -> peer.last_seen_p2p_address

Added clarifications and examples from #97 

@AgeManning is this ok?

resolves #97

### DIFF
--- a/apis/node/peer_count.yaml
+++ b/apis/node/peer_count.yaml
@@ -1,0 +1,32 @@
+get:
+  operationId: getPeerCount
+  tags:
+    - Node
+  summary: Get peer count
+  description: Retrieves number of known peers.
+  responses:
+    "200":
+      description: Request successful
+      content:
+        application/json:
+          schema:
+            title: GetPeerCountResponse
+            type: object
+            properties:
+              data:
+                type: object
+                properties: 
+                  disconnected:
+                    type: number
+                    example: 12
+                  connecting:
+                    type: number
+                    example: 34 
+                  connected:
+                    type: number 
+                    example: 56
+                  disconnecting:
+                    type: number
+                    example: 5
+    "500":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
--- a/apis/node/peer_count.yaml
+++ b/apis/node/peer_count.yaml
@@ -17,16 +17,20 @@ get:
                 type: object
                 properties: 
                   disconnected:
-                    type: number
-                    example: 12
+                    allOf:
+                      - $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+                      - example: "12"
                   connecting:
-                    type: number
-                    example: 34 
+                    allOf:
+                      - $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+                      - example: "34"
                   connected:
-                    type: number 
-                    example: 56
+                    allOf:
+                      - $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+                      - example: "56"
                   disconnecting:
-                    type: number
-                    example: 5
+                    allOf:
+                      - $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+                      - example: "5"
     "500":
       $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
--- a/apis/node/peers.yaml
+++ b/apis/node/peers.yaml
@@ -4,6 +4,23 @@ get:
     - Node
   summary: "Get node network peers"
   description: "Retrieves data about the node's network peers"
+  parameters:
+    - name: limit
+      in: query
+      required: false
+      schema:
+        description: "Max amount of items per page"
+        type: number
+        default: 10
+        minimum: 1
+    - name: page
+      in: query
+      required: false
+      schema:
+        description: "Current page"
+        type: number
+        default: 0
+        minimum: 0
   responses:
     "200":
       description: Request successful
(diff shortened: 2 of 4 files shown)


mcdee:
Do we really want pagination in this API?  What is the use case for fetching an arbitrary subset of peers?

mpetrunic:
> Do we really want pagination in this API? What is the use case for fetching an arbitrary subset of peers?

you are suppose to include disconnected peers as well, it could be response with 500-600 peers or even more on mainnet

mcdee:
Surely that is an argument to have filtering, not pagination.

mpetrunic:
> Surely that is an argument to have filtering, not pagination.
 
On what fields would you filter? 

Let's say I want all disconnected outgoing peers, that could still be 300 responses

mcdee:
> Let's say I want all disconnected outgoing peers, that could still be 300 responses

Yes, but if that is the information you want then why receive it piecemeal going through multiple requests rather than all in one response?

From my experience, pagination is useful when results are _ordered_ and _end-user facing_.  For example, if someone wanted to list validators ordered by balance it's possible they are doing so to just see the first _x_ validators, in which pagination makes sense because the user may well stop at page 1 or 2 of several hundred.  However, if the common requirement is to obtain all results, possibly restricted by filter, I don't see the benefit of doing this.

A good example of where this caused problems was in the beacon chain explorers recently.  They use the prysm API to fetch validators, and it's paginated.  The APi is by default limited to 250 responses, so the explorer required around 300 calls to obtain the data.  The problem was that fetching validators for a non-head state required recreating the state from the last checkpoint, so what happened was:

  - the beacon node recreated the state for all 70-odd thousand validators (which was taking up to 1 minute)
  - the beacon node returned 250 of the validators
  - the rest of the results were discarded

This caused the explorers to be unable to provide information.  It _could_ be possible to cache the data, but then you're in to a world of second-guessing which results are going to be requested again, cache management, and all that fun.

All that said, if client teams want to build pagination in to their APIs that's up to them.  But if it's going to happen I'd far rather it was purely optional, and standard across all APIs (perhaps by adding generic `pageSize` and `cursor` options to all queries).  The default should always be to return all requested values.

apis/node/peers.yaml
```diff
@@ -3,7 +3,24 @@ get:
   tags:
     - Node
   summary: "Get node network peers"
-  description: "Retrieves data about the node's network peers"
+  description: "Retrieves data about the node's network peers. Peers should be sorted by their peer_id propery alphabetically."
```
AgeManning, line 6:
```suggestion
  description: "Retrieves data about the node's network peers. Peers should be sorted by their peer_id property alphabetically."
```

AgeManning:
Originally I was concerned about there not being an endpoint for peer count. 

I imagine there are going to be a lot of applications wanting to know how many connected peers a client has. If I've understood the current API's correctly, in order to do that, they have to receive all the data for all peers the client currently knows about and then filter. I would imagine they would poll regularly for this. 

Perhaps we have a peer_count endpoint which can have the queries `connected`, `disconnected` etc, which returns a single uint.

mcdee:
> Perhaps we have a peer_count endpoint which can have the queries `connected`, `disconnected` etc, which returns a single uint.

Or not take a query and return the number in each state, which could be easier on the implementer.

mpetrunic:
As per comments here, removed pagination and peer ordering. I've added peer filtering by connection direction and state.

Since we cannot really have `/node/peers/peer_count` since peer_count would be peer id in that case, I've added `peer_count` (number of connected peers) to `/node/indentity`. I'm not too happy by that solution so if you have better proposal, I can revert last commit and do that instead.

AgeManning:
What about `/node/peer_count` ? 

Or maybe we could add it to `health`? 

I'm not too concerned either way. Just some suggestions. I assume its something that would get polled regularly for misc applications.

mcdee:
A separate `/eth/v1/node/peer_counts` seems to follow the existing API endpoint layout most closely.  A per-state count would be useful, _e.g._:

```
{
 "data": {
    "disconnected": "12",
    "connecting": "34",
    "connected": "56",
    "disconnecting": "78"
  }
}
```

(As an aside, we should specify all possible states somewhere).

apis/node/peer_count.yaml
```diff
@@ -0,0 +1,32 @@
...
+          schema:
+            title: GetPeerCountResponse
+            type: object
+            properties:
+              data:
+                type: object
+                properties: 
+                  disconnected:
+                    type: number
+                    example: 12
```
djrtwo, line 21: Do we use "number" anywhere else? or all just uint64?

apis/node/peers.yaml
```diff
@@ -3,7 +3,24 @@ get:
   tags:
     - Node
   summary: "Get node network peers"
-  description: "Retrieves data about the node's network peers"
+  description: "Retrieves data about the node's network peers. By default it's gonna return all peers. Multiple query params are combined using AND conditions"
```
djrtwo, line 6:
```suggestion
  description: "Retrieves data about the node's network peers. By default this returns all peers. Multiple query params are combined using AND conditions"
```

djrtwo (commented):
looks good, quick question and a minor copy suggestion

apis/node/peer_count.yaml
mpetrunic: I think we should switch to uint64 to be consistent

djrtwo (approved):

protolambda:
@mpetrunic I still see a `meta` field in the `peers` list endpoint with a `count` in it, is that a remnant from the removed peers pagination? Shouldn't it always be equal to the length of the returned json list?

mpetrunic:
> @mpetrunic I still see a `meta` field in the `peers` list endpoint with a `count` in it, is that a remnant from the removed peers pagination? Shouldn't it always be equal to the length of the returned json list?

Seems like I forgot to remove it. Yeah it should always be length of list

PR #102, opened 2020-10-21
Move attestation submission to array.

mcdee:
At current the POST to `/eth/v1/beacon/pool/attestations` takes a single `Attestation`.  This PR proposes that this is changed to an array of `Attestation`s.

Rationale behind this is that beacon nodes often carry out a fair chunk of work to decide if to accept and/or broadcast an attestation.  This can include processing slots to obtain the relevant state, validating attestations against said state, and checking for the existence of subnets to broadcast the attestation.  These operations are commonly repeated for each attestation, which means that if a validator client has multiple attestations to submit for the same slot it can lead to more work, longer delays before broadcasting attestations, and ultimately a less efficient network.

To provide more concrete information related to the above, the following is a histogram of response times for submitting individual attestations against a current beacon node:

screenshot_2020-10-21-170123

The validator client submits approximately 90 attestations per slot.  It can be seen that most of them are processed relatively quickly, however many attestations take over 1/2 a second to process, and there are a number of occasions where the time to process an attestation is over 4 seconds.  Note that although this graph is for a single beacon node, similar results have been seen with other beacon node implementations.

By providing multiple attestations at the same time it is possible for beacon nodes to significantly reduce duplicate work and reduce the duration from start of slot to time broadcast last attestation for the slot.  Note that testing with fewer attestations per slot the entire graph shifts to the left but retains a similar shape, implying that even with smaller numbers of attestations beacon nodes will be more efficient if they receive multiple attestations in a single request.

### DIFF
--- a/apis/beacon/pool/attestations.yaml
+++ b/apis/beacon/pool/attestations.yaml
@@ -44,7 +44,12 @@ get:
 post:
   operationId: submitPoolAttestations
   summary: Submit Attestation objects to node
-  description: Submits Attestation objects to node. If attestations pass all validation constraints, node MUST publish attestations on appropriate subnet.
+  description: |
+    Submits Attestation objects to the node.  Each attestation in the request body is processed individually.
+    
+    If an attestation is validated successfully the node MUST publish that attestation on the appropriate subnet.
+
+    If one or more attestations fail validation the node MUST return a 400 error with details of which attestations have failed, and why.
   tags:
     - Beacon
     - ValidatorRequiredApi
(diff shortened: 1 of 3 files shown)


mpetrunic:
Curious how to implement that. What happens when one of the attestations received in array is invalid?

mcdee:
For a general-purpose "submit multiple things" request I'd suggest we return a structured list of failures if there are any, something like:

```
{
  "code": 400,
  "message": "Some failures",
  "failures": [
    {
      "index": 5,
      "message: "invalid signature"
    },
    {
      "index": 6,
      "message: "invalid attestation"
    }
}
```

We need this anyway for submitting `aggregate_and_proofs` so may as well define something here and port it over to that call as well.

mpetrunic:
Yeah, I guess that's fine and we should port it to aggregate and proofs as well

mcdee:
Cool; I'll update this PR with details of the error response and once this has been merged I'll open another for `aggregate_and_proofs`.

mcdee:
Made the error object a separate type, as we'll be reusing it for `aggregate_and_proofs`.  Still not _au fait_ with swagger, so please take a look and let me know if I mucked anything up.

mpetrunic (approved):
LGTM

djrtwo:
Have any client teams chimed in on this?

paulhauner:
We had previously been using arrays for attestation submission, duties collection, etc. Then, such a format was declared untenable for a REST API so we refactored to single-item endpoints saw significant performance degradation. Now, we're adding them back in again and it's fixing the performance issues. So, we support these changes.

djrtwo (approved):

PR #111, opened 2020-10-29
add state not found possibility to state validators endpoint

rolfyone:
Similar to other endpoints pulling data from the beacon state, it's possible to not be able to retrieve a state when this request is made.

The state_id parameter is valid, but the state was not fetchable, (bytes32 didn't exist is one possibility, or the state is not retrievable for another reason)

This is distinct from a 400 error where it might be invalid input (state_id = "current" would be a 400), and it is useful to determine that the state isn't available vs. the user has requested invalid data.

### DIFF
--- a/apis/beacon/states/validators.yaml
+++ b/apis/beacon/states/validators.yaml
@@ -54,6 +54,16 @@ get:
               - example:
                   code: 400
                   message: "Invalid state ID: current"
+    "404":
+      description: "State not found"
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+              - example:
+                  code: 404
+                  message: "State not found"
     "500":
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

mpetrunic:
gonna take a look at that ci error, it apears to be happening because of latest spectral release. Probably gonna open PR to pin action to specific version

rolfyone:
thanks, I have no idea what to do with the CI error 👍

djrtwo:
@mpetrunic This looks good to me. Plan is the pin spectral release still?

mpetrunic:
> @mpetrunic This looks good to me. Plan is the pin spectral release still?

Yeah, sorry got distracted a bit. I will look into it, but I'm confused since it's working on other PRs that are more recent than this one.

mpetrunic:
I removed state_id ref, it's a bit annoying as it's duplicated param definition but I would go with this. Pinning spectral to some version is gonna take time as we need to upstream it and we are left out of any progress and bugfixes

djrtwo (approved):

Issue #108, opened 2020-10-24
`/eth/v1/beacon/states/{state_id}/committees/{epoch}` unclear


mcdee:
The comment for `epoch` on this endpoint states: "Epoch for which to calculate committees. Defaults to beacon state epoch."  This doesn't fit with our existing usage of endpoints, where none of the other path values are optional.

I can think of a couple of solutions.  We could make two endpoints:

- `/eth/v1/beacon/states/{state_id}/committees` which returns the committees for the current epoch as per the `state_id`
- `/eth/v1/beacon/states/{state_id}/committees/{epoch}` which returns the committees for the specified `epoch` and `state_id`

If it is felt that this stretches the sub-resource definition past breaking point then as an alternative we could go with just the first endpoint and change `epoch` to an optional query parameter.

mpetrunic:
> If it is felt that this stretches the sub-resource definition past breaking point then as an alternative we could go with just the first endpoint and change epoch to an optional query parameter.

I'm in favour of this

rolfyone:
I think that the epoch being an optional query parameter is a good idea, it fits better with the usage

mpetrunic:
@mcdee is this resolved? Can we close?

mcdee:
Yep.

PR #112, opened 2020-10-29
Make epoch parameter query rather than path

mcdee:
As per #108, this moves the optional `epoch` parameter from being a path parameter to a query parameter.  It also expands on the use of other query parameters as filters.


mpetrunic (approved):
lgtm

djrtwo (approved):

PR #114, opened 2020-11-08
Convert chain ID to uint64

mcdee:
Ethereum 1 chain ID is an anomaly in that it is presented as a number rather than a string.  This changes the value to be a uint64, and hence presented as a string, as per the conversation at https://discord.com/channels/595666850260713488/710831784991916072/775098382095482910.


rolfyone (approved):
LGTM

mpetrunic (approved):

djrtwo (approved):

PR #115, opened 2020-11-11
Specify that SSE events must contain valid objects

paulhauner:
I've specified that objects sent on SSE events must be valid. This PR serves to provide my suggestion (see code changes) and gather feedback from others.

### Pros for including invalid objects

- Allows for identifying faulty validators.
- Probably useful for research.

### Cons for including invalid objects

- Introduces the only scenario where we tell API consumers about invalid objects.
    - If we go down this path, I think we must have a `valid` field on those objects.
- Potentially opens our API consumers up to damage from DoS attacks.


mpetrunic (approved):

djrtwo (approved):
very reasonable

side question: shouldn't we have SSEs for receiving *valid* proposer and attester slashings? Yes, some (most?) nodes might horde them, but the gossip channels exist and altruistic actors can broadcast

PR #119, opened 2020-12-08
Add ssz response for debug state

mpetrunic:


mcdee:
Does it make sense to have this as a generic comment about use of the "Accept" header and list the methods that should return SSZ data if requested (as well as specifics in each method)?  I can see this being useful for state, blocks, and attestations at the least.

mpetrunic:
> Does it make sense to have this as a generic comment about use of the "Accept" header and list the methods that should return SSZ data if requested (as well as specifics in each method)? I can see this being useful for state, blocks, and attestations at the least.

Maybe but how would you make it generic?

mcdee:
> Maybe but how would you make it generic?

Have a section up top that says something like:

> All requests by default send and receive JSON, and as such should have either or both of the "Content-Type: application/json" and "Accept: application/json" headers.  In addition, some request can return data in the SSZ format.  To indicate that SSZ data is required in response to a request the header "Accept: application/octet-stream" should be sent.  Note that only a subset of requests can respond with data in SSZ format; these are noted in each individual request.

djrtwo (approved):

PR #120, opened 2021-01-05
Epoch information for ValidatorResponse and ValidatorBalanceResponse

Neurone:
- Added information about **epoch** when responding to Validator and ValidatorBalance query. This allow the client to easily understand what epoch is referred by response data and this is essential when the query URL contains a label ("head", "finalized", etc.) and not a defined state.

### DIFF
--- a/types/api.yaml
+++ b/types/api.yaml
@@ -15,6 +15,10 @@ ValidatorResponse:
       $ref: "#/ValidatorStatus"
     validator:
       $ref: "./validator.yaml#/Validator"
+    epoch:
+      allOf:
+        - $ref: "./primitive.yaml#/Uint64"
+        - description: "Data are referred to this epoch."
 
 ValidatorBalanceResponse:
   type: object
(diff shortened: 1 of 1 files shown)


ajsutton:
I can see that this could be useful, but there are quite a few other APIs which this would also apply to - pretty much everything under `/eth/v1/beacon/states/:state_id/` basically so I'm not sure it makes sense to add this just for validator balances. It would also apply to things like `/eth/v1/beacon/blocks/:block_id/attestations` which don't currently include the block slot the attestations wound up coming from.

I believe currently the work around for not having this data would be to query the current head slot via `/eth/v1/node/syncing` or justified/finalized checkpoint epochs via `/eth/v1/beacon/states/head/finality_checkpoints` then explicitly query for those balances for the required epoch.  You could also follow `head` events if you wanted to track the latest epoch.

It would be useful to get more detail on your particular use case to understand why you're querying the state using a keyword but are particularly interested in which epoch it's from.

mpetrunic:
closing because no activity

PR #125, opened 2021-02-18
Allow the getBlock endpoint to return SSZ

michaelsproul:
It useful to be able to fetch blocks as SSZ bytes, just as it is useful to be able to fetch states.

### DIFF
--- a/apis/beacon/blocks/block.yaml
+++ b/apis/beacon/blocks/block.yaml
@@ -1,7 +1,9 @@
 get:
   operationId: getBlock
   summary: Get block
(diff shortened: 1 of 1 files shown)


mpetrunic (changes requested):
Hey, great idea, tnx for opening PR.

Could you add additional response type like here: https://github.com/ethereum/eth2.0-APIs/blob/master/apis/debug/state.yaml#L30 ?

michaelsproul:
Thanks for the tip, done!

mpetrunic (approved):

PR #129, opened 2021-03-18
proof of concept altair blocks

rolfyone:
demonstrate the concept of a response object containing `version`, and an object that is version dependent.

Block could alternatively just have sync_aggregate as an optional field, but it's likely to be better to be consistent with the versioning approach.

There's a number of issues with this particular change, but I was hoping to get something as a demonstration rather than final version. Happy to update with any changes if it's close, but equally happy to discard if this has been done incorrectly.

### DIFF
--- a/api-v2/beacon/blocks/block.yaml
+++ b/api-v2/beacon/blocks/block.yaml
@@ -10,20 +10,7 @@ get:
       $ref: '../../../beacon-node-oapi.yaml#/components/parameters/BlockId'
   responses:
     "200":
-      description: Success
-      content:
-        application/json:
-          schema:
-            title: GetBlockResponse
-            type: object
-            properties:
-              version:
-                type: string
-                enum: [phase0, altair]
-              data:
-                oneOf:
-                 - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/SignedBeaconBlock"
-                 - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/v2/SignedBeaconBlock"
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/v2/GetBlockResponse'
     "400":
       description: "The block ID supplied could not be parsed"
       content:
--- a/api-v2/beacon/blocks/block.yaml
+++ b/api-v2/beacon/blocks/block.yaml
@@ -1,0 +1,48 @@
+get:
+  operationId: getBlock
+  summary: Get block
+  description: Retrieves block details for given block id.
+  tags:
+    - Beacon
+  parameters:
+    - name: block_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/BlockId'
+  responses:
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            title: GetBlockResponse
+            type: object
+            properties:
+              version:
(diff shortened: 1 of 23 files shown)


api-v2/beacon/blocks/block.yaml
```diff
@@ -0,0 +1,48 @@
...
+      description: Success
+      content:
+        application/json:
+          schema:
+            title: GetBlockResponse
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [phase0, altona]
```
ralexstokes, line 22:
```suggestion
                enum: [phase0, altair]
```

rolfyone: thanks, i had it in my mind it was altona, and changed most but missed this one, pushed up the change.

beacon-node-oapi.yaml
```diff
@@ -199,6 +201,9 @@ components:
       $ref: './types/http.yaml#/ErrorMessage'
     IndexedErrorMessage:
       $ref: './types/http.yaml#/IndexedErrorMessage'
+    v2:
```
mpetrunic, line 204: maybe better to group schemas under altair?

api-v2/beacon/blocks/block.yaml
```diff
@@ -0,0 +1,48 @@
...
+  responses:
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            title: GetBlockResponse
+            type: object
+            properties:
+              version:
```
mpetrunic, line 20: should be defined in schema components and ref-ed here

types/sync_aggregate.yaml
```diff
@@ -0,0 +1,11 @@
+SyncAggregate:
+  type: object
+  description: "The [`SyncAggregate`](https://github.com/ethereum/eth2.0-specs/blob/dev/specs/altair/beacon-chain.md#syncaggregate) object from the Eth2.0 Altair spec."
```
mpetrunic, line 3: It should be linked with spec tag or commit to prevent repo reorg break our links

mpetrunic (commented):
👍🏻 

I left some comments related to structure :)

beacon-node-oapi.yaml
rolfyone: I figured that v2 might be more than just altair, if the version is moving to a lower level... I don't mind moving it, just not sure?

mpetrunic: It makes sense for GetBlockResponse but IMO for SignedBeaconBlock (and other objects changed) it would be better to group by fork.

In case of another fork, we would add to v2.GetBlockResponse new fork name and newFork.SignedBeaconBlock

Does that make sense?

mcdee: Yep this sounds good.  Version the API with simple numbers and version the structures with the fork name (internally we could rename the v1 data structures with 'genesis' to make this clearer, but that wouldn't be part of this PR).

rolfyone: I think so.

So v2 response objects under a v2 group, and 'altair' objects under an `altair` group? I'll push a change...

mpetrunic (approved):

YeeMie16 (approved):

PR #130, opened 2021-04-15
Create v2 of getState, which supports an Altair state object being returned

rolfyone:
In a similar design to GetBlockV2, GetStateV2 can return phase0 or altair states.

Also updated links in GetBlockV2 to refer to the newer spec, after validating there are no new changes.

### DIFF
--- a/apis/debug/state.v2.yaml
+++ b/apis/debug/state.v2.yaml
@@ -9,6 +9,7 @@ get:
   parameters:
     - name: state_id
       in: path
+      required: true
       $ref: '../../beacon-node-oapi.yaml#/components/parameters/StateId'
   responses:
     "200":
--- a/apis/debug/state.v2.yaml
+++ b/apis/debug/state.v2.yaml
@@ -9,13 +9,7 @@ get:
   parameters:
     - name: state_id
       in: path
-      example: "head"
-      required: true
-      schema:
-        type: string
-      description: |
-        State identifier.
-        Can be one of: "head" (canonical head in node's view), "genesis", "finalized", "justified", \<slot\>, \<hex encoded stateRoot with 0x prefix\>.
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
   responses:
     "200":
       description: Success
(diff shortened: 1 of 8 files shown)


apis/debug/state.v2.yaml
```diff
@@ -0,0 +1,61 @@
+get:
+  operationId: "getStateV2"
+  summary: "Get full BeaconState object"
+  description: |
+     Returns full BeaconState object for given stateId.
+     Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ
+  tags:
+    - Debug
+  parameters:
+    - name: state_id
```
mpetrunic, line 10: param can be reused like here: https://github.com/ethereum/eth2.0-APIs/blob/master/apis/beacon/states/committee.yaml#L8

types/primitive.yaml
```diff
@@ -52,3 +52,8 @@ Signature:
   type: string
   pattern: "^0x[a-fA-F0-9]{192}$"
   example: "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"
+
+Uint8:
+  type: string
+  pattern: "^0x[a-fA-F0-9]{2}$"
+  example: "0xff"
```
mpetrunic, line 59: Hex? Aren't we using stringified decimal numbers?

rolfyone: I wasn't sure here, they're flags which would make more sense in hex, I could change back to stringified numbers if thats preferred? not sure if its easy to specify limits that way, but I can try again?

apis/debug/state.v2.yaml
rolfyone: this was literally a copy of state.yaml, should i update that too?

mpetrunic: Those things can sometime slip trough cracks, tnx for fixing 😄

mpetrunic (approved):

PR #134, opened 2021-04-21
add `/eth/v1/validator/duties/sync/{epoch}`

rolfyone:
create a spec for sync committee duties for altair.

I'm open to making this more clearly altair specific, for now I've worked it into the typing information.

Basically I transcribed @mcdee hackmd from https://hackmd.io/@QYHAVYiHRg65pdI_CEm7Eg/syncommitteeapi#POST-ethv1validatordutiessyncepoch

### DIFF
--- a/apis/validator/duties/sync.yaml
+++ b/apis/validator/duties/sync.yaml
@@ -7,7 +7,7 @@ post:
   description: "Requests the beacon node to provide a set of sync committee duties for a particular epoch."
   parameters:
     - name: epoch
-      description: "Should only be allowed 1 epoch ahead"
+      description: "epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD <= current_epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD + 1"
       in: path
       required: true
       schema:
--- a/apis/validator/duties/sync.yaml
+++ b/apis/validator/duties/sync.yaml
@@ -1,0 +1,68 @@
+post:
+  tags:
+    - ValidatorRequiredApi
+    - Validator
+  summary: "Get sync committee duties"
+  operationId: "getSyncCommitteeDuties"
+  description: "Requests the beacon node to provide a set of sync committee duties for a particular epoch.
+
+    Duties should only need to be checked once per epoch,
+    however a chain reorganization (of > MIN_SEED_LOOKAHEAD epochs) could occur,
+    resulting in a change of duties. For full safety, you should monitor head events and confirm the
+    dependent root in this response matches:
+
+      - event.previous_duty_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
+
+      - event.current_duty_dependent_root when `compute_epoch_at_slot(event.slot) + 1 == epoch`
+
+      - event.block otherwise
+
+    The dependent_root value is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
+    or the genesis block root in the case of underflow."
+  parameters:
+    - name: epoch
+      description: "Should only be allowed 1 epoch ahead"
+      in: path
+      required: true
+      schema:
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+  requestBody:
+    description: "An array of the validator indices for which to obtain the duties."
+    required: true
+    content:
+      application/json:
+        schema:
+          title: GetSyncCommitteeDutiesBody
+          type: array
+          items:
+            $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+            minItems: 1
+  responses:
+    "200":
+      description: Success response
+      content:
+        application/json:
+          schema:
+            title: GetSyncCommitteeDutiesResponse
+            type: object
+            properties:
+              dependent_root:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/DependentRoot"
+              data:
+                type: array
+                items:
+                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Altair.SyncDuty'
+    "400":
+      description: "Invalid epoch or index"
+      content:
+        application/json:
+          schema:
+            allOf:
+              - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+              - example:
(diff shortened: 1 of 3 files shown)


apis/validator/duties/sync.yaml
```diff
@@ -0,0 +1,68 @@
...
+    dependent root in this response matches:
+
+      - event.previous_duty_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
+
+      - event.current_duty_dependent_root when `compute_epoch_at_slot(event.slot) + 1 == epoch`
+
+      - event.block otherwise
+
+    The dependent_root value is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
+    or the genesis block root in the case of underflow."
```
ajsutton, line 21: This won't be the right dependent root for sync committees as they span multiple epochs.  I think it winds up being:

```
previous_sync_commitee_period = epoch / EPOCHS_PER_SYNC_COMMITTEE_PERIOD - 1
first_epoch = previous_sync_committee_period * EPOCHS_PER_SYNC_COMMITTEE_PERIOD

get_block_root_at_slot(state, compute_start_slot_at_epoch(first_epoch) - 1)
```
again with genesis block in the case of underflow.

In other words it should be the last block of the epoch prior to the start of the previous sync committee period. That however is a bit of a problem because individual block root history only goes back 256 epochs which is a single sync committee period.  So you can't use `get_block_root_at_slot` to get the dependent root in this case and would actually need a quite old state.

Notably the head event doesn't include that value so it would also have to have a current and previous sync committee dependent root field added for this to be useful.

Given the exceptionally long re-org that would be required to change sync committee assignments I wonder if it's worth including the dependent root at all.

mcdee: > Given the exceptionally long re-org that would be required to change sync committee assignments I wonder if it's worth including the dependent root at all.

I agree.  When I wrote the original API spec the sync committee was based on the current epoch so this would have been required, but if we now have this much lookahead I don't see the dependent root being relevant.

rolfyone: ok I'll remove it from the response.

rolfyone:
@mcdee follow up questions:
 - if i request 10 validator indexes, and only 1 is in a subcommittee, then the output would be only the 1 that is in a subcommittee?
 - if the head epoch is pre-altair, should an error be raised or return an empty committee response? It's looking like we can't look forward from altair-1 to get the altair committee duties, so this will be relatively important to have a consistent approach... Our current theory is an empty committee response, but open to opinions on that...

mpetrunic (approved):
LGTM, would love to get feedback from some implementers from other teams though

apis/validator/duties/sync.yaml
```diff
@@ -0,0 +1,52 @@
+post:
+  tags:
+    - ValidatorRequiredApi
+    - Validator
+  summary: "Get sync committee duties"
+  operationId: "getSyncCommitteeDuties"
+  description: "Requests the beacon node to provide a set of sync committee duties for a particular epoch."
+  parameters:
+    - name: epoch
+      description: "Should only be allowed 1 epoch ahead"
```
michaelsproul, line 10: I guess this is leftover from copying the attester duties API spec.

I think we should definitely specify a bound on the `epoch`, but I think the correct one requires `EPOCHS_PER_SYNC_COMMITTEE_PERIOD` to be expressed properly:

```python
epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD <= current_epoch // EPOCHS_PER_SYNC_COMMITTEE_PERIOD + 1
```

i.e. `epoch` must be within the next sync committee period, or earlier

rolfyone: I had removed the description implementing, but hadn't updated here because I wasn't sure how to express it formally... I can copy/paste this description and go from there...

types/validator.yaml
```diff
@@ -76,6 +76,23 @@ ProposerDuty:
...
+  SyncDuty:
+    type: object
+    properties:
+      pubkey:
+        $ref: './primitive.yaml#/Pubkey'
+      validator_index:
+        allOf:
+          - $ref: "./primitive.yaml#/Uint64"
+          - description: "Index of validator in validator registry."
+      sync_committee_indices:
```
michaelsproul, line 89: Can we consider renaming this field?

If it stays then the name `sync_committee_indices` will be overloaded to mean:

1) The validator indices of a sync committee
2) The positions of a single validator index in a sync committee

Given that these two meanings appear together frequently it's quite confusing. I think a better name for the 2nd definition would be `sync_committee_positions`, although I'm also open to other suggestions. Another option would be `validator_sync_committee_indices`, in keeping with the similarly named `validator_committee_index` for attestation duties.

dapplion:
What's the range of correct `epoch` values if you want to get the nextSyncCommittee duties?

dapplion:
In relation to the comment above I believe it would be simpler if we change the route to
```
/eth/v1/validator/duties/sync/{period}
```

rolfyone:
Query by epoch is fine, the server side can calculate the period trivially. Introducing a new parameter to query by is likely to be more confusing rather than helpful.

The server will need to be able to get an appropriate state to process the request, so if it's too far ahead, or the beacon node is not storing old states, then the window will be limited for what is going to be a serviceable request. Any node that's not syncing should be able to respond with the current sync committee period and next, so that's a lot of epochs that will be valid.

All the client wants is 'give me the sync committee duties at epoch X', and it the beacon node can figure out the details.

In terms of when to re-query, at the simplest clients can re-query each epoch if they desire and it shouldn't be a drain. if they want to save some queries, then the sync committee period is well documented in the spec...

The valid epochs should be from the altair fork epoch to a sync committee period ahead of the current epochs committee period.

dapplion:
Already this PR in Lodestar. I would merge.

> if i request 10 validator indexes, and only 1 is in a subcommittee, then the output would be only the 1 that is in a subcommittee?

Yes

> if the head epoch is pre-altair, should an error be raised or return an empty committee response? It's looking like we can't look forward from altair-1 to get the altair committee duties, so this will be relatively important to have a consistent approach... Our current theory is an empty committee response, but open to opinions on that...

Empty response is good

dapplion (approved):

mpetrunic (approved):

PR #136, opened 2021-04-28
add POST /eth/v1/validator/sync_committee_subscriptions

rolfyone:
https://hackmd.io/@QYHAVYiHRg65pdI_CEm7Eg/syncommitteeapi#POST-ethv1validatorsync_committee_subscriptions

notable differences:
 - `sync_committee_index` became an array: sync_committee_indices, as multiple subnets per validator is possible.

 - `is_aggregator` is not really needed in this instance as far as we can tell, and potentially won't be known at the time of making the subscription.

 - added `until_epoch` so that the BN can unsubscribe based on this request.

### DIFF
--- a/apis/validator/sync_committee_subscriptions.yaml
+++ b/apis/validator/sync_committee_subscriptions.yaml
@@ -1,0 +1,27 @@
+post:
+  operationId: "syncCommitteeSubscriptions"
+  summary: "Subscribe to sync committee subnets"
+  description: |
+    Subscribe to a number of sync committee subnets
+
+    Sync committees are not present in phase0, but are required for Altair networks.
+
+    Subscribing to sync committee subnets is an action performed by VC to enable network participation in Altair networks, and only required if the VC has an active validator in an active sync committee.
+
+  tags:
+    - ValidatorRequiredApi
+    - Validator
+  requestBody:
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+            $ref: '../../beacon-node-oapi.yaml#/components/schemas/Altair.SyncCommitteeSubscription'
+  responses:
+    "200":
+      description: "Successful response"
+    "400":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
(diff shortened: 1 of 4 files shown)


apis/validator/sync_committee_subscriptions.yaml
```diff
@@ -0,0 +1,27 @@
+post:
+  operationId: "syncCommitteeSubscriptions"
```
mpetrunic, line 2: maybe better to be `prepareSyncCommitteeSubnet` matches better beacon committee subsction and generate dclient method would be more verbose  `client.prepareSyncCommitteeSubnet({params})`

rolfyone: Updated but left as plural, given it's an array

mpetrunic (approved):

dapplion:
> is_aggregator is not really needed in this instance as far as we can tell, and potentially won't be known at the time of making the subscription.

@rolfyone Could you explain in more detail why it's not needed?

ajsutton:
> > is_aggregator is not really needed in this instance as far as we can tell, and potentially won't be known at the time of making the subscription.
> 
> @rolfyone Could you explain in more detail why it's not needed?

The beacon node is required to subscribe to the subnet regardless of whether the validator is an aggregator or not, so there's no change in beacon node behaviour.  That provides the stable backbone of peers on the subnet.  Also given sync committee periods are 256 epochs long (might have even changed to 512?) a validator would have to sign 256*32 = 8192 slots to check if it was an aggregator and it almost always will be at some point during the sync committee period.

dapplion:
@rolfyone `until_epoch`  means until the start that epoch or the end of that epoch? Could you commit a clarification as a comment?

rolfyone:
The intent was that you'd unsubscribe once that epoch was past, so if it's 'untilEpoch' 20, you unsubscribe at the first slot of 21.
https://github.com/ethereum/eth2.0-APIs/pull/143
I'm not the biggest fan of that wording, but the PR is up.

PR #140, opened 2021-05-09
add GET `/eth/v2/validator/blocks/{slot}`

rolfyone:
Added a note on the v1 endpoint that it only supports phase0 block production

 Added a note in the description of v2 to indicate the list of supported blocks will change over time as forks progress, hopefully setting the expectation that we're not creating a new api each time a fork happens.


mpetrunic (approved):

mcdee:
lgtm

PR #141, opened 2021-05-09
POST `/eth/v2/beacon/blocks`

rolfyone:
- added a note to the `/eth/v1/beacon/blocks` endpoint to stipulate only supporting phase0 blocks.

### DIFF
--- a/apis/beacon/blocks/blocks.v2.yaml
+++ b/apis/beacon/blocks/blocks.v2.yaml
@@ -3,7 +3,7 @@ post:
     - Beacon
     - ValidatorRequiredApi
   summary: "Publish a signed block."
-  operationId: "publishBlock"
+  operationId: "publishBlockV2"
   description: |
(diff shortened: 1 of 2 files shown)


mpetrunic (approved):

dapplion:
> That said, given we're posting the block as JSON it's possible to extract the slot declared in the block and use that to determine which fork it should be from.

Implementing the V2 endpoint I was just merging the V1 and V2 routes with that approach. I think it makes sense to go for https://github.com/ethereum/eth2.0-APIs/pull/142

rolfyone:
I'm definitely leaning the way of #142. 
Similarly for producing an unsigned block, as block changes are additive, we could just add an extra type of response object being possible, and not create a new v2 for that, but it does mean validator clients still need to implement the new block type or bad things will happen... no different to needing to support altair technically, but it would be admittedly be easier to find if you had to change the endpoint in use.

mpetrunic:
closing in favour of #142

PR #139, opened 2021-05-06
GET `/eth/v1/beacon/states/{state_id}/sync_committees`

rolfyone:
add definition for sync_committees query of a state

basically adding https://hackmd.io/@QYHAVYiHRg65pdI_CEm7Eg/syncommitteeapi#GET-ethv1beaconstatesstate_idsync_committees to the spec

Names were changed slightly to more closely match the state object structure.

PR #146, opened 2021-05-17
Remove extra array wrapping the sync committees response

rolfyone:


rolfyone:
#139 was almost right, but theres an extra array I found while implementing that's not needed.

mpetrunic (approved):

PR #150, opened 2021-06-02
add syncnets to identity metadata for altair

rolfyone:
- Added as an optional field, it's only required once we hit altair and this will maintain compatibility with the current interface.

### DIFF
--- a/types/p2p.yaml
+++ b/types/p2p.yaml
@@ -34,6 +34,12 @@ MetaData:
         - description: "Bitvector representing the node's persistent attestation subnet subscriptions."
         - $ref: "./primitive.yaml#/Hex"
         - example: "0x0000000000000000"
+    syncnets:
+      allOf:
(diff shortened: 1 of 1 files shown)


types/p2p.yaml
```diff
@@ -34,6 +34,12 @@ MetaData:
         - description: "Bitvector representing the node's persistent attestation subnet subscriptions."
         - $ref: "./primitive.yaml#/Hex"
         - example: "0x0000000000000000"
+    syncnets:
+      allOf:
+        - description: "Bitvector representing the node's sync committee subnet subscriptions."
```
mpetrunic, line 39: Should we add notice that it will be present after altair fork is activated?

rolfyone: I wasn't sure of the wording, but did make an update. Happy to change it further if there's a better way of expressing it.

mpetrunic (approved):

PR #151, opened 2021-06-04
Add local keymanager API

dapplion:
**Why?**

The main goal of this effort is to increase client diversity. UIs are very important to some users so this standard should ensure all clients have at least one.

The secondary goal is to offload UI work from core dev teams to other parties that can prioritize this work over Altair, the merge, etc.

> WIP: This standard must be cheap to implement to be successful. Core devs are encouraged to voice concerns around complexity to ensure a timely implementation.

----

**EDIT**: Note - this proposal is initially motivated for a specific type of validator operator: the solo home DIY staker without good technical knowledge. This API is a simplification of the existing Prysm's validator API, which is also designed for that type of staker

-----

Minimal validator API to enable a basic validator manager workflow. **Supports**:

- Import, delete keystores
- Import, export slashing db

Does **not support**:

- Metrics visualization or plotting any type of chain data. Instead, should be done with metrics tooling such as Grafana or block explorers
- VoluntaryExit. Instead, should be done with the deposit cli tooling.
- Backup accounts. When migrating validator clients stakers should re-generate the keystores from the mnemonic or retrieve them from a self-managed backup.
- Displaying logs.

All routes SHOULD be disabled by default and only be enabled with a CLI flag such as `--enable-validator-api`.

All sensitive routes MUST be authenticated. There SHOULD exist a single password per validator client binary such that multiple users are not supported.

The validator API MUST allow to add and delete accounts but SHOULD not allow to retrieve them. Users are encouraged to use alternative locations to backup their keystores and use them to perform validator client migration.

----

See the hackmd document for more context https://hackmd.io/@dapplion/eth2-validator-api


rolfyone:
I'm a little concerned about this. Allowing remote access to this kind of functionality may make it very easy to get slashed...

Having CLI access implies a level of admin access, and I guess with that a level of responsibility, and it'd be potentially less open to error with export/import processes...

If a validator client with a password doesn't stop their validator from where ever its running and imports it to another client, that'll result in a slashing event, so we need to think really carefully whether this actually needs to be accessible from the standard rest api...

If we were to allow it, I think you'd want the key/slashing data in a single payload, so that you can't load a key without having slashing protection data, or do something like without the slashing protection enforce a 2 epoch delay or something... 

We might be able to do something like verify it the specified validator hasn't attested or something currently, i'm not sure what validations could be worked in for altair - potentially there's a way to see the validator isn't active...

We also would need to be very careful about stipulations on how this is secured. I'm not sure we want to stipulate that a single password is used, as it should be up to the person running the client and in line with security requirements that the company has.  If they want to go and do one-time-passwords or 2FA, all power to them IMHO.

We can definitely suggest that it's disabled by default without providing a flag etc, and that they need to be authenticated routes, but ultimately that's really the level we can stipulate to, and maybe that at least basic auth should be supported or something like that.

The response of deleting keystores may need to contain more context about slashing summary of the keys removed... 

If we do go this way, it would probably be pertinent to not require a login/logout structure, but also to be able to just issue an authenticated api call that has no real 'session'. Partially I say this because sessions can be particularly painful to manage in balanced environments.

Create/list accounts is something that needs more discussion. It's a definite attack vector to start creating millions of random accounts, and list accounts is something that non admin users shouldn't really be able to do (get half of the user/password problem, then brute force passwords)

Potentially this should be a separate repo, given its dealing purely with keys etc, and won't be dealing with any of the structures in eth2.0-api? It's a whole bunch of infrastructure to confuse the existing apis if we add it all here, and fairly unrelated in real terms... Just a thought...

mcdee:
I would like to echo the concerns of @rolfyone and add a few of my own.

First, and most importantly, this locks in a specific way of thinking about a validator, specifically that there is an endpoint that is responsible for both validating and signing.  This is becoming more commonly not the case, with multiple remote signer implementations available and being used.  This is not only a conceptual issue but a practical one: which process runs the API?  This repo builds an API that is run on the beacon node, and that's no use for these endpoints in many cases.  Ditto the validator client, as it may not have ownership of the keys (but _does_ need to know the pubkeys for which it is validating).  The signer part is the most likely place to put this, or some of this, but they are generally built to be highly secure and allowing creation and deletion of keys seems to be an unlikely fit wit with their existing security model.

Second, the general idea of accessing slashing protection data from an active endpoint is a Bad Thing.  By definition, fetching this data from a running system means that it is out of date as soon as it has been retrieved and so cannot be trusted to be a complete picture when stored or imported elsewhere.

More of an implementation "detail", but the lack of HTTPS is something that seems to be massively insecure.  The idea that the keystore, plus the passphrase to access its private data, are both sent in the same request and in plaintext is a very concerning.

Regarding the goals of this PR: I think that both are laudable, however I suspect that this would be very low impact.  Client diversity, from what I have seen with those that install the software, is much more a question of ease of installation (and, long-term, maintenance) of the beacon node software.  The operations to import validators, although not standard between implementations, is relatively simple and so would not benefit massively from having an API rather than its existing CLI.  As for UI, the majority of any such project would be around the monitoring of an active beacon chain rather than its setup, in which I include the creation of validators.

dapplion:
You are right, in that this proposal is initially motivated for a specific type of validator operator: the solo home DIY staker without good technical knowledge. This API is a simplification of the existing Prysm's validator API, which is also designed for that type of staker. This proposal is definitely not for professional staking pools or institutional stakers.

@rolfyone 

> Potentially this should be a separate repo, given its dealing purely with keys etc, and won't be dealing with any of the structures in eth2.0-api? It's a whole bunch of infrastructure to confuse the existing apis if we add it all here, and fairly unrelated in real terms... Just a thought...

Moving this to a different repo is great idea since I agree this functionality doesn't have any resemblance with existing routes. The server for the validator manager API should be served by the a different process, the validator client. Happy to move this PR to a new repo once it exists.

> If we were to allow it, I think you'd want the key/slashing data in a single payload, so that you can't load a key without having slashing protection data, or do something like without the slashing protection enforce a 2 epoch delay or something...

I agree, that's a great point. The counter-argument there is to not allow to extract the keys in anyway with this API, to contain the damage if auth fails.

> We also would need to be very careful about stipulations on how this is secured. I'm not sure we want to stipulate that a single password is used, as it should be up to the person running the client and in line with security requirements that the company has. If they want to go and do one-time-passwords or 2FA, all power to them IMHO.

This simple auth scheme is meant to be served and used locally for solo stakers. In a company setting a more complex setup would have to be used.

> The response of deleting keystores may need to contain more context about slashing summary of the keys removed...

That's a great point, I will add it.

@mcdee 

> This is not only a conceptual issue but a practical one: which process runs the API? This repo builds an API that is run on the beacon node, and that's no use for these endpoints in many cases.

This API should **not** be run by the beacon node, it should be run by the validator client unless using remote signing.

> First, and most importantly, this locks in a specific way of thinking about a validator, specifically that there is an endpoint that is responsible for both validating and signing. This is becoming more commonly not the case, with multiple remote signer implementations available and being used.

That's a great point. I noted this API is motivated for the solo home staker. However I will research how new approaches to validating can influence this.

> Second, the general idea of accessing slashing protection data from an active endpoint is a Bad Thing. By definition, fetching this data from a running system means that it is out of date as soon as it has been retrieved and so cannot be trusted to be a complete picture when stored or imported elsewhere.

@rolfyone idea of fetching slashing protection data only after deleting the keystores from the validator is good step in that direction.

> More of an implementation "detail", but the lack of HTTPS is something that seems to be massively insecure. The idea that the keystore, plus the passphrase to access its private data, are both sent in the same request and in plaintext is a very concerning.

If users are able to get SSL certificates they should, or just do self-signed certs for local networks. Should the openapi doc reflect that?

> Client diversity, from what I have seen with those that install the software, is much more a question of ease of installation (and, long-term, maintenance) of the beacon node software

Given that all projects publish to docker registries it's quite easy to build systems around that. The goal of this is to allow staking without **forcing** CLI interaction.

> As for UI, the majority of any such project would be around the monitoring of an active beacon chain rather than its setup, in which I include the creation of validators.

We have the metrics standardization effort for that front.

apis/accounts/auth/signup.yaml
```diff
@@ -0,0 +1,28 @@
+post:
+  tags:
+    - AccountAuth
+  operationId: "Signup"
+  summary: "Signup."
+  description: |
+    Register a new password to be used to encrypt the local wallet. The password MUST:
```
paulhauner, line 7: As I mentioned in the previous comment, I think we should keep the API auth account separate from the validator key encryption.

I would say that even baking in any concept of a "wallet" is best avoided. The Lighthouse validator client doesn't pay any attention to wallets (as defined in [EIP-2386](https://eips.ethereum.org/EIPS/eip-2386)). We simply have a bunch of encrypted keystores; we don't care if they were [derived](https://eips.ethereum.org/EIPS/eip-2334) from a common seed or not.

apis/accounts/index.yaml
```diff
@@ -0,0 +1,148 @@
...
+    - name: page_size
+      in: query
+      required: false
+      description: "The maximum number of accounts to return in the response."
+      schema:
+        type: number
+    - name: page_token
+      in: query
+      required: false
+      description: "A pagination token returned from a previous call to `ListAccounts`. that indicates where this listing should continue from."
```
paulhauner, line 20: I believe we generally steer clear of pagination in this repository. I'd be a fan of that here, I can't imagine any responses being prohibitively large and it saves a lot of complexity.

paulhauner (commented):
Hey, this is just a quick fly-by of a few comments I raised during a discussion out of band.

Here is the link to the endpoints we currently have on the Lighthouse API now: https://lighthouse-book.sigmaprime.io/api-vc-endpoints.html This API needs some work before it can become GUI-ready. Among other things, we need to add some endpoints around importing/exporting slashing protection data.

I'll need to come back for a more detailed review, especially around the authentication method. Presently Lighthouse protects the API by requiring an [authorization token](https://lighthouse-book.sigmaprime.io/api-vc-auth-header.html).

For the record, I generally support the standardization of the VC API.

apis/accounts/auth/login.yaml
```diff
@@ -0,0 +1,55 @@
...
+          properties:
+            password:
+              type: string
+            password_confirmation:
+              type: string
+  responses:
+    "200":
+      description: |
+        Successfully authenticated. The session ID is returned in a cookie named `SESSION`. You need to include this cookie in subsequent requests.
+        - HttpOnly: MUST be included
```
cheatfate, line 44: Is it possible to avoid cookies? Because as soon as you introduce cookies for authentication server become stateful. Also because `Max-Age` is not specified its possible to create cookies which will never be deleted, and so can be used if cookie become stolen. Is it possible to use token authentication insteed?

dapplion: Could you describe in more detail what you refer to with "token authentication"?

Given that each node will likely implement different flavours of authentication, using HttpOnly cookies would simplify the UI implementation. The browser will just re-send whatever that specific node has sent in the Set-Cookie header.

- You can make the cookie stateless by signing it with a known key and adding data to it.
- You can also add a timestamp to the cookie's content and use that to invalidate it in the backend, irrespective of the Max-Age directive

cheatfate: You can check for Json Web Token about token authentication which could be passed via headers or request string. The biggest difference is that tokens will not be managed by browsers.
This feature protects validators from loosing cookies through malicious browser extensions or malware.
As soon as communication finishes, tokens become not valid anymore.

paulhauner: Lighthouse has an auth scheme in its existing VC API:

- A secret is passed by the `Authorization` HTTP header: https://lighthouse-book.sigmaprime.io/api-vc-auth-header.html
- Since the secret is a secp256k1 public key, the server also includes a `Signature` header in all responses (note, this is not safe from replay attacks).

Here's an example of a `curl` request using this scheme:

```bash
curl -v localhost:5062/lighthouse/version -H "Authorization: Basic api-token-0x03eace4c98e8f77477bb99efb74f9af10d800bd3318f92c33b719a4644254d4123"
```

This approach is simpler for the VC, but it doesn't allow for user specified passwords. Our design ethos is to keep the VC HTTP server as simple as possible and to handle all the user-facing authentication in the GUI. Here's a design doc we drew up a while ago: https://hackmd.io/zS3h-Kk_TTmYhvyoHap6DQ

Basically, we expect the VC and BN to be hiding behind a reverse-proxy provided by the GUI web app. This avoids needing to provide HTTPS certificates for the BN, VC *and* web application.

dapplion: @paulhauner This Signature Header is very cool idea! Is this like a roll-your-own TLS strategy to guarantee integrity of the responses? If so, could we take it a step further and make the client encrypt the requests with the pubkey so that the keystores + password are not sent in plain text? https://lighthouse-book.sigmaprime.io/api-vc-sig-header.html

paulhauner: >  If so, could we take it a step further and make the client encrypt the requests with the pubkey so that the keystores + password are not sent in plain text?

As I understand it, encrypting the body of a HTTP request is generally not done. That's because the path, headers, etc still remain plain-text and it does not prevent replay attacks. Usually transport-level security (HTTPS) is used to fully solve all these issues.

Encrypting *all* the request bodies seems overkill to me, it would break the ability to do a simple `curl` request. However, perhaps encrypting passwords with that key is a useful last-line-of-defense. I'm unsure.

I am cautious to provide a false sense of security when using HTTP. I was even unsure about adding the `Signature` header, but I went ahead since it's fully optional on the client side.

apis/accounts/index.yaml
```diff
@@ -0,0 +1,148 @@
...
+delete:
+  tags:
+    - Accounts
+  operationId: "DeleteAccounts"
+  summary: "Delete accounts."
+  description: |
+    Delete keystores for the pubkeys provided in the request. The validator client MUST stop validating
+    with these keystores immediately and MUST return a 200 status only after confirming no more signatures
+    can be created with those keys.
+  security:
```
cheatfate, line 116: This operation is unsafe one. Without proper authentication its very easy for somebody drop all the accounts from validator's client database. The only thing which you should know is validator's public key. I think even cookie (and mostly because of cookie authentication) is not enough here. I think the most appropriate one is to request `private key` or `keystore + password` to confirm deletion. So if you key owner it will be not a problem for you, but if you some stranger you obviously could not recover validator's private key from known public key.

apis/accounts/index.yaml
```diff
@@ -0,0 +1,148 @@
...
+      $ref: "../../validator-oapi.yaml#/components/responses/InternalError"
+
+post:
+  tags:
+    - Accounts
+  operationId: "ImportAccounts"
+  summary: "Import accounts."
+  description: |
+    Import keystores generated by the Eth2.0 deposit CLI tooling. All keystores MUST be encrypted with the same password.
+  security:
```
cheatfate, line 70: Process of import keystores is CPU-heavy operation, so its very easy to put `validator_client` into busy state of procession (decrypting) keystores. So this request can be used to perform DoS attack on `validator_client`.
To resolve this issue we should specify maximum amount of keystores allowed for single call to `ImportAccounts`.

apis/accounts/index.yaml
dapplion: Sending the `private key` or `keystore + password` seems like a much worse trade-off to achieve authentication.

apis/accounts/index.yaml
dapplion: I don't think any client restricts the number of keystores to be imported at once. Picking a number would be very arbitrary and restricting this a the spec level doesn't seem right. If a UI implementation wants to restrict it, it would be fine.

apis/accounts/index.yaml
cheatfate: You already sending this pair while adding key, so this is not a trade-off. If you want to avoid it, you need to create some kind of token when adding keystores, so this tokens will not be known by public.

apis/accounts/slashing_protection.yaml
```diff
@@ -0,0 +1,70 @@
+get:
+  tags:
+    - SlashingProtection
+  operationId: "ExportSlashingProtection"
+  summary: "Export slashing protection data."
+  description: |
+    MUST return a file with the format defined in EIP-3076: Slashing Protection Interchange Format.
+    If pubkeys is not specified MUST return data for all pubkeys in the local slashing protection DB.
```
michaelsproul, line 8: Exporting the slashing protection data is futile if the validator client continues to sign messages with the keys after the export. The exported data will become stale almost immediately and offer no protection.

I think any API that exports slashing protection data should also atomically disable the keys for which the data was exported. We could support this either by keeping the API as-is and specifying that the exported keys will also be disabled, or by creating a new unified `/export` endpoint that:

1. Exports the keystores for the requested validator public keys (optionally all of them), including passwords and
2. Exports the slashing protection data for the requested validators

whilst

3. Disabling all requested validator public keys so that no new messages are signed which aren't included in the slashing data from (2)

We could create nice symmetry by allowing the API for _importing_ keystores to accept optional slashing protection data, so that you can feed the output of the export endpoint into the import endpoint in order to migrate validators from one VC to another (we've been discussing this for Lighthouse here: https://github.com/sigp/lighthouse/issues/2557).

If there are no objections to the above design I could try speccing it as a PR to this PR.

michaelsproul (changes requested):

apis/accounts/auth/login.yaml
stefa2k: The beacon client should not provide extensive user authentication & authorization. A simple API key mechanism is enough for putting a UI-based user authenticatoin & authorization on top. In the case of implementing authentication & authorization on the beacon client, where do you draw the line with features? What about 2FA? Different storage techniques for sensible data? Etc.

Handling a JWT (Json Web Token) only makes sense when complex user management is needed.

When HTTPS is needed because access is provided on an unsecure network, a simple SSL termination service (e. g. nginx, haproxy, ...) would suffice and is more secure than handling SSL on the beacon client itself.

apis/accounts/index.yaml
stefa2k: If pagination is implemented, what about sorting? What happens when I access the pages and in the mean time the list changes because of internal reorg, import, deletion?

apis/accounts/index.yaml
stefa2k: Using the method @paulhauner described at the `operationId: "Login"` using api keys (secrets) would resolve this.

dapplion:
I've updated and simplified the routes to do a simple atomic keymanager flow:
1. POST `/` Import keystores + slashing protection data
2. GET `/` List current known keys
3. DELETE `/` Atomically stop validating and export slashing protection data

I've removed the login, logout routes and added notes about the need to protect this routes. I left the suggestion of using an auth token in all requests that should be distributed in some way.

key-manager-oapi.yaml
```diff
@@ -0,0 +1,199 @@
...
+tags:
+  - name: Accounts
+    description: Endpoints to manage key manager accounts.
+  - name: SlashingProtection
+    description: Endpoints to import and export slashing protection db data.
+  - name: AccountAuth
+    description: Endpoints to authenticate accounts requests.
+
+paths:
+  /eth/v1/accounts:
```
james-prysm, line 42: should this api be called accounts? I was wondering if keys or some other key around keys makes more sense to me... at least for internal discussion accounts gets confused with the webui account/user. What do others think? It was one of the terms we wanted to review on the Prysm side at least.

michaelsproul (commented):
I'm a big fan of the new simplified API @dapplion, thank you for putting it together!

Some observations, which I'm not sure are actionable:

1. We're pushing the responsibility for keystore storage on to the user of this API. There's no way to get a keystore out of the API, which means if one wanted to migrate between VCs one would need an independent copy of the keystore files (and password). This is probably desirable from a security point of view, because it means in the worst case, unauthorized access to this API can only cause a liveness fault, and _not_ a slashing. The current Lighthouse VC API also has this property.
2. Deleting keystores increases the risk of data loss slightly compared to just disabling them. On the other hand, given (1) the user should have copies of their keystores handy anyway, and should _definitely_ have a back up of their mnemonic to restore from. Deleting is also safer for preventing slashing (compared to disabling, where the key itself is still present and we rely on the absence of bugs to prevent it being used). 

Aside from that I've raised two items which might be actionable in the code itself.

key-manager-oapi.yaml
```diff
@@ -0,0 +1,199 @@
...
+                          type: string
+                          description: "The validating public key."
+                        derivation_path:
+                          type: string
+                          description: "The derivation path (if present in the imported keystore)."
+                        slashing_protection_last_entry:
+                          type: string
+                          description: |
+                            Latest slashing protection DB data item associated with this pubkey.
+                            Signal no slashing protection DB data with a value of TBD (0, null, "")
```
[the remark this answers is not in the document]
dapplion: Agree, should remove for simplicity since now you can only get slashing protection data after stopping the validator

key-manager-oapi.yaml
```diff
@@ -0,0 +1,199 @@
...
+        "500":
+          $ref: "../../validator-oapi.yaml#/components/responses/InternalError"
+
+    post:
+      tags:
+        - Accounts
+      operationId: "ImportAccounts"
+      summary: "Import accounts."
+      description: |
+        Import keystores generated by the Eth2.0 deposit CLI tooling. All keystores MUST be encrypted with the same password.
```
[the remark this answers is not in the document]
dapplion: I would import one by one then since this seems a minority use case already supported by the API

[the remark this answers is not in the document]
michaelsproul: Yeah I'm happy with that :+1:

key-manager-oapi.yaml
james-prysm: /eth/v1/keys sounds good to me at least.

key-manager-oapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+            application/json:
+              schema:
+                title: ListKeysResponse
+                type: object
+                properties:
+                  keys:
+                    type: array
+                    items:
+                      type: object
+                      properties:
```
rolfyone, line 55: might be worth having an extra field in here that's not a required field, but basically means the key can't be removed.
In teku it's possible to load keys explicitly by filename, and they would be reloaded each startup in that scenario, so it'd be nice to tell people that specific keys aren't able to be deleted via this api...

I'm not sure of the best term, `locked`, or `externally_managed` or something?
If this is part of the standard it would allow others to know when the key can't be removed via this API, which could be of benefit rather than delete fails and no way of understanding why...

key-manager-oapi.yaml
```diff
@@ -0,0 +1,225 @@
...
+        default: "https://public-mainnet-node.ethereum.org"
+
+tags:
+  - name: Keystores
+    description: Endpoints to manage key manager Keystores.
+  - name: SlashingProtection
+    description: Endpoints to import and export slashing protection db data.
+
+paths:
+  /eth/v1/keystores:
```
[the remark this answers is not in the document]
ajsutton: All those validator endpoints under `/eth/v1` are supplied by the beacon node (all the standard apis so far are).  So `/eth` is effectively beacon node and we need to pick something else for the validator client.  `/eth-validator` is probably the right option though it's longer than I'd like.  Tempting to just use `/eth-vc` or even `/ethvc`

key-manager-oapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+          content:
+            application/json:
+              schema:
+                title: ImportKeystoresResponse
+                type: object
+                properties:
+                  statuses:
+                    type: array
+                    description: Status result of each `request.keystores` with same length and order of `request.keystores`
+                    type: object
```
ajsutton, line 110: There's two `type` definitions at the same level here. I think there's a missing `items:` and indenting.

key-manager-oapi.yaml
```diff
@@ -0,0 +1,220 @@
...
+          content:
+            application/json:
+              schema:
+                title: DeleteKeysResponse
+                type: object
+                properties:
+                  statuses:
+                    type: array
+                    description: List of public keys successfully deleted.
+                    type: object
```
ajsutton, line 172: Two type declarations here too

key-manager-oapi.yaml
dapplion: Good suggestion! :+1:

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+          description: Success response
+          content:
+            application/json:
+              schema:
+                title: DeleteKeysResponse
+                type: object
+                properties:
+                  statuses:
+                    type: array
+                    description: List of public keys successfully deleted.
```
rkapka, line 174: The list contains all keys with their statuses, not only the ones successfully deleted.

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+                          description: error message if status == error
+        "401":
+          $ref: "#/components/responses/Unauthorized"
+        "403":
+          $ref: "#/components/responses/Forbidden"
+        "500":
+          $ref: "#/components/responses/InternalError"
+
+    delete:
+      operationId: DeleteKeys
```
james-prysm, line 136: question on intended behavior, if you send an empty request body is the intention to return the current state of slashing protection data? Also is the slashing protection data only on keys that are inactive or returning the data on the active ones as well 🤔

james-prysm: probably never mind on the second part, it doesn't hurt to have more data right 🤔

michaelsproul: No, it's very risky to return slashing protection data for keys that haven't been deleted, as it represents stale information that could get you slashed.

Slashing protection data should only be returned for keys from the request that were successfully deleted (status == deleted). We should clarify that in the description

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        validating_pubkey:
+                          $ref: "#/components/types/Pubkey"
+                        derivation_path:
+                          type: string
+                          description: The derivation path (if present in the imported keystore).
+                        not_deletable:
```
rolfyone, line 61: maybe `readonly` ?

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+      responses:
+        "200":
+          description: Success response
+          content:
+            application/json:
+              schema:
+                title: ListKeysResponse
+                type: object
+                properties:
+                  keys:
```
rolfyone, line 51: I think this might be normally `data` in all the others?

michaelsproul: I think if we have a more specific term we can use then we should use it, e.g. delete and insert return `statuses` at the top-level. However `keys` doesn't match perfectly in this case, so maybe `data` is preferable? Could also just do `data` for all of them?

rolfyone: on the main beacon-api, post doesn't require the wrapper, only the responses... I think it was in case we need to send extra context outside of the data at some point...

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+              properties:
+                keystores:
+                  type: array
+                  description: JSON-encoded keystore files generated with the Launchpad.
+                  items:
+                    type: object
+                passwords:
+                  description: Passwords to unlock imported keystore files. `request.passwords[i]` must unlock `request.keystores[i]`.
+                  items:
+                    type: string
```
rolfyone, line 96: i do prefer the array of passwords, but it does make the description slightly incorrect, it looks like we no longer require the same password, just an entry in the passwords list corresponding to the keystore, where `keystore[i]` uses `password[i]`

key-manager-oapi.yaml
```diff
@@ -0,0 +1,223 @@
...
+                slashing_protection:
+                  type: object
+                  required: false
+                  description: Slashing protection of the imported keys data in EIP-3076 JSON format.
```
rolfyone, line 100: we're going to have to define this here if we want it listed in the schema...

michaelsproul: I think we could use a [URL reference](https://swagger.io/docs/specification/using-ref/) to https://github.com/eth2-clients/slashing-protection-interchange-tests/blob/master/schema.json

key-manager-oapi.yaml
```diff
@@ -0,0 +1,225 @@
...
+                      properties:
+                        status:
+                          type: string
+                          description: |
+                            - deleted: Successfully deleted from persistent storage and not able to sign any messages
+                            - not-found: Pubkey not known to keymanager
+                            - error: Any other status different to the above: I/O errors, etc.
+                          enum:
+                            - deleted
+                            - not-found
```
rolfyone, line 187: not sure about other languages, but 'not-found' isn't enum friendly at least in java - could we have `not_found` or something?

For the scenario where the key is already removed and we're just giving slashing protection data back, might be worth a value so we can use other than the error conditions

```
deleted: key was active and removed
not_active: slashing protection data returned but key was not active
not_found: key was not found to be removed, and no slashing data can be returned
error: unexpected condition meant the key could not be removed (the key was actually found, but we couldn't stop using it) - this would be a sign that making it active elsewhere would almost certainly cause you headaches / slashing conditions etc.
```

michaelsproul: As discussed on Discord, I'm on-board with the above. Underscore naming is also more consistent with the beacon APIs (e.g. the validator statuses)

Oleksandr-prog (approved):

key-manager-oapi.yaml
```diff
@@ -0,0 +1,227 @@
...
+    Forbidden:
+      $ref: "./types/http.yaml#/Forbidden"
+    InternalError:
+      $ref: "./types/http.yaml#/InternalError"
+
+  securitySchemes:
+    cookieAuth: # arbitrary name for the security scheme; will be used in the "security" key later
+      type: apiKey
+      in: cookie
+      name: SESSION # cookie name
```
michaelsproul, line 220: I think the rough consensus around cookies vs tokens fell in favour of tokens (see https://github.com/ethereum/beacon-APIs/pull/151#discussion_r662211483). Prysm have implemented a JWT, and Lighthouse has a bearer token that's compatible.

We can specify token auth using the template here: https://swagger.io/docs/specification/authentication/bearer-authentication/

james-prysm: yup in favor of a token for now as it keeps it more simple.

dapplion:
This code has been moved to this repo https://github.com/ethereum/keymanager-APIs/blob/master/keymanager-oapi.yaml

Thank you so much everyone for engaging in this initiative :heart: To continue the discussion please open an issue to new repo.

- I can open an issue for token auth and format since that's the most contentious issue right now. https://github.com/ethereum/keymanager-APIs/issues/3

PR #159, opened 2021-07-28
Update description of fork schedule.

mcdee:
This updates the description of the config/fork_schedules endpoint to state that all forks should be returned rather than just the upcoming forks.


rolfyone (approved):
LGTM

mpetrunic (approved):

PR #170, opened 2021-10-01
Add version header for ssz responses

### LINKED ISSUE #167: V2 API calls and SSZ encoded response.
Calls like https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2 and https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2 has discriminator field for `application/json` encoding, but its unclear how response should be encoded for `application/octet-stream` (SSZ) encoding - should the response include `version` field too?
Because without discriminator it could be pretty hard to decoded SSZ encoded `SignedBeaconBlock` or `BeaconState` object.

mpetrunic:
resolves #167

### DIFF
--- a/apis/beacon/blocks/block.v2.yaml
+++ b/apis/beacon/blocks/block.v2.yaml
@@ -13,6 +13,13 @@ get:
   responses:
     "200":
       description: "Successful response"
+      headers:
+        Eth-Consensus-Version:
+          description: Required in ssz responses so client can deserialize returned data.
+          schema:
+            type: string
+            enum: [phase0, altair]
+            example: "phase0"
       content:
         application/json:
           schema:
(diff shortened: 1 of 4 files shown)


apis/beacon/blocks/block.v2.yaml
```diff
@@ -13,6 +13,13 @@ get:
   responses:
     "200":
       description: "Successful response"
+      headers:
+        Eth-Consensus-Version:
+          description: Required in ssz responses so client can deserialize returned data.
```
mcdee, line 18: Do we want to state that this is required in all responses, to allow more efficient decoding of JSON streams?

mpetrunic: Should I just extract the header to "constant" and put it on every endpoint?

mcdee: Sorry when I said "all responses" I meant for both JSON and SSZ responses for the particular endpoints for which it is relevant.

We can always add it to additional endpoints if it is required in future, but not sure we should add it as a blanket header for all endpoints.

PR #177, opened 2021-11-08
Update `publishBlock` to disambiguate meaning of `success` in broadcast

arnetheduck:
The current description says:

> and a successful response (20X) only indicates that the broadcast has
been successful

"Successful" can be interpreted in two ways:

1. the block was sent to neighbouring peers (ie the node could parse the
`json`, more or less).
2. the block passed gossip validation and was then sent to neighbouring
peers.

If a beacon node gossips blocks that fail gossip validation, they stand
to lose peer score and may be disconnected from the network - as such,
publishing an invalid block poses a risk to all validators using the
same node.

This PR resolves the ambiguity by catering to the latter interpretation
and assumes gossip validation is performed on blocks before sending
them.

It also stresses that broadcast may be performed before integration so
as to aid timely delivery.


apis/beacon/blocks/blocks.yaml
```diff
@@ -24,9 +25,9 @@ post:
     "200":
       description: "The block was validated successfully and has been broadcast. It has also been integrated into the beacon node's database."
     "202":
-      description: "The block failed validation, but was successfully broadcast anyway. It was not integrated into the beacon node's database."
+      description: "The block failed full validation, but was successfully broadcast anyway. It was not integrated into the beacon node's database."
```
rolfyone, line 28: Maybe 
`The block could not be integrated into the beacon node's database as it failed validation, but was successfully broadcast.`

rolfyone (dismissed):
Overall LGTM, just may be an option to describe 202 slightly differently if we're changing it anyway...

dapplion (dismissed):

rolfyone (approved):

apis/beacon/blocks/blocks.yaml
```diff
@@ -35,9 +36,9 @@ post:
...
```
arnetheduck, line 39:
```suggestion
      description: "The block could not be integrated into the beacon node's database as it failed validation, but was successfully broadcast."
```

rolfyone (approved):
LGTM

PR #182, opened 2021-12-20
Add not found responses for aggregates and contributions

rolfyone:
There are times when an aggregate or contribution is requested by a VC, and the beacon node simply doesn't have any information to return.

Returning 400 'bad request' is probably excessive in this case, the request may be perfectly valid and it's just that the BN wasn't able to respond with data.

We believe that a 404 in these cases would be warranted and allow us a better idea of actual bad requests vs. no data being found at the BN.

### DIFF
--- a/apis/validator/aggregate_attestation.yaml
+++ b/apis/validator/aggregate_attestation.yaml
@@ -9,7 +9,7 @@ get:
     - name: attestation_data_root
       in: query
       required: true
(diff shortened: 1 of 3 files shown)


rolfyone:
I have no idea how I broke that lint!

mpetrunic (approved):

mpetrunic (approved):

PR #185, opened 2022-01-17
Added optional fee_recipient parameter to /eth/v2/validator/blocks/{s…

pk910:
It would be nice to be able to set the fee recipient on a per-validator basis so that validators sharing a BN don't need to use the same address.
Currently this is not possible as there is no way to tell the BN to use a custom fee-recipient address.

This PR adds an optional fee_recipient parameter to `/eth/v2/validator/blocks/{slot}` which allows VC to specify a address that should be used to receive transaction fees collected in the block.

### DIFF
--- a/apis/validator/block.v2.yaml
+++ b/apis/validator/block.v2.yaml
@@ -28,6 +28,12 @@ get:
       description: "Arbitrary data validator wants to include in block."
       schema:
         $ref: '../../beacon-node-oapi.yaml#/components/schemas/Graffiti'
+    - name: fee_recipient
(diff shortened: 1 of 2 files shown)


mcdee:
You cannot add the fee recipient at this point, as blocks have to be prepared in advance.  #178 provides an endpoint for this functionality.

pk910:
Why do we need to prepare EL blocks before calling this api?

mcdee:
Because they take time to build, and the `blocks` request needs to return pretty much immediately.

PR #189, opened 2022-02-04
Add note about caching

paulhauner:
This one took me a while to figure out. Both Chrome and Firefox wouldn't display changes (even with Ctrl+Shift+r and Shift+F5).


mpetrunic (approved):
Ah yeah, tnx!

PR #191, opened 2022-02-04
Replace deprecated docker actions and use npm packages directly

mpetrunic:
There was an increasing number of flaky ci, this should hopefully fix it as we are running the latest version of npm packages instead of deprecated ones in docker actions.


mpetrunic:
@djrtwo Could you review when you can, this should prevent those flaky CI issues

djrtwo (approved):

Issue #179, opened 2021-11-15
validator produceBlockHeader


djrtwo:
We should add a `produceBlockHeader` endpoint on the `validator` namespace. This is to enable any proposer/build separation (PBS) constructions -- both constructions built by third parties or any PBS constructions that are added to L1 natively in the coming years.

The block header contains the critical anti-slashing data -- `slot` -- and the VC does no other validations on the data (especially into the block body) as far as I know. So it is sufficient to just send the block header to the validator and enables any construction in which the body (or some subset of the body) is no known at the time of signing. By moving to the `header` method, the VC can be agnostic to the PBS games the BN may or may not be playing (which could even change on a per-slot or per-epoch basis).

The main potential downside I see is that a VC might multiplex on multiple BNs and want to publish the signed block to multiple nodes at once. Just getting the header  would not allow for this. Due to this feature that is potentially used today, we *can* keep `produceBlock` in addition to `produceBlockHeader`, at least for now. 

Side note: any Proof-of-Execution constructions would only be on attestations rather than block proposal because the chance of sometihng like PBS making into into the protocol is very high. And because an attster attests every epoch, a validator couldn't really be lazy in this regard.

thegostep:
This change is referred to as "blind transaction signing" in mev-boost milestone 1 consensus client changes: https://github.com/flashbots/mev-boost#implementation-plan

mcdee:
Obtaining only a beacon block header would cause problems for systems like Vouch that obtain beacon blocks from multiple sources and select the one with the best impact on the consensus layer (parent block, most suitable attestations, size of sync committee _etc._).

As such, if a VC is not trusted with details of execution payload then that section should be returned as a root, but the remainder of the block should be fully returned.  This should still work in a PBS world, as all fields returned are public.

paulhauner:
> As such, if a VC is not trusted with details of execution payload then that section should be returned as a root, but the remainder of the block should be fully returned. This should still work in a PBS world, as all fields returned are public.

I'm also fine with a solution where we only stub-out the `transactions` field, rather than the entire `body` field.

djrtwo:
Although by default I wish not to create another data structure reference creating the following seems reasonable to suit this purpose:

*Please provide suggestions for the name! It's not a header really...

```python
class BeaconBlockRedactedExecution(Container):
    slot: Slot
    proposer_index: ValidatorIndex
    parent_root: Root
    state_root: Root
    body: BeaconBlockBodyRedactedExecution

[...]

PR #194, opened 2022-02-25
Bellatrix and blinded blocks

realbigsean:
Resolves https://github.com/ethereum/beacon-APIs/issues/179

I wanted to progress #179 because I think multiple CL teams are now working on MEV-boost compatibility implementations so it would be good to have a clear definition. I noticed this repo doesn't yet define API changes for Bellatrix, so I added those in this PR let me know if we want them in a separate PR though.


types/bellatrix/execution_payload.yaml
```diff
@@ -0,0 +1,49 @@
...
+      fee_recipient:
+        $ref: '../primitive.yaml#/ExecutionAddress'
+      state_root:
+        $ref: '../primitive.yaml#/Root'
+      receipts_root:
+        $ref: '../primitive.yaml#/Root'
+      logs_bloom:
+        $ref: '../primitive.yaml#/LogsBloom'
+      prev_randao:
+        $ref: '../primitive.yaml#/Root'
```
mcdee, line 18: Is this item a root?  Thought it was the mix, although not 100% sure on that.

mcdee (commented):
Thanks for taking this on.  A couple of comments.

types/bellatrix/execution_payload.yaml
realbigsean: Yes this is a mix, not a root. We actually use `Root` for `randao_mixes` in `BeaconState` already which I think is why I used it here. But in [this commit](https://github.com/ethereum/beacon-APIs/pull/194/commits/0e93e46248841fd90b266b59607205ac612ef79a) I updated both places to use a new `Bytes32` type. Could also rename `Bytes32` to `RandaoMix` ?

apis/beacon/blocks/blinded_blocks.yaml
```diff
@@ -0,0 +1,43 @@
+post:
+  tags:
+    - Beacon
+    - ValidatorRequiredApi
+  summary: "Publish a signed block."
+  operationId: "publishBlindedBlock"
+  description: |
+    Instructs the beacon node to use the components of the `SignedBlindedBeaconBlock` to construct and publish a 
+    `SignedBeaconBlock` by swapping out the `transactions_root` for the corresponding full list of `transactions`
```
rkapka, line 9:
```suggestion
    `SignedBeaconBlock` by swapping out the `transactions_root` for the corresponding full list of `transactions`.
```

apis/beacon/blocks/blinded_blocks.yaml
```diff
@@ -0,0 +1,43 @@
...
+    The beacon node should broadcast a newly constructed `SignedBeaconBlock` to the beacon network,
+    to be included in the beacon chain. The beacon node is not required to validate the signed
+    `BeaconBlock`, and a successful response (20X) only indicates that the broadcast has been
+    successful. The beacon node is expected to integrate the new block into its state, and
+    therefore validate the block internally, however blocks which fail the validation are still
+    broadcast but a different status code is returned (202)
```
rkapka, line 15:
```suggestion
    broadcast but a different status code is returned (202).
```

paulhauner:
> i.e. change communications between the VC and BN such that only blinded blocks are passed between them, and as a result make these the official block-related endpoints rather than having the separate blinded_blocks endpoints?

I generally like this idea, but perhaps we'd be prudent to add some other method to allow "revealing" of blinded payloads? Otherwise the BN API would be slightly odd in the way that it can't actually produce complete blocks.

mcdee:
> I generally like this idea, but perhaps we'd be prudent to add some other method to allow "revealing" of blinded payloads?

`GET /eth/v2/beacon/blocks/{block_id}` would still be available to provide full blocks.

(Sorry realized that the original proposal wasn't clear; the first endpoint I was talking about is a `POST` for sending signed blocks from the VC to the BN, not the `GET` which would remain with the current functionality.)

mcdee:
Given that kiln is nearly upon us and there could be implications for deprecating the `blocks` API I suggest that we accept this PR as is and revisit the idea of deprecating the validator-related `blocks` endpoints at a later date.

As such, I'm happy with the PR as it stands.

mpetrunic:
> Given that kiln is nearly upon us and there could be implications for deprecating the `blocks` API I suggest that we accept this PR as is and revisit the idea of deprecating the validator-related `blocks` endpoints at a later date.
> 
> As such, I'm happy with the PR as it stands.

Should we release this under v2.2.0?

rkapka:
I noticed that several references lead to Altair types. Should we move these types to a separate `shared` folder?

mcdee:
Shared structs may still be change over time, though, so we'd need some sort of versioned `shared` and may as well just keep using the names of the forks in which they are introduced/updated.

apis/beacon/blocks/blinded_blocks.yaml
```diff
@@ -0,0 +1,43 @@
+post:
+  tags:
+    - Beacon
+    - ValidatorRequiredApi
+  summary: "Publish a signed block."
+  operationId: "publishBlindedBlock"
+  description: |
+    Instructs the beacon node to use the components of the `SignedBlindedBeaconBlock` to construct and publish a 
+    `SignedBeaconBlock` by swapping out the `transactions_root` for the corresponding full list of `transactions`.
```
rkapka, line 9: I think something is not correct here. Which one do we actually publish, a `SignedBeaconBlock` or a `SignedBlindedBeaconBlock`?

djrtwo:
Concur with deprecating the full block endpoint for blinded-block all around.
And definitely agree in releasing this asap (after minimum correctness review is complete)

apis/beacon/blocks/blinded_blocks.yaml
realbigsean: The beacon node can only publish a `SignedBeaconBlock` via gossip but this endpoint accepts a `SignedBlindedBeaconBlock`. So the beacon node will have to transform the `SignedBlindedBeaconBlock` into a `SignedBeaconBlock` before publishing.

The wording here was meant to cover what using blinded blocks in the normal proposal flow would look like. If flashbots is used instead, this is a little different because `SignedBlindedBeaconBlock`'s are "published" to the flashbots relay, but in reality the flashbots relay will have to make this same `SignedBlindedBeaconBlock` to `SignedBeaconBlock` transformation before *actually* publishing the block via gossip. 

> Also would it make sense to extend https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlock with a new type of block for Bellatrix?

This should be sufficient, right? https://github.com/ethereum/beacon-APIs/blob/67b8f4de30039bf46e92df2e78eebda7461be105/apis/beacon/blocks/blocks.yaml#L23

rkapka: Thanks for explaining! And yes, the linked line is sufficient. I just missed it.

realbigsean:
> Concur with deprecating the full block endpoint for blinded-block all around.
> And definitely agree in releasing this asap (after minimum correctness review is complete)

@djrtwo do you mean deprecating the full block endpoints in this PR or in a future release as suggested by @mcdee  here: https://github.com/ethereum/beacon-APIs/pull/194#issuecomment-1060758090H ?

djrtwo:
Ah, I see @realbigsean. I think given time constraints that deprecating it in a subsequent release is the best path

beacon-node-oapi.yaml
```diff
@@ -129,7 +131,9 @@ paths:
   /eth/v1/validator/blocks/{slot}:
     $ref: "./apis/validator/block.yaml"
   /eth/v2/validator/blocks/{slot}:
-      $ref: "./apis/validator/block.v2.yaml"
+    $ref: "./apis/validator/block.v2.yaml"
+  /eth/v2/validator/blinded_blocks/{slot}:
```
mpetrunic, line 135: Isn't this suppose to be :
```suggestion
  /eth/v1/validator/blinded_blocks/{slot}:
```

realbigsean:
I made an update [here](https://github.com/ethereum/beacon-APIs/pull/194/commits/28fa540a2fd8ca305088a2ae389487b0a2b104e4) to handle pre-Belltrix blocks on blinded endpoints. It makes the merge transition logic easier for the VC because it doesn't have to worry about which endpoints to hit. And if we deprecate full blocks endpoints we will need to serve pre-Bellatrix blocks from blinded endpoints regardless.

mpetrunic:
> Merge branch 'bellatrix-and-blinded-blocks' of

Is there a reason why we need a new blinded blocks endpoint (`POST /eth/v2/validator/blinded_blocks/{slot}`)?


Couldn't we just add another BlindedBlock request type to `POST /eth/v2/validator/blocks/{slot}` as it's not breaking change?

rkapka:
I guess the same could be said about `publishBlindedBlock` (add a new type to `/eth/v1/beacon/blocks`).

The reason why I like this distinction as a client implementer is because currently I can easily discern posted block types by slot. If there are two possible types for the same fork, `Bellatrix.BeaconBlock` and `Bellatrix.BlindedBeaconBlock`, the logic will become more convoluted. I also personally think it's better to have several specialized, well-scoped endpoints, rather than one catch-all endpoint.

mpetrunic (dismissed):

mpetrunic (approved):

PR #196, opened 2022-03-09
Move execution_optimistic into the event data

ajsutton:
The addition of execution_optimistic as an additional field isn't allowed by the event stream spec. It needs to be added into the `data:` content.


ajsutton:
Correction: it's not entirely invalid to add a field like that, but it is specifically ignored which still means it needs to be moved into the actual content so it can be used.

mpetrunic:
👍🏻 totally agree

mpetrunic (approved):

PR #197, opened 2022-03-23
`getDebugChainHeadsV2`: move `execution_optimistic` into metadata

rkapka:
The spec of `getDebugChainHeadsV2` has `execution_optimistic` as part of the `data` return value. It should be outside because it's metadata.

### DIFF
--- a/apis/debug/heads.v2.yaml
+++ b/apis/debug/heads.v2.yaml
@@ -13,6 +13,8 @@ get:
             title: GetDebugChainHeadsResponse
             type: object
             properties:
+              execution_optimistic:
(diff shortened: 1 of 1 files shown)


rkapka:
Actually, why do we even need this endpoint? We should add `execution_optimistic` to metadata of `getDebugChainHeads`, right?

ajsutton:
Each head may be independently optimistic so the data has to go inside data rather than as metadata.  There's a whole argument about whether that field could be added to the existing v1 or whether it requires a v2, but regardless of that outcome it should definitely be in the data response, not metadata.

Issue #198, opened 2022-03-24
Weak Subjectivity API endpoint


kasey:
It would be helpful to query the Beacon Node API to see what it believes the current [Weak Subjectivity Checkpoint](https://notes.ethereum.org/@adiasg/weak-subjectvity-eth2) should be. This can be used by block explorers or client teams who wish to automate the publication of WS Checkpoints to be used for client-side validation. In addition, if the endpoint were to return the `hash_tree_root` of the `BeaconState` with the given `Checkpoint` applied, the state root could be by Checkpoint Sync implementations, to download and/or verify a state retrieved by slot.

kasey:
You can compute it client side based on a head state retrieved from the debug state endpoint of the beacon API: `/eth/v2/debug/beacon/states/head`. 

prysm has an experimental CLI tool that can compute the weak subjectivity checkpoint from a remote beacon node API, check out the relevant source code [1] if you want to implement the same thing in your client.

Note that if you are interacting with a prysm beacon node, special flags need to be specified at start up to enable remote clients to retrieve states[2]. 

[1] https://github.com/prysmaticlabs/prysm/blob/2bc3ef29e02caf11850258cfb2fc8af3d631851a/api/client/beacon/checkpoint.go#L165
[2] `--enable-debug-rpc-endpoints --grpc-max-msg-size=65568081`

PR #199, opened 2022-03-24
New endpoint for weak subjectivity data

kasey:
Resolves https://github.com/ethereum/beacon-APIs/issues/198

This adds an endpoint for retrieving weak subjectivity data. Please read #198 for rationale and uses. Thank you!


mpetrunic:
It seems that there is no consensus with implementation teams to implement this. We can either not include or include it as optional with disclaimer

kasey:
I'm going to withdraw this proposal. I think the inclusion of state_root is pretty specific to the use case of using this endpoint for checkpoint sync, which prysm is no longer doing.

PR #209, opened 2022-05-06
Add `register_validator` endpoint

lightclient:
This is another take on #206.

Rather than extending `prepare_beacon_proposer`, in this PR we create a new, separate endpoint `register_validator`. This addresses two pieces of feedback raised against the extension:

1) `prepare_beacon_proposer` performs a critical operation and probably shouldn't be encumbered by any additional responsibilities
2) `register_validator` will likely be executed on a different rhythm than `prepare_beacon_proposer`, so it feels more appropriate to separate out that logic

--

Now a short recap on why it's important this data is sent to the beacon node in the first place.

After the merge, validators will want to extract MEV from blocks. The mechanism that will facilitate this extraction a network of external builders. These builders will provide `ExecutionPayloadHeader`s to proposers to include in their beacon blocks with the promise that they will be paid a portion of the extracted value. From the perspective of the builder, most values in the header are deterministic depending on what the parent is. However, it is important that a few parameters remain under control of the proposer.

First is obviously the `fee_recipient`. Although it's not critical that the payload's actual `fee_recipient` be set to the proposer's preferred address (a simple transfer at the end of the block would suffice), it is crucial block builders have this information ahead of time so that they can begin constructing blocks immediately after their forkchoice moves forward.

The second parameter that is on the critical path to production is `gas_limit`. Strictly speaking, builders could decide whatever limit they like under the rules of consensus, however this gives a small number of builders a lot of power over this protocol parameter. Validators are more aligned with choosing a healthy gas limit, so it's best that value is left up to them.

In order to provide these two values to builders ahead of time in an form that can be authenticated, we introduce this new endpoint. By signing over these values and a timestamp, builders can determine validator's latest preferences before they are scheduled to propose.

The Builder API and spec is currently being developed in [`ethereum/builder-specs`](https://github.com/ethereum/builder-specs). A new domain type has also [been proposed ](https://github.com/ethereum/consensus-specs/pull/2884)which will be utilized by this endpoint to avoid a conflict with consensus-level messages.

### DIFF
--- a/apis/validator/register_validator.yaml
+++ b/apis/validator/register_validator.yaml
@@ -23,8 +23,7 @@ post:
               $ref: '../../beacon-node-oapi.yaml#/components/schemas/SignedValidatorRegistration'
   responses:
     "200":
-      description: |
-        Preparation information has been received.
+      description: Registration information has been received.
     "400":
       $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
     "500":
--- a/apis/validator/register_validator.yaml
+++ b/apis/validator/register_validator.yaml
@@ -3,12 +3,12 @@ post:
   summary: Provide beacon node with registrations for the given validators to the external builder network.
   description: |
     Prepares the beacon node for engaging with external builders. The
-    information will be sent by the beacon node to each external builder it is
-    connected to. It is expected that the validator client will send this
-    information periodically to ensure the beacon node has correct and timely
-    registration information to provide to builders. The validator client
-    should not sign blinded beacon blocks that do not adhere to their latest
-    fee recipient and gas limit preferences.
+    information will be sent by the beacon node to the builder network. It is
+    expected that the validator client will send this information periodically
+    to ensure the beacon node has correct and timely registration information
+    to provide to builders. The validator client should not sign blinded beacon
+    blocks that do not adhere to their latest fee recipient and gas limit
+    preferences.
 
     Note that requests containing currently inactive or unknown validator
     pubkeys will be accepted, as they may become active at a later epoch.
--- a/apis/validator/register_validator.yaml
+++ b/apis/validator/register_validator.yaml
@@ -1,0 +1,31 @@
+post:
+  operationId: "registerValidator"
+  summary: Provide beacon node with registrations for the given validators to the external builder network.
+  description: |
+    Prepares the beacon node for engaging with external builders. The
+    information will be sent by the beacon node to each external builders it is
+    connected to. It is expected that the validator client will send this
+    information periodically to ensure the beacon node has correct and timely
+    registration information to provide to builders. The validator client
+    should not sign blinded beacon blocks that do not adhere to their latest
+    fee recipient and gas limit preferences.
+
+    Note that requests containing currently inactive or unknown validator
+    indices will be accepted, as they may become active at a later epoch.
+  tags:
+    - Validator
+  requestBody:
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+              $ref: '../../beacon-node-oapi.yaml#/components/schemas/SignedValidatorRegistration'
+  responses:
+    "200":
+      description: |
+        Preparation information has been received.
+    "400":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
+    "500":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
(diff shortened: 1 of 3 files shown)


types/registration.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - $ref: 'primitive.yaml#/Uint64'
+        - description: "Unix timestamp of registration."
+    pubkey:
+      allOf:
+        - $ref: 'primitive.yaml#/Pubkey'
+        - description: "BLS public key of validator."
+
+SignedValidatorRegistration:
+  type: object
+  description: "The [`SignedValidatorRegistration`]() object from the Builder API spec."
```
ralexstokes, line 24: I think we should define these as SSZ in the `consensus-specs` style in https://github.com/ethereum/builder-specs

can just have a simple `spec.md` for now

with stuff like:

```python
class SignedValidatorRegistration(Container):
    ....
```

and then we have concrete links to toss into these empty `()`

ralexstokes (dismissed):
left one comment below about defining the common types in the pyspec style *somewhere* and otherwise this looks great.


I strongly support this solution over #206

beacon-node-oapi.yaml
```diff
@@ -161,6 +161,8 @@ paths:
     $ref: "./apis/validator/sync_committee_contribution_and_proof.yaml"
   /eth/v1/validator/prepare_beacon_proposer:
     $ref: "./apis/validator/prepare_beacon_proposer.yaml"
+  /eth/v1/validator/register_validator:
```
ralexstokes, line 164: this may be splitting hairs but this *feels* like it belongs under the `builder` namespace

even if the beacon node is the one implementing it.

I don't feel too strongly about this but I think its worth considering

beacon-node-oapi.yaml
```diff
@@ -161,6 +161,8 @@ paths:
...
+    $ref: "./apis/validator/register_validator.yaml"
```
ralexstokes, line 165:
```suggestion
  /eth/v1/validator/register_validators:
    $ref: "./apis/validator/register_validators.yaml"
```

this does take an array so I think we should pluralize

apis/validator/register_validator.yaml
```diff
@@ -0,0 +1,31 @@
+post:
+  operationId: "registerValidator"
+  summary: Provide beacon node with registrations for the given validators to the external builder network.
+  description: |
+    Prepares the beacon node for engaging with external builders. The
+    information will be sent by the beacon node to each external builders it is
```
ralexstokes, line 6:
```suggestion
    information will be sent by the beacon node to each external builder it is
```

apis/validator/register_validator.yaml
```diff
@@ -0,0 +1,31 @@
...
+    connected to. It is expected that the validator client will send this
+    information periodically to ensure the beacon node has correct and timely
+    registration information to provide to builders. The validator client
+    should not sign blinded beacon blocks that do not adhere to their latest
+    fee recipient and gas limit preferences.
+
+    Note that requests containing currently inactive or unknown validator
+    indices will be accepted, as they may become active at a later epoch.
```
ralexstokes, line 14: indices or pubkeys?

types/registration.yaml
ralexstokes: this is also where we can specify the signing schema https://github.com/ethereum/consensus-specs/pull/2884 et al.

beacon-node-oapi.yaml
lightclient: Okay to resolve then to maintain consistency?

apis/validator/register_validator.yaml
lightclient:
```suggestion
    pubkeys will be accepted, as they may become active at a later epoch.
```

beacon-node-oapi.yaml
lightclient: I don't have a preference here. Will keep it as-is until we hear more feedback.

types/registration.yaml
lightclient: added in `ethereum/builder-specs`

beacon-node-oapi.yaml
ralexstokes: yeah, we can ignore this

beacon-node-oapi.yaml
djrtwo: I wouldn't change the namespace. These are for the validator client (the thing that has the keys) to call. 

A validator has many things it's trying to accomplish -- sign blocks, sign attestations, register duties, send signals to builders

djrtwo (commented):
Looks good! minor comment on the language.
plus I prefer keeping it in the validator namespace as this is something the validator uses in it's BN communications

lightclient:
Hi @mpetrunic, would you mind taking a look at this? Thanks!

mpetrunic (approved):
LGTM

PR #216, opened 2022-05-18
Don't produce messages for optimistic blocks

paulhauner:
This PR ensures that the BN does not produce attestations, aggregate attestations or sync committee contributions that refer to optimistic blocks.

Attesting to an optimistic block is *very bad* and could result in finalizing an invalid execution payload. The changes in this PR mean that the VC doesn't need to check the optimistic status for each object. This saves calls to the API and also protects against faults in VC implementations.

Although sync committee contributions to optimistic blocks couldn't result in an invalid finalized block, it could cause havoc for light clients.

### DIFF
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -4,7 +4,12 @@ get:
     - Validator
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
-  description: "Requests that the beacon node produce an AttestationData."
+  description: |
+    Requests that the beacon node produce an AttestationData.
+
+    A 503 error must be returned if the block identified by the response
+    `beacon_block_root` is optimistic (i.e. the attestation attests to a block
+    that has not been fully verified by an execution engine).
   parameters:
     - name: slot
       in: query
--- a/apis/validator/sync_committee_contribution.yaml
+++ b/apis/validator/sync_committee_contribution.yaml
@@ -4,7 +4,12 @@ get:
     - Validator
   operationId: "produceSyncCommitteeContribution"
   summary: "Produce a sync committee contribution"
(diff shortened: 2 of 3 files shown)


apis/validator/attestation_data.yaml
```diff
@@ -4,7 +4,12 @@ get:
     - Validator
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
-  description: "Requests that the beacon node produce an AttestationData."
+  description: |
+    Requests that the beacon node produce an AttestationData.
+
+    A 503 error must be returned if the block identified by the response
```
mpetrunic, line 10: emm but user doesn't send `beacon_block_root ` in the request? 

Seems like we should just write here that the beacon node should never produce attestation for optimistic block root.

apis/validator/sync_committee_contribution.yaml
```diff
@@ -4,7 +4,12 @@ get:
     - Validator
   operationId: "produceSyncCommitteeContribution"
   summary: "Produce a sync committee contribution"
-  description: "Requests that the beacon node produce a sync committee contribution."
+  description: |
+    Requests that the beacon node produce a sync committee contribution.
+
+    A 503 error must be returned if the block identified by the response
```
mpetrunic, line 10:
```suggestion
    A 503 error must be returned if the block identified by the request field
```
Also isn't it better to return "400 invalid request"?

paulhauner: > Also isn't it better to return "400 invalid request"?

The 503 error is used for "syncing". I figured that having an optimistic block is an artifact of an unsynced node (i.e., the EL is still syncing). I don't feel strongly either way.

djrtwo:
Looks reasonable to me. Will approve when @mpetrunic gets a chance to rview your response above

PR #217, opened 2022-05-22
Add 415 response as a reason to fail POSTs.

rolfyone:
Now that block posts are able to accept SSZ input, it is useful to be able to fail them on the basis that the server doesn't support the nominated type.

A caveat is probably that JSON should be supported by all servers, but its possible that some servers don't yet support SSZ type.

The client is then able to determine that the server they're connected to doesn't support SSZ (for example) and stop attempting requests that can't succeed.


rolfyone:
some context on http status 415
https://httpwg.org/specs/rfc9110.html#status.415

mcdee:
Given that we are looking to add SSZ support across the board in some future PR, and that there will be a transition phase as clients adopt it, having this in sooner rather than later (so that clients can support this) would be a significant benefit to consumers of the API to help understand what the servers can support.

michaelsproul (approved):
LGTM! Happy to have this in the spec so we can improve SSZ support

PR #218, opened 2022-05-22
Cleanup old endpoints that have been deprecated

rolfyone:
- v1/debug/states has been deprecated for a long time now, as has the v1/blocks endpoint.

 - v1/debug/heads seems to be a new deprecation so leaving that in the spec. I'm not sure why v2 exists, it looks like the execution_optimistic should have just been optional so that it was non breaking, but that horse hs bolted.


mpetrunic (approved):

beacon-node-oapi.yaml
```diff
@@ -78,12 +78,8 @@ paths:
     $ref: "./apis/beacon/blocks/headers.yaml"
   /eth/v1/beacon/headers/{block_id}:
     $ref: "./apis/beacon/blocks/header.yaml"
-  /eth/v1/beacon/blocks:
-    $ref: "./apis/beacon/blocks/blocks.yaml"
```
michaelsproul, line 82: I don't think this should have been deleted. This API endpoint isn't deprecated and is the only way for validator clients to publish their non-blinded blocks to the beacon node

rolfyone: hrm i may have deleted the wrong thing? will check now...

rolfyone: you're correct, raising a pr to put back

michaelsproul: thanks!

PR #220, opened 2022-05-27
remove the deprecated `v1/validator/blocks/{slot}` endpoint

rolfyone:
This endpoint is not able to serve altair blocks, so it is no longer of use.


rkapka:
I still see https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Debug/getDebugChainHeads in the list of endpoints. You missed it in https://github.com/ethereum/beacon-APIs/pull/218, can we remove it now?

rolfyone:
I wasn't sure abut that one, it's a lot more recent than the ones I've raised, so it may be still used?

mpetrunic (approved):

PR #222, opened 2022-07-04
Optional randao verification when producing blocks

michaelsproul:
Recently I've been speculatively creating blocks using the `/eth/v2/validator/blocks/{slot}` endpoint, polling the endpoint live at every slot.

Some beacon node implementations verify the `randao_reveal` signature when this endpoint is used, which causes speculative block proposal to fail (because the private key for the true `proposer_index` is not known).

A few months ago I added an optional `verify_randao` parameter to disable the verification in Lighthouse: https://github.com/sigp/lighthouse/pull/3116. Recently I've discovered that Nimbus have started to verify the `randao_reveal` by default too, so I'd like to implore them to adopt the same parameter.

As far as I know the other clients (Teku/Prysm/Lodestar) don't verify the `randao_reveal` currently, so I think it would be acceptable for them to delay implementing the `verify_randao` parameter until they decide to. Alternatively they could maintain the same semantics and accept `verify_randao=false/null`, and error if `verify_randao=true`. I would like to impose the minimum burden on implementers, given that this is a relatively niche use case.

More background:

- Blockdreamer: https://github.com/michaelsproul/blockdreamer
- My use for blockdreamer: https://twitter.com/sproulM_/status/1543422741218439170

### DIFF
--- a/apis/validator/block.v2.yaml
+++ b/apis/validator/block.v2.yaml
@@ -28,6 +28,15 @@ get:
       description: "Arbitrary data validator wants to include in block."
       schema:
         $ref: '../../beacon-node-oapi.yaml#/components/schemas/Graffiti'
+    - name: skip_randao_verification
+      in: query
+      required: false
+      description: |
+        Skip verification of the `randao_reveal` value. If this flag is set then the
+        `randao_reveal` must be set to the point at infinity (`0xc0..00`). This query parameter
+        is a flag and does not take a value.
+      schema: {}
+      allowEmptyValue: true
   responses:
     "200":
       description: Success response


michaelsproul:
CI seems to be failing on an unrelated error in the blinded_blocks API, weirdly it doesn't occur locally for me with spectral 6.3.0

apis/validator/block.v2.yaml
```diff
@@ -18,8 +18,8 @@ get:
         $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
     - name: randao_reveal
       in: query
-      required: true
-      description: "The validator's randao reveal value."
+      required: false
```
djrtwo, line 21: If this is empty, what is BN supposed to return? 0x00? random?

would it be simpler to just keep this as *required* but that the caller can/should fill in the junk if `verify_randao=false`

djrtwo (commented):
looks reasonable but would prefer more client teams to take a look

ajsutton:
Since Teku isn't validating the randao reveal when creating blocks now so it would be nice if we could just ignore this param and not need to make any changes. As such I'd have a mild preference for `rando_reveal` remaining required and callers can just fill it with junk.

It would also mean that `verify_randao` can only be used to disable validation but setting `verify_randao=true` still wouldn't perform verification in Teku. I'm not sure I see a real need to force the beacon node to verify randao when there are already APIs to get the validator public key and do the verification yourself.

michaelsproul:
Thanks for the feedback @ajsutton, @djrtwo :pray:

How does this sound moving forward?

- `randao_reveal` is required but when `verify_randao=false` must be set to the point at infinity (`0xc0..00`). This was @arnetheduck's idea here: https://github.com/status-im/nimbus-eth2/pull/3837#discussion_r914454321.
- Implementations which do not verify the randao reveal are free to ignore the `verify_randao` flag (e.g. Prysm, Teku, Lodestar)

ajsutton:
Yep that works for me.

michaelsproul:
Alternatively we could use a value-less `?no_verify_randao` flag instead, to make the `verify_randao=true` case unrepresentable. I think I prefer this actually.

djrtwo:
Can you rebase with master? `redocli` is merged

michaelsproul:
Done!

djrtwo (approved):
approving. will merge in 2 days unless any additional comments

michaelsproul:
looks like we're good to merge @djrtwo! thanks!

arnetheduck (approved):

PR #230, opened 2022-08-19
Registrations for exited or unknown validators will not be sent to the builder network

StefanBratanov:
Clarify in the `registerValidator` description that registrations for `exited` or unknown validators will not be sent to the builder network.

Related to https://github.com/flashbots/mev-boost/issues/245

### DIFF
--- a/apis/validator/register_validator.yaml
+++ b/apis/validator/register_validator.yaml
@@ -10,8 +10,8 @@ post:
     blocks that do not adhere to their latest fee recipient and gas limit
     preferences.
 
-    Note that requests containing currently inactive or unknown validator
-    pubkeys will be accepted, as they may become active at a later epoch.
+    Note that requests must not contain exited validator pubkeys. However, requests may contain inactive or unknown validator
+    pubkeys, as they may become active at a later epoch.
   tags:
     - Validator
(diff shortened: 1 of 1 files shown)


ralexstokes:
noting that the builder specs only accept `active` or `pending` validators: https://github.com/ethereum/builder-specs/blob/main/apis/builder/validators.yaml#L7

it would make sense to ensure alignment on this so the beacon node/validator client don't end up doing extra work they don't need to do

StefanBratanov:
> noting that the builder specs only accept `active` or `pending` validators: https://github.com/ethereum/builder-specs/blob/main/apis/builder/validators.yaml#L7
> 
> it would make sense to ensure alignment on this so the beacon node/validator client don't end up doing extra work they don't need to do

Thank you. Makes sense. Changed the description.

michaelsproul (dismissed):

rolfyone:
one minor comment from @mcdee re. wording, which we can fix easily before merging:
```
I'm not a massive fan of the use of "will" in the text.  If this is a specification they should be "must" (if I'm reading the spec correctly).
```

rolfyone (approved):
LGTM

PR #234, opened 2022-08-30
Remove format: hex

dapplion:
Not specified as valid format in openapi https://swagger.io/docs/specification/data-models/data-types/

Makes [AJV](https://ajv.js.org) throw error on schema validation "format unknown"

### DIFF
--- a/types/primitive.yaml
+++ b/types/primitive.yaml
@@ -55,7 +55,6 @@ Bytes32:
 
 Graffiti:
   type: string
-  format: hex
   example: "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
   pattern: "^0x[a-fA-F0-9]{64}$"
 


mpetrunic (commented):
Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format

dapplion:
> Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format

What's the point of listing format: hex here? It's not done for any other hex value and it already has a regex.

mpetrunic:
> > Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format
> 
> What's the point of listing format: hex here? It's not done for any other hex value and it already has a regex.

could help tooling to figure out code generation for parsing and type conversion

dapplion:
> > > Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format
> > 
> > 
> > What's the point of listing format: hex here? It's not done for any other hex value and it already has a regex.
> 
> could help tooling to figure out code generation for parsing and type conversion

Then why is it not applied to all hex strings in the codebase?

rolfyone:
While most of the formats are not strictly valid, I think if we don't use something, we end up with a whole bunch of 'string' and no context. Unless there's it's really breaking something, it's the only context of a lot of the data formats that we have...

It would be my preference to keep formats...

mpetrunic:
> > > > Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format
> > > 
> > > 
> > > What's the point of listing format: hex here? It's not done for any other hex value and it already has a regex.
> > 
> > 
> > could help tooling to figure out code generation for parsing and type conversion
> 
> Then why is it not applied to all hex strings in the codebase?

Would you be willing to open PR to add it to all hex strings in the codebase? 😉

dapplion:
> > > > > Hexy is not a type here, it's format. According to spec format allows any value: https://swagger.io/docs/specification/data-models/data-types/#format
> > > > 
> > > > 
> > > > What's the point of listing format: hex here? It's not done for any other hex value and it already has a regex.
> > > 
> > > 
> > > could help tooling to figure out code generation for parsing and type conversion
> > 
> > 
> > Then why is it not applied to all hex strings in the codebase?
> 
> Would you be willing to open PR to add it to all hex strings in the codebase? wink

Yeah sure

dapplion:
Closing for https://github.com/ethereum/beacon-APIs/pull/238

PR #235, opened 2022-08-30
Remove unnecessary allOf from error reponse schemas

dapplion:
The build step done on build resolves all $ref references.

https://github.com/ethereum/beacon-APIs/blob/9cab46ad3c94a4a2779b42fa21f6bb1955b60b56/.github/workflows/main.yml#L18

If a property is specified both in the $ref and the existing schema the latter overrides $ref without needing allOf syntax. So
- If you need to override the description property or add a new example, no need for allOf
- If you need to merge object "properties" you need allOf to perform deep merging at runtime

### DIFF
--- a/apis/validator/prepare_beacon_proposer.yaml
+++ b/apis/validator/prepare_beacon_proposer.yaml
@@ -37,6 +37,6 @@ post:
       description: |
         Preparation information has been received.
     "400":
-      $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
+      $ref: "../../beacon-node-oapi.yaml#/components/responses/InvalidRequest"
     "500":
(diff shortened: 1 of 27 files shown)


apis/validator/prepare_beacon_proposer.yaml
```diff
@@ -37,6 +37,6 @@ post:
       description: |
         Preparation information has been received.
     "400":
-      $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
+      $ref: "../../beacon-node-oapi.yaml#/components/responses/InvalidRequest"
```
rolfyone, line 40: i didn't think double quotes should matter? single quotes should be fine...

rolfyone (approved):
LGTM

mpetrunic:
Surprisingly, even though it's against spec, it actually works in swagger UI

PR #236, opened 2022-08-30
Remove allOf usage where property merging is not necessary

dapplion:
The build step done on build resolves all $ref references.

https://github.com/ethereum/beacon-APIs/blob/9cab46ad3c94a4a2779b42fa21f6bb1955b60b56/.github/workflows/main.yml#L18

If a property is specified both in the $ref and the existing schema the latter overrides $ref without needing allOf syntax. So
- If you need to override the description property or add a new example, no need for allOf
- If you need to merge object "properties" you need allOf to perform deep merging at runtime

Related to https://github.com/ethereum/beacon-APIs/pull/235
Closes https://github.com/ethereum/beacon-APIs/issues/237

### DIFF
--- a/apis/beacon/blocks/root.yaml
+++ b/apis/beacon/blocks/root.yaml
@@ -30,9 +30,8 @@ get:
                 type: object
                 properties:
                   root:
-                    allOf:
-                      - description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
-                      - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
+                    $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
+                    description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
     "400":
       description: "The block ID supplied could not be parsed"
       content:
--- a/apis/beacon/states/committee.yaml
+++ b/apis/beacon/states/committee.yaml
@@ -7,7 +7,7 @@ get:
   parameters:
     - name: state_id
       in: path
-      $ref: "../../../beacon-node-oapi.yaml#/components/parameters/StateId"
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
     - name: epoch
       description: "Fetch committees for the given epoch.  If not present then the committees for the epoch of the state will be obtained."
       in: query
@@ -41,7 +41,7 @@ get:
               data:
                 type: array
                 items:
-                  $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Committee"
+                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Committee'
     "400":
       description: "Invalid state ID, index, epoch, slot, or combination thereof"
       content:
--- a/apis/beacon/states/committee.yaml
+++ b/apis/beacon/states/committee.yaml
@@ -7,32 +7,26 @@ get:
   parameters:
     - name: state_id
       in: path
-      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+      $ref: "../../../beacon-node-oapi.yaml#/components/parameters/StateId"
     - name: epoch
       description: "Fetch committees for the given epoch.  If not present then the committees for the epoch of the state will be obtained."
       in: query
       required: false
       allowEmptyValue: false
       schema:
-        allOf:
-          - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
-          - example: ""
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
     - name: index
       description: "Restrict returned values to those matching the supplied committee index."
       in: query
       required: false
       schema:
-        allOf:
-          - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
-          - example: ""
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
     - name: slot
       description: "Restrict returned values to those matching the supplied slot."
       in: query
       required: false
       schema:
-        allOf:
-          - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
-          - example: ""
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
   responses:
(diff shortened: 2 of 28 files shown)


apis/beacon/blocks/root.yaml
```diff
@@ -30,9 +30,8 @@ get:
                 type: object
                 properties:
                   root:
-                    allOf:
-                      - description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
-                      - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
+                    $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
+                    description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
```
mpetrunic, line 34: This is not working in swagger

mpetrunic (commented):
This doesn't appear to be working in swagger. 

If we cannot make ajv and swagger happy, in my opinion, it makes more sense to do some preprocessing of beacon API schema to make ajv happy 🤷🏻‍♂️

apis/beacon/blocks/root.yaml
dapplion: How are you testing swagger exactly? What does it mean "not working"? Please share reproduction steps and I can look for alternative solutions

dapplion: Are you running bundle before serving the file?

```
swagger-cli bundle ./beacon-node-oapi.yaml -r -t yaml -o ./deploy/beacon-node-oapi.yaml
```

 I get different results than you

apis/beacon/states/committee.yaml
```diff
@@ -7,32 +7,26 @@ get:
   parameters:
     - name: state_id
       in: path
-      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+      $ref: "../../../beacon-node-oapi.yaml#/components/parameters/StateId"
```
rolfyone, line 10: makes it harder to see changes if we're changing quotes when not needed..

apis/beacon/blocks/root.yaml
mpetrunic: @dapplion You probably didn't switch schema to "dev" default it latest version on github without your changes

dapplion: Sorry I still follow how are you getting a `$ref` property rendered in swagger. The pre-processing `swagger-cli bundle` resolves all $ref such that the resulting yaml has cleanly merge properties. So to clarify:

(a)
- Serve root `./beacon-node-oapi.yaml` and render that in swagger

(b)
- You run `swagger-cli bundle ./beacon-node-oapi.yaml -r -t yaml -o ./deploy/beacon-node-oapi.yaml`
- Serve `./deploy/beacon-node-oapi.yaml` and render that in swagger

apis/beacon/states/committee.yaml
dapplion: Right, auto-formater messed the commit, let me revert

dapplion: Reverted all formatting changes :+1:

JurbinM (approved):

apis/beacon/blocks/root.yaml
rolfyone, line 34: here's a specific example of where the description is not rendered...
Screen Shot 2022-09-01 at 8 42 09 am

basically i downloaded your PR code, started `simplehttpserver` like instructions say and changed to `dev` in "Select a definition" at the top of page, 
 - then opened `/eth/v1/beacon/blocks/{block_id}/root`, 
 - then went to 200, changed to schema tab

Expected to see the above description, it wasn't there.

This is basically exactly the problem we used to have before we used allOf, so I'm glad I could find a concrete example. I haven't gone through every single change here - there's a lot and i'm on leave :)

Hopefully that gives context about the kinds of things that can break...

rolfyone (changes requested):
this is definitely not in a state to merge at the moment, adding 'request changes' so that the approval doesn't mean it's merged...

apis/beacon/blocks/root.yaml
dapplion: Did you bundle the oapi file before hand?

- run `swagger-cli bundle ./beacon-node-oapi.yaml -r -t yaml -o ./deploy/beacon-node-oapi.yaml`
- Serve ./deploy/beacon-node-oapi.yaml and render that in swagger

at least in my environment, `apis/beacon/blocks/root.yaml:24` prebundle is

```yaml
            type: object
            title: GetBlockRootResponse
            properties:
              execution_optimistic:
                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ExecutionOptimistic"
              data:
                type: object
                properties:
                  root:
                    $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
                    description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
```

after bundle 

```yaml
                type: object
                title: GetBlockRootResponse
                properties:
                  execution_optimistic:
                    type: boolean
                    example: false
                    description: 'True if the response references an unverified execution payload. Optimistic information may be invalidated at a later time. If the field is not present, assume the False value.'
                  data:
                    type: object
                    properties:
                      root:
                        description: HashTreeRoot of BeaconBlock/BeaconBlockHeader object
                        type: string
                        example: '0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2'
                        pattern: '^0x[a-fA-F0-9]{64}$'
```

which renders the description fine

Screenshot from 2022-09-01 13-44-25

rolfyone: I wasn't aware that an extra command was needed, I've been running this the way the readme says, so if that works, I'll have another look but also we should be updating the run instructions...

rolfyone: That's a different screenie (is for GetStateRootResponse), I don't see the description of the root in `GetBlockHeaderResponse`...
To get mine up:
 - `npm install -g swagger-cli`
 - did your bundle command
 - ran `./simplehttpserver` and switched to `dev` version

I'm not sure if this is consistent with what you asked me to do, so wrote it all out in case I've run it incorrectly...

dapplion:
Closing as following OpenAPI spec is a higher priority than simpler synax

PR #241, opened 2022-09-13
Endpoint for fetching blinded blocks

michaelsproul:
This PR adds a new endpoint analogous to `/eth/v2/beacon/blocks/{block_id}` that serves _blinded_ blocks rather than full blocks. The new endpoint is `/eth/v1/beacon/blinded_blocks/{block_id}`.

We already have other endpoints for creating new blinded blocks and POSTing signed blinded blocks back to the BN, so this PR completes the trilogy.

There are several reasons why such an endpoint is useful and desirable:

- Most consensus clients do not store full blocks with execution payloads in their databases. Fetching a finalized blinded block can be done more quickly and efficiently than serving a full block which has had its payload pruned and would require a call to the EL to reconstruct.
- Blinded blocks contain a lot of useful information that isn't in block headers, including the execution block hash, execution block number, attestations, exits, etc. For many consensus-oriented tools blinded blocks are sufficient to conduct their analysis (this is true of my block fingerprinting tools, and @jclapis's Rocket Pool tools). Blinded blocks can also be useful for builders and execution-oriented tools which might want to know e.g. the `block_number` of the head block, without fetching all transactions.

### DIFF
--- a/apis/beacon/blocks/blinded_block.yaml
+++ b/apis/beacon/blocks/blinded_block.yaml
@@ -1,0 +1,61 @@
+get:
+  operationId: getBlindedBlock
+  summary: Get blinded block
+  description: |
+    Retrieves blinded block for given block ID.
+    Depending on `Accept` header it can be returned either as JSON or as bytes serialized by SSZ
+  tags:
+    - Beacon
+  parameters:
+    - name: block_id
+      in: path
+      required: true
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/BlockId'
+  responses:
+    "200":
+      description: "Successful response"
+      headers:
+        Eth-Consensus-Version:
+          $ref: '../../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+      content:
+        application/json:
+          schema:
+            title: GetBlindedBlockResponse
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [phase0, altair, bellatrix]
+                example: "phase0"
+              execution_optimistic:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ExecutionOptimistic"
+              data:
+                oneOf:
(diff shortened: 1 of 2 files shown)


rolfyone (approved):
LGTM. should be easy to implement in teku, I know you were seeking opinions on that too...

djrtwo (approved):
looks reasonable

apis/beacon/blocks/blinded_block.yaml
```diff
@@ -0,0 +1,61 @@
...
+          $ref: '../../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+      content:
+        application/json:
+          schema:
+            title: GetBlindedBlockResponse
+            type: object
+            properties:
+              version:
+                type: string
+                enum: [phase0, altair, bellatrix]
```
mpetrunic, line 28:
```suggestion
                enum: [bellatrix]
```

apis/beacon/blocks/blinded_block.yaml
```diff
@@ -0,0 +1,61 @@
...
+                example: "phase0"
```
mpetrunic, line 29:
```suggestion
                example: "bellatrix"
```

mpetrunic (commented):
Small comment, otherwise looks good!

apis/beacon/blocks/blinded_block.yaml
michaelsproul: I think it makes sense to return full phase0 and altair blocks on this endpoint as if they are blinded. It enables tools to just use a single endpoint without switching URLs at the Bellatrix fork.

There's a precedent for doing this already with the [blinded block production endpoint](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Validator/produceBlindedBlock), so I imagine all clients can implement it easily:

> Pre-Bellatrix, this endpoint will return a BeaconBlock.

mpetrunic: Sounds good!

PR #242, opened 2022-09-13
Endpoint for state RANDAO mix

michaelsproul:
Add an endpoint to fetch the RANDAO mix from a given state at some epoch.

The endpoint path is `/eth/v1/beacon/states/{state_id}/randao`.

With the merge the RANDAO mix has taken on increased significance, as it is embedded in execution payloads and available in the EVM.

Currently for tools that want to access the `prev_randao` from outside the beacon node for the purposes of payload construction or validation, they must load an entire beacon state using the debug endpoint. This is prohibitively slow for something that _should_ be fast with the right API.

With this API, they can immediately get access to the correct `prev_randao` value using `/eth/v1/beacon/states/head/randao` prior to block construction (or can use the `state_id` corresponding to the parent block).

### DIFF
--- a/apis/beacon/states/randao.yaml
+++ b/apis/beacon/states/randao.yaml
@@ -1,0 +1,63 @@
+get:
+  operationId: "getStateRandao"
+  summary: "Get the RANDAO mix for some epoch."
+  description: |
+    Fetch the RANDAO mix for a requested epoch from the state identified by `state_id`.
+
+    If an epoch is not specified then the RANDAO mix for the state's current epoch will be returned.
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+    - name: epoch
+      description: "Epoch to fetch the RANDAO mix for. Default: state's current epoch."
+      in: query
+      required: false
+      allowEmptyValue: false
+      schema:
+        allOf:
+          - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+          - example: ""
+  responses:
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            title: GetStateRandaoResponse
+            type: object
+            properties:
+              execution_optimistic:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ExecutionOptimistic"
+              data:
+                type: object
+                properties:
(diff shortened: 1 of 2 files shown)


apis/beacon/states/randao.yaml
```diff
@@ -0,0 +1,63 @@
+get:
+  operationId: "getStateRandao"
+  summary: "Get the RANDAO mix for some epoch."
```
ralexstokes, line 3: don't we want per-slot granularity here?

right now you will call this endpoint with a given `epoch` and its value will change throughout the epoch

if we specify per-slot then infra can cache the value and consumers can be more precise with exactly what value they want

realbigsean: You already get that granularity with the `StateId` path param, right? 

So maybe the description here should be more like "Get the RANDAO mix for some epoch in a specified state."

ralexstokes: right, yeah that makes sense

djrtwo: agreed!

djrtwo (commented):
should there be a bound on state accessor or is it fine/easy to do for any depth?

apis/beacon/states/randao.yaml
michaelsproul: done!

michaelsproul:
> should there be a bound on state accessor or is it fine/easy to do for any depth?

Like the other state APIs it depends on the node's configuration & pruning settings. In Lighthouse's default config these queries will be fast close to the head, and unavailable for most historic states unless the user has opted into reconstructing and storing them (in which case they could be slow or fast).

michaelsproul:
This is ready for review again

riposteX:
Any chance you could also return the beacon node's current slot in the response?

Otherwise you may not be receiving the randao you're expecting.

dapplion (dismissed):
LGTM!

michaelsproul:
@riposteX For consistency with the other state APIs I'd prefer to leave the `slot` out. If you want to be sure of the state from which you're drawing the RANDAO mix you can use the `state_id` to specify one. A good method for reliably getting the head's RANDAO mix is to listen for the `head` event on the SSE endpoint (https://ethereum.github.io/beacon-APIs/#/Events/eventstream), and then query for the RANDAO mix using the `state` root you get from there.

rolfyone (approved):
LGTM

PR #245, opened 2022-09-20
Endpoint for EIP-4881 Deposit Contract Snapshot

ethDreamer:
Adds the endpoint for downloading the [`EIP-4881`](https://eips.ethereum.org/EIPS/eip-4881#specification) Deposit Tree Snapshot during weak subjectivity sync.

### DIFF
--- a/apis/beacon/deposit_snapshot.yaml
+++ b/apis/beacon/deposit_snapshot.yaml
@@ -1,0 +1,34 @@
+  get:
+    operationId: getDepositSnapshot
+    summary: "Get Deposit Tree Snapshot"
+    description: |
+      Retrieve [EIP-4881](https://eips.ethereum.org/EIPS/eip-4881) Deposit Tree Snapshot.
+      Depending on `Accept` header it can be returned either as json or as bytes serialzed by SSZ
+    tags:
+      - Beacon
+    responses:
+      "200":
+        description: Success
+        content:
+          application/json:
+            schema:
+              type: object
+              title: GetDepositSnapshotResponse
+              properties:
+                data:
+                  $ref: '../../beacon-node-oapi.yaml#/components/schemas/DepositSnapshotResponse'
+          application/octet-stream:
+            schema:
+              description: "SSZ serialized block bytes. Use Accept header to choose this response type"
+      "404":
+        description: "No Finalized Snapshot Available"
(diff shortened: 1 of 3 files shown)


michaelsproul (dismissed):
LGTM once merge conflicts are addressed

apis/beacon/deposit_snapshot.yaml
```diff
@@ -0,0 +1,34 @@
...
+              description: "SSZ serialized block bytes. Use Accept header to choose this response type"
+      "404":
+        description: "No Finalized Snapshot Available"
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: "../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+                - example:
+                    code: 404
```
arnetheduck, line 31: we use 404 for when the "thing" is known not to exist (for example an empty slot) - a different code would probably be useful here for ephemeral error conditions (503/syncing for example)

apis/beacon/deposit_snapshot.yaml
```diff
@@ -0,0 +1,34 @@
+  get:
```
arnetheduck, line 1: should there be support for downloading historical snapshots? ie when checkpoint-syncing, one will want to have a matching snapshot to the state that's being synced, which may or may not (due to race conditions or checkpoint age choice) be the same as the snapshot being served by "default"

rolfyone: I think the main idea is just to get a starting point rather than download the whole deposit log, so I think there's probably not a need currently to get historical snapshots...

rolfyone:
resolved conflicts. Let me know if you're interested in continuing discussion or happy to merge @arnetheduck

apis/beacon/deposit_snapshot.yaml
rolfyone: I do tend to agree 404 isn't ideal (a server with a route that is missing is also 404) - even if it was a 400, or potentially 503 might work... I'll see if i can get some more opinions on this one.

rolfyone:
There's been no further feedback in more than 10 days. resolved conflicts again and will merge.

rolfyone (approved):
LGTM

PR #248, opened 2022-09-30
Add capella types and endpoints.

mcdee:
Update existing types to capella spec, use them where appropriate (blocks, execution payloads).

Add `BLSToExecutionChange` type and related types and endpoints.

Spec references point to master at current, will change once a tagged capella release is made.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -26,6 +26,7 @@ get:
             - block
             - attestation
             - voluntary_exit
+            - bls_to_execution_change
             - finalized_checkpoint
             - chain_reorg
             - contribution_and_proof
@@ -58,6 +59,11 @@ get:
               value: |
                 event: voluntary_exit
                 data: {"message":{"epoch":"1", "validator_index":"1"}, "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}
+            bls_to_execution_change:
+              description: The node has received a BLS to execution change (from P2P or API)
+              value: |
+                event: bls_to_execution_change
+                data: {"message":{"validator_index":"1", "from_bls_pubkey":"0x933ad9491b62059dd065b560d256d8957a8c402cc6e8d8ee7290ae11e8f7329267a8811c397529dac52ae1342ba58c95", "to_execution_address":"0x9be8d619c56699667c1fedcd15f6b14d8B067f72"}, "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}
             finalized_checkpoint:
               description: Finalized checkpoint has been updated
               value: |
--- a/types/bls_to_execution_change.yaml
+++ b/types/bls_to_execution_change.yaml
@@ -1,0 +1,23 @@
+BLSToExecutionChange:
+  type: object
+  description: "The [`BLSToExecutionChange`](https://github.com/ethereum/consensus-specs/blob/master/specs/capella/beacon-chain.md#blstoexecutionchange) object from the CL spec."
+  properties:
+    validator_index:
+      $ref: "./primitive.yaml#/Uint64"
+      description: "Index of the validator for which credentials will be changed."
+    from_bls_pubkey:
+      $ref: "./primitive.yaml#/Pubkey"
+      description: "Public key of existing credentials."
+    to_execution_address:
+      $ref: "./primitive.yaml#/ExecutionAddress"
+      description: "Execution address to which the credentials will be changed."
+
+SignedBLSToExecutionChange:
+  type: object
+  description: "The [`SignedBLSToExecutionChange`](https://github.com/ethereum/consensus-specs/blob/master/specs/capella/beacon-chain.md#signedblstoexecutionchange) object from the CL spec."
+  properties:
+    message:
+      $ref: "#/BLSToExecutionChange"
+    signature:
+      $ref: './primitive.yaml#/Signature'
+
--- a/types/capella/block.yaml
+++ b/types/capella/block.yaml
@@ -42,6 +42,10 @@ Capella:
         properties:
           execution_payload:
             $ref: './execution_payload.yaml#/Capella/ExecutionPayload'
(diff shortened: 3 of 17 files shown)


dapplion (dismissed):

types/capella/withdrawals.yaml
```diff
@@ -0,0 +1,7 @@
+Capella:
+  Withdrawals:
+      type: array
+      items:
+        allOf:
+          - $ref: '../withdrawal.yaml#/Withdrawal'
```
rolfyone, line 6: should be able to be just a ref? not sure we need `allOf` if we're not overriding anything...

types/capella/state.yaml
```diff
@@ -0,0 +1,109 @@
...
+        items:
+          allOf:
+            - $ref: '../validator.yaml#/Validator'
+      balances:
+        type: array
+        description: "Validator balances in gwei"
+        maxItems: 1099511627776
+        items:
+          allOf:
+            - $ref: '../primitive.yaml#/Uint64'
```
rolfyone, line 58: found a concrete example of not using allOf:
```
      items:
        $ref: '../primitive.yaml#/Uint64'
```

rolfyone: I haven't put comments everywhere, but at least this is the idea...

types/capella/withdrawals.yaml
mcdee: Removed where applicable.

types/capella/state.yaml
mcdee: Thanks, updated,

apis/eventstream/index.yaml
```diff
@@ -60,6 +61,11 @@ get:
               value: |
                 event: voluntary_exit
                 data: {"message":{"epoch":"1", "validator_index":"1"}, "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}
+            bls_to_execution_change:
+              description: The node has received a BLS to execution change (from P2P or API)
+              value: |
+                event: bls_to_execution_change
+                data: {"message":{"validator_index":"1", "from_bls_pubkey":"0x933ad9491b62059dd065b560d256d8957a8c402cc6e8d8ee7290ae11e8f7329267a8811c397529dac52ae1342ba58c95", "to_execution_address":"0x9be8d619c56699667c1fedcd15f6b14d8B067f72"}, "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}
```
rolfyone, line 68: not sure we would actually need these in event stream?

types/bls_to_execution_change.yaml
```diff
@@ -0,0 +1,23 @@
...
+    from_bls_pubkey:
+      $ref: "./primitive.yaml#/Pubkey"
+      description: "Public key of existing credentials."
+    to_execution_address:
+      $ref: "./primitive.yaml#/ExecutionAddress"
+      description: "Execution address to which the credentials will be changed."
+
+SignedBLSToExecutionChange:
+  type: object
+  description: "The [`SignedBLSToExecutionChange`](https://github.com/ethereum/consensus-specs/blob/master/specs/capella/beacon-chain.md#signedblstoexecutionchange) object from the CL spec."
```
rolfyone, line 17: might be worth mentioning this is signed with the Withdrawal key, rather than the validator key?

types/capella/block.yaml
```diff
@@ -0,0 +1,89 @@
...
+        type: array
+        items:
+          $ref: '../voluntary_exit.yaml#/SignedVoluntaryExit'
+      sync_aggregate:
+        $ref: '../altair/sync_aggregate.yaml#/Altair/SyncAggregate'
+
+  BeaconBlockBody:
+    allOf:
+      - $ref: '#/Capella/BeaconBlockBodyCommon'
+      - type: object
```
rolfyone, line 41: block body would get a `bls_to_execution_changes` added...

apis/eventstream/index.yaml
mcdee: The work here follows that of voluntary exits, which are in the event stream.  Given their low volume it's probably better to have them available than not.

types/bls_to_execution_change.yaml
mcdee: Shouldn't that be in the spec rather than in the API documentation?  We're just following the spec, after all.

types/withdrawal.yaml
```diff
@@ -0,0 +1,12 @@
+Withdrawal:
+  type: object
+  properties:
+    index:
```
rolfyone, line 4: it looks like the current spec also has `validator_index` here 
https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#withdrawal

types/bls_to_execution_change.yaml
rolfyone: sure, just wondering if its useful context to someone trying to consume the api doc

types/withdrawal.yaml
mcdee: Added; thanks.

apis/eventstream/index.yaml
rolfyone: fair enough..

types/capella/block.yaml
rolfyone: sorry should have added reference https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#beaconblockbody

mcdee: Added; thanks.

types/bls_to_execution_change.yaml
mcdee: Unsure, to be honest.  We don't specify this type of information anywhere else, but then this is somewhat an exception when compared to the other validator signatures.

I'm coming down on the side of having it just in the spec, because as a rule I don't like to have things defined in multiple places.  But it's not a strong preference.

rolfyone (approved):
LGTM

types/bls_to_execution_change.yaml
rolfyone: yep that's fair, if we get repeated issues we can clarify...

apis/eventstream/index.yaml
potuz: I know this is merged already, but I don't think having this in the event stream is a good idea, we will see a flood of these right at the fork

rolfyone: I'm not sure who's likely to look at them to be honest, but it's consistent...

I'm not against removing them...

PR #249, opened 2022-10-05
Add `skip_randao_verification` for blinded blocks

michaelsproul:
Follow-up to https://github.com/ethereum/beacon-APIs/pull/222.

The endpoint for producing blinded blocks needs the `skip_randao_verification` parameter for the same reasons as the regular endpoint (see #222). It was an oversight on my part to not include it from the beginning.

The parameter is already implemented in Lighthouse (because code is shared with the non-blinded backend), and it seems that in Nimbus the blinded endpoint is not yet supported. Other clients don't verify RANDAO reveals during block production (yet) and do not need to implement the flag or can implement it as a no-op.


rolfyone (approved):
LGTM

PR #253, opened 2022-10-27
Give liveness epoch same "shape" as attester duties

arnetheduck:
https://github.com/ethereum/beacon-APIs/pull/131 introduced a new liveness endpoint - the request is similar to attester duties in that we query a list of validators for data pertaining to a particular epoch - as such, it seems reasonable to keep the two requests similar in terms of their URL/postdata structure.

### DIFF
--- a/apis/validator/liveness.yaml
+++ b/apis/validator/liveness.yaml
@@ -18,7 +18,7 @@ post:
       in: path
       required: true
       schema:
-        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
   requestBody:
     description: "An array of the validator indices for which to detect liveness."
     required: true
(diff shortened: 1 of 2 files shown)


arnetheduck:
cc @paulhauner

rolfyone:
my understanding was that it's not really something we can query for anything but current or previous epoch... it seems that the epoch field is just going to result in false unless it's very carefully used...

arnetheduck:
> it's not really something we can query for anything but current or previous epoch

it's allowed, just not required - the data _could_ be synthesized from a state for example, or if a client wants to implement "deep" doppelganger protection (ie more than 2 epochs of silence)

rolfyone:
to me this makes sense overall, just would like to hear from @paulhauner and potentially @mehdi-aouadi to get opinions of people implementing...

dapplion:
Change looks good to me :+1:

paulhauner (approved):
LGTM!

rolfyone (approved):
LGTM

PR #258, opened 2022-11-02
Remove maxItems from lists in debug/states

### LINKED ISSUE #257: `https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2` freezes page
`https://ethereum.github.io/beacon-APIs/#/Debug/getStateV2` appears to be broken. The endpoint's details do not load.

rolfyone:
When rendering example data, huge lists will cause the page to stop responding.

This is particularly evident in BeaconState, where our limits are huge, and we only specify a maximum.

The limits are functionally huge so I removed them, as setting minItems didn't seem as effective in keeping the swagger page alive...

fixes #257

### DIFF
--- a/types/altair/state.yaml
+++ b/types/altair/state.yaml
@@ -32,7 +32,6 @@ Altair:
         items:
           allOf:
             - $ref: '../primitive.yaml#/Root'
-        maxItems: 16777216
       eth1_data:
         $ref: "../eth1.yaml#/Eth1Data"
       eth1_data_votes:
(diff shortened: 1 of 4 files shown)


rolfyone:
adding back in the randao_mixes appears to have tipped us over the edge, as the page now has issues rendering again. may need to think more about how we do these examples.

mpetrunic (approved):
Makes sense, don't have better idea eh

PR #260, opened 2022-11-07
WIP: Add endpoints for ```per validator rewards```

kevinbogner:
# WIP

Issue: https://github.com/sigp/lighthouse/issues/3661

Currently, there is no way to get block rewards paid on the consensus side. 

```/eth/v1/beacon/blocks/{slot}/rewards```

```json
{
  "proposer_index": "123",
  "total": "5000",
  "attestations": "3000",
  "sync_aggregate": "2000",
  "proposer_slashings": "0",
  "attester_slashings": "0",
}
```
---

```/eth/v1/beacon/attestations/{epoch}/rewards```

```json
{
  "0": {
     "head": "0",
     "target": "-2000",
     "source": "2000",
  },
  "1": {
    "head": "2000",
    "target": "4000",
    "source": "4000",
  },
 ..
}
```

### DIFF
--- a/apis/beacon/blocks/rewards.yaml
+++ b/apis/beacon/blocks/rewards.yaml
@@ -1,0 +1,7 @@
+get:
+  operationId: WIP
+  summary: WIP
+  description: WIP
+  tags:
+    - Beacon
+    
--- a/apis/beacon/blocks/attestations_rewards.yaml
+++ b/apis/beacon/blocks/attestations_rewards.yaml
@@ -1,0 +1,7 @@
+get:
+  operationId: WIP
+  summary: WIP
+  description: WIP
+  tags:
+    - Beacon
+    
--- a/beacon-node-oapi.yaml
+++ b/beacon-node-oapi.yaml
@@ -90,8 +90,12 @@ paths:
     $ref: "./apis/beacon/blocks/root.yaml"
   /eth/v1/beacon/blocks/{block_id}/attestations:
     $ref: "./apis/beacon/blocks/attestations.yaml"
+  /eth/v1/beacon/attestations/{epoch}/rewards:
+    $ref: "./apis/beacon/blocks/attestations_rewards.yaml"
   /eth/v1/beacon/blinded_blocks/{block_id}:
     $ref: "./apis/beacon/blocks/blinded_block.yaml"
+  /eth/v1/beacon/blocks/{slot}/rewards:
+    $ref: "./apis/beacon/blocks/rewards.yaml"
   /eth/v1/beacon/light_client/bootstrap/{block_root}:
     $ref: "./apis/beacon/light_client/bootstrap.yaml"
   /eth/v1/beacon/light_client/updates:


apis/beacon/attestations/rewards.yaml
```diff
@@ -0,0 +1,37 @@
...
+  description: Retrieves rewards info for attestations 
+  tags:
+    - Beacon
+  parameters:
+    - name: epoch
+      in: path
+      required: true
+      description: "The epoch to get rewards info from"
+      schema:
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
```
michaelsproul, line 13: It would also be handy to have a query parameter like `?id=0,4,5,6` which can specify a subset of validators for which rewards should be computed. This will save processing time and bandwidth. You can copy the `id` parameter from one of the other endpoints like `getStateValidators`.

michaelsproul (changes requested):
This is awesome! Thanks @naviechan & @kevinbogner!

In addition to the minor edits I've proposed the main thing missing is an API for the sync committee rewards. I didn't describe one in my original messages, but it should be similar to the others, maybe something like:

```
/eth/v1/beacon/blocks/{block_id}/sync_committee_rewards?id=...
```

```json
{
    "data": [
        {
            "validator_index": "0",
            "reward": "-1000",
        },
        {
            "validator_index": "1",
            "reward": "1000",
        },
    ]
}
```

The function that defines these rewards is here: https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/beacon-chain.md#sync-aggregate-processing

I'd be open to speccing that endpoint in a separate PR if you like! Maybe one of you could take that on while the other updates the existing endpoints

apis/beacon/attestations/rewards.yaml
```diff
@@ -0,0 +1,49 @@
...
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            InvalidRequest:
+              value:
+                code: 400
+                message: "Invalid request to retrieve attestations rewards info"
+    "404":
+      description: "Invalid epoch"
```
michaelsproul, line 38: I think an actually invalid epoch like something non-numeric should be a 400. The 404 should be just for unavailable data.

```suggestion
      description: "Epoch not known or required data not available"
```

apis/beacon/blocks/rewards.yaml
```diff
@@ -0,0 +1,49 @@
...
+  tags:
+    - Beacon
+    - Rewards
+  parameters:
+    - name: block_id
+      in: path
+      required: true
+      description: "The id of the block to get rewards info from"
+      schema:
+        $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Uint64"
```
michaelsproul, line 14:
```suggestion
        $ref: "../../../beacon-node-oapi.yaml#/components/parameters/BlockId"
```

types/rewards.yaml
```diff
@@ -0,0 +1,70 @@
...
+    source:
+      allOf:
+        - $ref: "./primitive.yaml#/Int64"
+        - example: 4000
+        - description: "attester's reward for source vote in gwei"
+    inclusion_delay:
+      allOf:
+        - $ref: "./primitive.yaml#/Uint64"
+        - example: 2000
+        - description: "attester's inclusion_delay reward in gwei (Optional)"
```
michaelsproul, line 35: we can also use `required` to signal that the other fields are mandatory

apis/beacon/attestations/rewards.yaml
```diff
@@ -0,0 +1,60 @@
+get:
+  operationId: getAttestationsRewards
+  summary: Get attestations rewards
+  description: Retrieves rewards info for attestations for validators specified in id. If none is specified, return rewards info for every validators.
```
michaelsproul, line 4:
```suggestion
  description: Retrieve attestation reward info for validators specified by `id`. If no `id` is specified, return reward info for every validator.
```

apis/beacon/attestations/rewards.yaml
```diff
@@ -0,0 +1,60 @@
...
+  tags:
+    - Beacon
+    - Rewards
```
michaelsproul, line 7: CI is complaining that this needs to be part of the global tags: https://github.com/ethereum/beacon-APIs/actions/runs/3438408607/jobs/5774102666

apis/beacon/attestations/rewards.yaml
```diff
@@ -0,0 +1,60 @@
...
+    "200":
+      description: "Success"
+      content:
+        application/json:
+          schema:
+            title: GetAttestationsRewardsResponse
+            type: object
+            properties:
+              data:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/AttestationsRewards'
```
michaelsproul, line 36: As well as the `data` we should also return `execution_optimistic` and `finalized` properties. The `finalized` one was just added recently in https://github.com/ethereum/beacon-APIs/pull/254.

apis/beacon/blocks/rewards.yaml
```diff
@@ -0,0 +1,47 @@
+get:
+  operationId: getBlockRewards
+  summary: Get block rewards
+  description: Retrieves block rewards info in a slot
```
michaelsproul, line 4:
```suggestion
  description: Retrieve block reward info for a single block
```

apis/beacon/blocks/rewards.yaml
```diff
@@ -0,0 +1,47 @@
...
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            InvalidRequest:
+              value:
+                code: 400
+                message: "Invalid request to retrieve block rewards info"
+    "404":
+      description: "Block not found"
```
michaelsproul, line 36:
```suggestion
      description: "Block or required state not found"
```

apis/beacon/blocks/rewards.yaml
```diff
@@ -0,0 +1,47 @@
...
+      content:
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+          examples:
+            InvalidRequest:
+              value:
+                code: 404
+                message: "Block not found"
```
michaelsproul, line 45:
```suggestion
                message: "Block or required state not found"
```

types/rewards.yaml
```diff
@@ -0,0 +1,71 @@
...
+    source:
+      allOf:
+        - $ref: "./primitive.yaml#/Int64"
+        - example: 4000
+        - description: "attester's reward for source vote in gwei"
+    inclusion_delay:
+      allOf:
+        - $ref: "./primitive.yaml#/Uint64"
+        - example: 2000
+        - description: "attester's inclusion_delay reward in gwei"
```
michaelsproul, line 36:
```suggestion
        - description: "attester's inclusion_delay reward in gwei (phase0 only)"
```

I think it's still good to indicate that this is optional, else consumers of the API might wonder why it's missing

michaelsproul (commented):
We're getting super close to being ready to merge. I think this is probably the final round of review from me before we can pass it to the other maintainers for merging!

michaelsproul (dismissed):
Ship it!

rolfyone:
as a nit, can we updated the endpoints to : 
`/eth/v1/beacon/rewards/blocks/{block_id}`
`/eth/v1/beacon/rewards/attestations/{epoch}`

kevinbogner:
Thanks @rolfyone 
I changed the endpoints and the folder structure. It should look cleaner now

zah:
In the Nimbus codebase, we have a set of routines which can produce not only the obtained rewards, but also the maximum reward that was possible if your attestations were perfect. This allows the creation of UIs and reports which focus on the deviation from perfect behavior. In other words, if a validator produces a wrong head vote for example (while hitting source and target correctly), this can be seen as an opportunity cost (a negative score) instead of as a positive increase of the balance.

michaelsproul:
@zah Great idea! I think all the clients will end up with that data on-hand while computing the attestation rewards, it's just a matter of how to present it.

We could take advantage of the fact that the rewards are fixed per effective balance with something like:

```json
{
    "ideal_rewards": [
        {
            "effective_balance": "1000000000",
            "head": "2500",
            "target": "5000",
            "source": "5000"
        },
        {
            "effective_balance": "2000000000",
            "head": "2500",
            "target": "5000",
            "source": "5000"
        }
    ],
    "rewards": [
        {
            "validator_index": "0",
            "effective_balance": "32000000000",
            "head": "0",
            "target": "5000",
            "source": "5000",
        }
    ],
}
```

Or maybe a `null`-able field per reward component?

```json
{
    "rewards": [
        {
            "validator_index": "0",
            "head": "0",
            "ideal_head": "2500",
            "target": "5000",
            "source": "5000",
        },
        {
            "validator_index": "1",
            "head": "0",
            "ideal_head": "2500",
            "target": "-5000",
            "ideal_target": "5000",
            "source": "-5000",
            "ideal_source": "5000",
        }
    ],
}
```

This has the advantage of being smaller over the wire when the majority of requested validators are behaving optimally, however it introduces more duplication when there is a lot of sub-optimal voting. I think I have a slight preference for the first `effective_balance` approach.

Another option would be an entirely separate API for the ideal reward information. However I don't think this would really be simpler, and it would likely be less efficient.

rolfyone (approved):
LGTM

PR #261, opened 2022-11-08
Fix typos

nflaig:
Just fixed some typos I found when looking through the specs.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -89,7 +89,7 @@ get:
             light_client_optimistic_update:
               description: The node's latest known `LightClientOptimisticUpdate` has been updated
               value: |
-                event: light_client_optimimstic_update
+                event: light_client_optimistic_update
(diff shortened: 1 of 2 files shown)


apis/eventstream/index.yaml
```diff
@@ -89,7 +89,7 @@ get:
             light_client_optimistic_update:
               description: The node's latest known `LightClientOptimisticUpdate` has been updated
               value: |
-                event: light_client_optimimstic_update
+                event: light_client_optimistic_update
```
nflaig, line 92: Is renaming the event a potential problem? I checked the [lodestar](https://github.com/ChainSafe/lodestar) implementation and they use the proper value `light_client_optimistic_update` without the typo in the code.

michaelsproul: IMO renaming to remove a typo is OK, if a client actually implemented the typo'd version I think it would be sensible to deprecate/remove the typo version.

michaelsproul (approved):

ralexstokes (approved):

apis/eventstream/index.yaml
dapplion: Ok with removing the typo :+1:

PR #263, opened 2022-11-09
Change forkchoice checkpoints to epochs

rkapka:
In https://github.com/ethereum/beacon-APIs/pull/232 it was agreed that we will return justified and finalized epochs instead of checkpoints. In https://github.com/ethereum/beacon-APIs/pull/232/commits/630fba53c11208e0ce21582b8b98d108b7e3a61b field names were changed, but types remained the same. For an unknown reason https://github.com/ethereum/beacon-APIs/pull/232/commits/15b6561d0dccd4577b6725994f6ea733eb808046 changed field names back to checkpoints and that's what got merged. This PR sets the correct names and types.

### DIFF
--- a/apis/debug/fork_choice.yaml
+++ b/apis/debug/fork_choice.yaml
@@ -13,12 +13,12 @@ get:
             title: GetForkChoiceResponse
             type: object
             description: "Debugging context of fork choice"
-            required: [justified_checkpoint, finalized_checkpoint]
+            required: [justified_epoch, finalized_epoch]
             properties:
-              justified_checkpoint:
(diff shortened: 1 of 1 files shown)


rolfyone:
@tbenr  ?

tbenr:
I can't recall the details but I thought that prysm wasn't tracking full checkpoints at node level but it was in the general forkchoice state, so we set epochs only at node level but full checkpoints at the top.
Are you proposing this for consistency with `Node` or it is a problem for prysm?

rkapka:
Oh, this is expected? Thanks for letting me know. We can close the issue.

PR #265, opened 2022-11-17
feat: add lightclient sync_committee_roots api

tuyennhv:
**Motivation**
- Support [kevlar](https://github.com/lightclients/kevlar) optimistic sync which needs the sync committee roots from different provers

**Description**
- Given `start_period` and `count` parameters, the api should return respective sync committee roots 
- If sync committee is not found for one of periods, return 404 - not found

part of https://github.com/ChainSafe/lodestar/issues/4733


rolfyone:
I think I need more of an understanding of what's actually the intent of this api...

I'm not sure I can decipher what start period is, and there's no description, which really doesn't help a consumer, or me, as a person that needs to potentially implement this...

If we could start with a use case for how it's intended to be used, then we probably need to look at making sure people that need to implement this would be able to understand what's being asked for and how to get that information...

My assumption is that you'll ask for effectively committees in a range from epoch X , and a count, so to epoch Y (which is X+count)

Then we'd go and get hashes for those epochs? which seems like a lot of work when they're the same for 256 epochs... so that doesn't make sense, and this is where I think I need a better understanding of the intent...

So I guess direct questions
 - what is `start_period`?
 - how does it relate to `count`?

Any further context would be great, I did read the ticket but I didn't find it overly helpful for context on how this will fit together and how I might as an implementer get this data for you...

rolfyone:
I'll plead ignorance in knowledge of whether we currently store sync committee period number...

My assumption is that you'll pretty much need to be an archive node to serve this data... Teku for example is non-archive by default, so it won't easily be able to retrieve old sync committee period information which is based on state...so if you were to request anything older than current, and count forward, it'll likely fail the entire request... It sounds like this is what you're expecting...

In terms of architecture, given that most beacon-apis aren't exposed publicly, I'm expecting that you're running this against known nodes that can be setup in a way that allows these requests to be serviced...

Most of this bit is just for my knowledge, I'm just getting an understanding of how this brick fits in the wall :)

apis/beacon/light_client/sync_committee_roots.yaml
```diff
@@ -0,0 +1,67 @@
...
+            MissingStartPeriodValue:
+              value:
+                code: 400
+                message: "Missing `start_period` value"
+            MissingCountValue:
+              value:
+                code: 400
+                message: "Missing `count` value"
+    "404":
+      description: "One of periods not found"
```
rolfyone, line 58: `One or more periods could not be found`

rolfyone: 404 isn't a great code though, its the same as if you hit a route thats not implemented...

rolfyone:
Closing as stale, we can re-open a current one if/when requried.

PR #268, opened 2022-11-22
chore: pin redocli version

mpetrunic:
cc @rolfyone

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
-    - run: npm i -g @apidevtools/swagger-cli@4 @redocly/cli
+    - run: npm i -g @apidevtools/swagger-cli@4 @redocly/cli@1.0.0-beta.111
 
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM, does look like a bug in the linter, so this is a reasonable measure.

jimmygchen:
Hi @rolfyone @mpetrunic do you recall the reason for pinning the version and what the linter bug was? We're introducing the same linter to [builder-specs](https://github.com/ethereum/builder-specs), so it'd be useful to keep track of this so we know when to upgrade. Thanks!

mpetrunic:
> Hi @rolfyone @mpetrunic do you recall the reason for pinning the version and what the linter bug was? We're introducing the same linter to [builder-specs](https://github.com/ethereum/builder-specs), so it'd be useful to keep track of this so we know when to upgrade. Thanks!

https://github.com/ethereum/beacon-APIs/actions/runs/3523507907/jobs/5907749543

This should work since type field was present in reference object

jimmygchen:
Ahh that one, I've been getting that error locally all the time.  Thank you for the quick response!

PR #271, opened 2022-11-30
Update types to support EIP-4844

jimmygchen:
Add types to support new fields in EIP-4844, main changes are:
- add `blob_kzg_commitments` to `BeaconBlockBodyCommon`
- add `excess_data_gas` to `ExecutionPayloadCommon`

Consensus spec:
https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#beaconblockbody


0xGabi (changes requested):
Well done. I left a couple of comments. And I notice we were missing some `eip4844` enum values so I create a PR against your branch with the addition: https://github.com/jimmygchen/beacon-APIs/pull/1

One thing I was considering is if we want to add an event stream for `blob_kzg_commitments`:  https://github.com/ethereum/beacon-APIs/blob/master/apis/eventstream/index.yaml#L18-L34

types/eip4844/block.yaml
```diff
@@ -0,0 +1,95 @@
...
+        items:
+          $ref: '../deposit.yaml#/Deposit'
+      voluntary_exits:
+        type: array
+        items:
+          $ref: '../voluntary_exit.yaml#/SignedVoluntaryExit'
+      sync_aggregate:
+        $ref: '../altair/sync_aggregate.yaml#/Altair/SyncAggregate'
+      blob_kzg_commitments:
+        $ref: '../primitive.yaml#/KZGCommitment'
```
0xGabi, line 38: This is `List[KZGCommitment, MAX_BLOBS_PER_BLOCK]`. What do you think of following a similar convention of creating a new file on the types folder for the `blob_kzg_commitments`, like [bls_to_execution_change.yaml](https://github.com/ethereum/beacon-APIs/blob/master/types/bls_to_execution_change.yaml)?

0xGabi (changes requested):
I missed this one on the first pass.

types/eip4844/execution_payload.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+  ExecutionPayload:
+    allOf:
+      - $ref: '#/Eip4844/ExecutionPayloadCommon'
+      - type: object
+        properties:
+          transactions:
+            $ref: './transactions.yaml#/Eip4844/Transactions'
+          withdrawals:
+            $ref: './withdrawals.yaml#/Eip4844/Withdrawals'
```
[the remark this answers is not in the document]
jimmygchen: Good catch! Updated to reference Capella sections as there's no change in EIP-4844

types/eip4844/block.yaml
jimmygchen: Ah good catch thanks, updated `type` to array. I actually started with creating a new file for  `blob_kzg_commitments ` but realised I only needed a primitive type, and looking at previous changes I guessed the convention is to add primitive types to `primitive.yaml`, but I could be completely wrong 😅

jimmygchen:
> Well done. I left a couple of comments. And I notice we were missing some `eip4844` enum values so I create a PR against your branch with the addition: [jimmygchen#1](https://github.com/jimmygchen/beacon-APIs/pull/1)
> 
> One thing I was considering is if we want to add an event stream for `blob_kzg_commitments`: https://github.com/ethereum/beacon-APIs/blob/master/apis/eventstream/index.yaml#L18-L34

I'm not sure about the usage of having a `blob_kzg_commitments` event stream, what's in your mind?

0xGabi:
I thought that it could be helpful to know when the node received a blob transaction so the event listener can perform validations like [`verify_kzg_commitments_against_transactions`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#verify_kzg_commitments_against_transactions).

jimmygchen:
> I thought that it could be helpful to know when the node received a block with a blob transaction so the event listener can perform validations similar to [`verify_kzg_commitments_against_transactions`](https://github.com/ethereum/consensus-specs/blob/dev/specs/eip4844/beacon-chain.md#verify_kzg_commitments_against_transactions).

These are Server-Sent-Events from beacon node though, and are for consumers of the Beacon API. For the beacon node itself it should already be able to handle blob messages sent via gossip, if that's what you mean? 

The p2p interfaces are part of the consensus spec [here](https://github.com/ethereum/consensus-specs/blob/30e975806f7b5d43e35378ccc3dc428ba09dddd5/specs/eip4844/p2p-interface.md#beacon_block_and_blobs_sidecar)

0xGabi:
> For the beacon node itself it should already be able to handle blob messages sent via [gossip](https://github.com/ethereum/consensus-specs/blob/30e975806f7b5d43e35378ccc3dc428ba09dddd5/specs/eip4844/p2p-interface.md#beacon_block_and_blobs_sidecar), which is specified as part of the consensus spec - if this is what you mean?

No, as we discussed async I was thinking about having a `blob_kzg_commitments` API endpoint so external components can work with the data (i.e. validations).

rolfyone:
Currently all the SSE are JSON with a leading line to say what kind of data. I'm not sure how sending blobs would go, potentially in the first instance it's better to get the types in, and follow up the SSE in a separate issue?

If there's no great call for an SSE, potentially we could live without it in the first instance, unless there's an actual use for it we know about...

jimmygchen:
Thanks @rolfyone - I agree, I'm not aware of any discussions on blobs over SSE before this and is probably not a prerequisite for 4844, and we can follow up in a separate issue if necessary.

0xGabi:
Make sense @rolfyone, thank you for your input.

jimmygchen:
@0xGabi @rolfyone I've just added `SignedBeaconBlockAndBlobsSidecar` schema as well - there were some discussions on the [builders-spec PR](https://github.com/ethereum/builder-specs/pull/58#discussion_r1047758214), and we think it belongs here, although it's not currently consumed by any endpoints yet - let me know if you think otherwise.

ralexstokes:
to follow up on @jimmygchen 's point, how do we expect users to download blobs from their local CL nodes?

I would assume we would have a beacon API endpoint to support this

0xGabi (changes requested):
LGTM. I left one small request for change.

> how do we expect users to download blobs from their local CL nodes?

I think we should only allow to download a list of `blobs_sidecar` using the `block_id` identifier. For example: `/eth/v2/beacon/blobs_sidecar/{block_id}`. Downloading a single blob could be handle with a higher level library and reduce CL complexity. 

On a final note, I propose to merge the current changes and address the new API endpoint on a new PR.

types/eip4844/block_and_blobs_sidecar.yaml
```diff
@@ -0,0 +1,38 @@
+EIP4844:
+  SignedBeaconBlockAndBlobsSidecar:
```
0xGabi, line 2: One last thing, I believe we can go ahead and remove this type definition.

types/bellatrix/execution_payload.yaml
```diff
@@ -45,6 +45,7 @@ Bellatrix:
     allOf:
       - $ref: '#/Bellatrix/ExecutionPayloadCommon'
       - type: object
+        additionalProperties: false
```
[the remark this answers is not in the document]
jimmygchen: otherwise the example validation fails, because the each of them have to match exactly one schema
without this, the 4844 paylaod would match all 3 schemas, because additional properties are ignored

types/eip4844/block_and_blobs_sidecar.yaml
```diff
@@ -0,0 +1,38 @@
...
+    type: object
+    description: "The `BlobsSidecar` object from the EIP-4844 CL spec."
+    properties:
+      beacon_block_root:
+        $ref: "../primitive.yaml#/Root"
+      beacon_block_slot:
+        $ref: "../primitive.yaml#/Uint64"
+      blobs:
+        type: array
+        maxItems: 16
```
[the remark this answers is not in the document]
jimmygchen: nice catch!

jimmygchen:
@0xGabi @ralexstokes thanks for the review guys! I have addressed the comments.
Agreed with you both on the blobs endpoint, happy to do it in a separate PR it's ok?

jimmygchen:
This PR has already grown quite big (Capella + EIP-4844 types), so I've created a separate issue to track the new endpoint for blob download: #282

0xGabi (dismissed):
LGTM

types/eip4844/block.yaml
```diff
@@ -0,0 +1,101 @@
...
+    allOf:
+      - $ref: '#/EIP4844/BeaconBlockBodyCommon'
+      - type: object
+        properties:
+          execution_payload_header:
+            $ref: './execution_payload.yaml#/EIP4844/ExecutionPayloadHeader'
+          bls_to_execution_changes:
+            type: array
+            items:
+              $ref: '../bls_to_execution_change.yaml#/SignedBLSToExecutionChange'
```
rolfyone, line 83: This can probably be in common... ordering is less important in JSON, so I think we can probably just put it in common and not worry that `execution_payload_header` is after everything else in the json definition...

types/eip4844/block.yaml
```diff
@@ -0,0 +1,101 @@
...
+    allOf:
+      - $ref: '#/EIP4844/BeaconBlockBodyCommon'
+      - type: object
+        properties:
+          execution_payload:
+            $ref: './execution_payload.yaml#/EIP4844/ExecutionPayload'
+          bls_to_execution_changes:
+            type: array
+            items:
+              $ref: '../bls_to_execution_change.yaml#/SignedBLSToExecutionChange'
```
rolfyone, line 52: This goes away if the other one moves to common

types/eip4844/block.yaml
jimmygchen: Thanks @rolfyone ! I've now moved this to common for both Capella and EIP-4844

types/eip4844/block.yaml
jimmygchen: Done

Seyvick (approved):

Seyvick (approved):

Seyvick (approved):

Seyvick (approved):

rolfyone (approved):
LGTM, may merge to a branch though so that we can progress this while we're waiting for a non EIP name...

rolfyone:
just be aware - new branch: EIP4844. Once we've got a better name for the fork, this can come onto master but this will unblock any further work needed for EIP-4844

Seyvick:
Ok, thank's!!!

PR #275, opened 2022-12-05
Changed bls to execution api to accept multiple submissions

rolfyone:
The bls_to_execution_change process is probably better modelled as a batch of changes to make.

Changed the api to model this, allowing for messages in responses in case there are errors that need specific reporting.

Given that this API isn't yet released, I made the change in place to the existing api, rather than versioning, and it's unlikely that we need both single and multi-submission, it's easy enough to wrap a single in an array object.


mcdee (dismissed):

apis/beacon/pool/bls_to_execution_changes.yaml
```diff
@@ -23,26 +23,54 @@ get:
...
+                      description: |
+                        - accepted: The bls_to_execution_change was accepted, placed in the local pool, and broadcast to the network
+                        - error: The bls_to_execution_change was not accepted. More context may be supplied in the associated message.
+                      enum:
+                        - accepted
+                        - error
+                      example: accepted
+                    message:
+                      type: string
+                      description: error message if status == error
```
michaelsproul, line 65: Maybe we should model this after the other batched pool endpoint rather than the key-manager API?

Some advantages of consistency with the `/pool/attestations` API:

- User gets an error status code if any of the messages fail. I think this is slightly clearer/more obvious when submitting manually, and could work well with `curl --fail-with-body` in simple scripts, etc. The failed messages are still in the error response in case of error.
- Consistency makes the API (slightly) easier to use and code against.

rolfyone: Fair point. keymanager is what I had working yaml for so it was easier, but I can have a look at the attestation pool and refactor closer to that...

g11tech:
I wouldn't be keen on this array submission format as it breaks semantics with other submissions like "validator exit" which are single object. 
Also a too big array can easily result into http timeout (which is why these submissions need to be byte sized). With respect to any too supporting multiple submissions, it may just call this api multiple times

rolfyone:
> Also a too big array can easily result into http timeout (which is why these submissions need to be bite sized). With respect to any too supporting multiple submissions, it may just call this api multiple times

I agree that people can go and create large payloads, and that can cause timeouts potentially, but it greatly simplifies the flow if the incoming data is the output of a tool like `ethdo` where an array of submissions have been given.

Currently, client teams would need to take that output, and break it down into individual entries and call the api for each one, where it would seem much simpler to allow that output to just be sent via curl to the endpoint.

Also http timeouts are client side driven, so it could easily be extended by the caller if that was the main issue... Being a POST, larger submissions do sometimes take a while.

Finally, I would suggest he naming of the endpoint indicates that there are multiple, so really the POST should have been `bls_to_execution_change` given its functionality if we were to be consistent with other parts of the api.

One option would be to have an endpoint that has single and one that has multiple, but it seems like over-engineering a solution, when this seems to fit the overall model better anyway.

mcdee:
Note that at this stage it is easy enough to alter `ethdo` to provide a different output format as per the suggestion by @ajsutton but, as said, it doesn't come without downsides.

Also, note that `ethdo` today abides by the current specification so on an online computer you can run `ethdo validator credentials set` with the `change-operations.json` file in the current directory and it will read the file in and send them to the beacon node one-by-one.  Not quite as easy as carrying out a simple `curl`, but not a million miles away.

Ultimately I have a slight preference for this PR to go through, but we can probably smooth over the UX enough if it doesn't to make the overall process relatively painless for users.

dapplion:
Submitting these messages should be really fast. Submitting one by one or in parallel requests of say 10 should be almost the same UX than batching. Implementing a batch approach makes implementation more complex thus adding extra delay on delivering the fork. I would prefer for downstream tooling to absorb and manage the complexity.

rkapka:
> Implementing a batch approach makes implementation more complex thus adding extra delay on delivering the fork.

This delay is negligible IMO. On the other hand I personally like when similarly functioning pieces of the system look alike - it reduces the amount of thought one has to put in - meaning that I slightly lean against this change to keep the API consistent with objects like exits.

rolfyone:
Sounds like this isn't worth the argument. I'll raise a separate PR to move the `changes.md`, and will probably just implement a client specific version to make life easier for users or something, depending on how it works out with ethdo etc.

PR #279, opened 2022-12-08
re-added bls_to_execution_changes as a batch

rolfyone:
After further discussion on discord about #275 which was closed, there's more positive sentiment about this now, so re-opening.

The user flow would be generating withdrawals potentially offline with a tool such as ethdo, and taking the generated file to submit on a BN.  Ethdo as a specific example does generate this structure already, which would allow users not to have to alter the file once it has been generated.

from original:

The bls_to_execution_change process is probably better modelled as a batch of changes to make.

Changed the api to model this, allowing for messages in responses in case there are errors that need specific reporting.

Given that this API isn't yet released, I made the change in place to the existing api, rather than versioning, and it's unlikely that we need both single and multi-submission, it's easy enough to wrap a single in an array object.


rkapka (approved):

rkapka (approved):

michaelsproul (approved):
Great, consensus seems to be in favour of the batch!

Let's merge this so we can move ahead with the changes for the next devnets

PR #284, opened 2022-12-22
Add version property to getPoolBLSToExecutionChanges

dapplion:
A beacon node must _guess_ with what fork BLSToExecutionChanges messages are signed on. Assuming the current fork is a good guess, but it's still a guess. I find it sound to allow for clients to specify a fork, which allows:
- error-less submission around fork boundaries
- allow to submit BLSToExecutionChanges early ahead of the fork. In Lodestar's operation pool we hold early messages until their fork comes.

Since the API is batched this PR assumes all messages are signed at the same fork, which is a good assumption IMO. Else it would be a bit ugly to support, while this PR mimics the response format of versioned APIs.

### DIFF
--- a/apis/beacon/pool/bls_to_execution_changes.yaml
+++ b/apis/beacon/pool/bls_to_execution_changes.yaml
@@ -13,6 +13,8 @@ get:
             title: GetPoolBLSToExecutionChangesResponse
             type: object
             properties:
+              version:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Version'
               data:
                 type: array
                 items:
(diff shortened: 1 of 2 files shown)


rolfyone:
We can't solve a spec problem with an api change...  I get the desire, but the correct fix is probably to fix the spec...

dapplion:
> We can't solve a spec problem with an api change... I get the desire, but the correct fix is probably to fix the spec...

Why is it a spec problem? As an API client you are just communicating to a node at which fork your messages should be included. To which fork a BLSToExecutionChanges is well defined everywhere except in the API:
- If from a block: fork = block's fork
- If from gossip: fork = gossip topic's fork
- If from API = ???

rolfyone:
This is redundant potentially now that we're signing differently (ie. fixed in spec) - Happy to close @dapplion ?

Issue #282, opened 2022-12-16
Add new endpoint to support downloading of blobs (EIP-4844)


jimmygchen:
Raised by @ralexstokes in this PR #271

The above PR adds new types to support EIP-4844l, however there is currently no endpoint for users to download blobs from their local CL nodes.

@0xGabi suggested the following:

> I think we should only allow to download a list of `blobs_sidecar` using the `block_id` identifier. For example: `/eth/v1/beacon/blobs_sidecar/{block_id}`. Downloading a single blob could be handle with a higher level library and reduce CL complexity.

0xGabi:
It was a mistake on my side. I copied one example from the beacon-api website and didn't realize it was a `v2` example. We have to start with `v1` 👍

jimmygchen:
Thanks, updated the description :)

jimmygchen:
I have implemented this endpoint in Lighthouse, [PR here](https://github.com/sigp/lighthouse/pull/3830). Will pick up the Beacon API spec early next week.

jimmygchen:
Completed in #286

PR #286, opened 2023-01-09
Add blob download endpoint (getBlobs)

jimmygchen:
## Description

Addresses #282. Add `/eth/v1/beacon/blobs/{block_id}` endpoint to support retrieval of blobs by block id.


jimmygchen:
From what I can see the request / response structures look similar, just some differences in paths. I think a combination of the two makes sense to me (e.g. /eth/v1/beacon/blobs/sidecar/{block_id}) for the following reasons:
1/ It feels like the blob sidecar belongs under beacon as it's closely related to the beacon block
2/ Having blobs/sidecars instead of blobs_sidecar may provide more flexibility, although I don't see any other potential endpoints that may fall under blobs.. so no strong opinion on this.

I'm still pretty new to this so I might be missing something - keen to hear suggestions / opinions.

arnetheduck:
One potential issue of note: in the underlying libp2p protocol, we don't support fetching finalized items by root, instead only supporting by-slot lookups - it might be prudent to note this in this spec as well, that sidecar retrieval for finalized blocks should by done by slot else the underlying index might not be available.

Doing it this way allows clients to implement this API using linear storage which is cheaper and more efficient and more closely matches the requirements that clients *have* to support in order to implement the mandatory parts of the protocol.

mcdee:
I'd be inclined to go for just  `/beacon/blobs` as the path, I don't see the `_sidecar` piece adding any useful information (it's a information about how we're treating the blobs, not what they are).  There's also an argument to have just `/blobs`, especially if the plan is to make them available from places other then the beacon node, although I'm not overly bothered about that one.

terencechain:
Added Prysm implementation here: https://github.com/prysmaticlabs/prysm/pull/11621/commits/50b672a4dbbffa4d17ac4853a5f604adcc8c7ce4

jimmygchen:
> One potential issue of note: in the underlying libp2p protocol, we don't support fetching finalized items by root, instead only supporting by-slot lookups - it might be prudent to note this in this spec as well, that sidecar retrieval for finalized blocks should by done by slot else the underlying index might not be available.
> 
> Doing it this way allows clients to implement this API using linear storage which is cheaper and more efficient and more closely matches the requirements that clients _have_ to support in order to implement the mandatory parts of the protocol.

@arnetheduck thanks for this! I don't fully understand the implication of this on the Beacon API, would appreciate if you could elaborate a bit more - if the blob is already available in CL storage and p2p lookup isn't required, would this still be an issue?

arnetheduck:
> blob is already available in CL storage

in the core protocol, there is no requirement to index finalized data by hash, only by slot - this applies equally to blobs as it does to blocks. 

Because finalized data is strictly linear, the index-by-hash is not needed and looking things up by hash is comparatively inefficient, as is building and maintaining the hash index itself. 

"CL storage" for finalized data can be implemented as a flat file to serve core protocol needs - by suggesting consumers to use slots, the same efficiency can be had in the REST API as well, which saves on disk space, processing and complexity, in the CL.

jimmygchen:
> I'd be inclined to go for just `/beacon/blobs` as the path, I don't see the `_sidecar` piece adding any useful information (it's a information about how we're treating the blobs, not what they are). There's also an argument to have just `/blobs`, especially if the plan is to make them available from places other then the beacon node, although I'm not overly bothered about that one.

Good point on `_sidecar` - I think I'm just getting used to the term as a spec type now.  One reason I think it could be helpful is that it tells user that the endpoint returns the `BlobsSidecar` spec object rather than just the blobs. If that's not a strong enough reason, I think we're ok with dropping `_sidecar`.

beacon-node-oapi.yaml
```diff
@@ -95,6 +95,8 @@ paths:
     $ref: "./apis/beacon/blocks/root.yaml"
   /eth/v1/beacon/blocks/{block_id}/attestations:
     $ref: "./apis/beacon/blocks/attestations.yaml"
+  /eth/v1/beacon/blobs_sidecars/{block_id}:
```
rolfyone, line 98: I'd actually be inclined to make this the same pattern as attestations, so the path would be more like
`/eth/v1/beacon/blocks/{block_id}/blobs`

mcdee: Wouldn't that imply that the blobs are part of the block structure, though?

rolfyone:
We should add to top level `CHANGES.md` too

jimmygchen:
@rolfyone @mcdee
I've taken your suggestions to remove `_sidecar` from the path and added the endpoint to `CHANGES.md`. I originally thought about having `blocks` in the path as well, but I think @mcdee made a good point and blobs aren't part of the block structure.

@arnetheduck thanks a lot for the explanation, If I understand you correctly, you're suggesting to not support lookup by root for blobs (and perhaps blocks too) as it would allow clients implement this more efficiently. I made the assumption we would want the same lookup (block id) for blobs as well, as looking up by root is already supported for blocks in the Beacon API. I don't really understand the libp2p protocol and storage implementations enough to make a comment about this (but interested to find out more if there's any resources you could point me to) - I'll wait for others to chime in on this.

arnetheduck:
> you're suggesting to not support lookup by root for blobs (and perhaps blocks too) as it would allow clients implement this more efficiently.

Perhaps more weakly than this, ie "recommend" that consumers look up by slot - it's possible that some clients build a by-root index for the entire history (consider archive nodes for example) but it is certainly more "compatible" to not rely on it and instead use `slot` (which also can offer better performance) - it would indeed be the case that the same recommendation should be added to block lookups - might get to this at some point unless someone beats me to it.

jimmygchen:
Just going through the changes in the "Free the blobs" PR https://github.com/ethereum/consensus-specs/pull/3244, and will be updating this branch shortly.

realbigsean:
@jimmygchen 

I don't think this will have to change too much, I think it should probably return the same struct we return in the `BlobsByRange` request as this is likely how we'll store blobs in our database. This is what it looks like right now https://github.com/ethereum/consensus-specs/pull/3244/files#diff-21cd1dde2a4815e714ef42d0cefa1bbd2997a4358ecf6e9e51025332c1d642fdR222-R224

jimmygchen:
Thanks @realbigsean !

I've updated the endpoint to return [`BlobSidecar`](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blobsidecar) container from latest CL spec.

jimmygchen:
> Perhaps more weakly than this, ie "recommend" that consumers look up by slot - it'

^ Added this to endpoint description https://github.com/ethereum/beacon-APIs/pull/286/commits/1d8e59965f2a2a2c95c88113b87447ad2d680697

jimmygchen:
Apologies the the diffs is a bit messy as we currently have a couple of unmerged deneb PRs.

A cleaner diff view here for anyone interested: https://github.com/jimmygchen/beacon-APIs/pull/2/files

dapplion (approved):

types/deneb/block.yaml
```diff
@@ -0,0 +1,97 @@
+Deneb:
+  BeaconBlockBodyCommon:
+    # An abstract object to collect the common fields between the BeaconBlockBody and the BlindedBeaconBlockBody objects
+    type: object
+    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/master/specs/deneb/beacon-chain.md#beaconblockbody) object from the CL Deneb spec."
+    properties:
+      randao_reveal:
+        allOf:
+          - $ref: '../primitive.yaml#/Signature'
+          - description: "The RanDAO reveal value provided by the validator."
```
rkapka, line 10: nitpick: The spelling `RanDAO` is quite unconventional. It's usually `Randao` or `RANDAO`

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,53 @@
+get:
+  operationId: getBlobs
+  summary: Get blobs
+  description: |
+    Retrieves blobs for a given block id.
+    Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.
+    It is recommended that blob retrieval for finalized blocks should by done by slot else the underlying index might not be available.
+  tags:
+    - Beacon
+  parameters:
```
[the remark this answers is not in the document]
jimmygchen: > For Prysm, we currently only index blobs by root and serve the BlobSidecarsByRange p2p request by finding the canonical chain of blocks for the given slot range and then looking up each set of blobs for a given slot by the canonical block's root. For us, requesting by slot would add some complexity to our API endpoint rather than making things simpler. That said, it's something we could support if other clients need it.

I _think_  Lighthouse is doing something similar to what you've described, I'm not sure if it's applicable either 🤔 if this is implementation-level detail, I'm inclined to leave it out of the spec - keen to hear thoughts!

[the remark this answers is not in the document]
jimmygchen: cc @realbigsean

jimmygchen:
> For Prysm, we currently only index blobs by root and serve the BlobSidecarsByRange p2p request by finding the canonical chain of blocks for the given slot range and then looking up each set of blobs for a given slot by the canonical block's root. For us, requesting by slot would add some complexity to our API endpoint rather than making things simpler. That said, it's something we could support if other clients need it.

I _think_  Lighthouse is doing something similar to what you've described, I'm not sure if it's applicable either 🤔 if this is implementation-level detail, I'm inclined to leave it out of the spec - keen to hear thoughts!

cc @realbigsean

rkapka:
This would require changing the `block_id` parameter from type `BlockId` to something else. If we use `BlockId`, we should support all possible variations, including slot number.

apis/beacon/blobs/blobs.yaml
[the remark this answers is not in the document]
jimmygchen: Added `indices` query parameter, thanks!

beacon-node-oapi.yaml
jimmygchen: Updated to `/eth/v1/beacon/blobs/{block_id}`

jimmygchen:
> Perhaps more weakly than this, ie "recommend" that consumers look up by slot - it's possible that some clients build a by-root index for the entire history (consider archive nodes for example) but it is certainly more "compatible" to not rely on it and instead use `slot` (which also can offer better performance) - it would indeed be the case that the same recommendation should be added to block lookups - might get to this at some point unless someone beats me to it.

@arnetheduck This seems to vary between implementations (Prysm & Lighthouse currently index by block root) - hence isn't true in some cases, so I'm inclined to leave this of the API spec since the implementation varies. Happy to hear more thoughts though.

jimmygchen:
> This would require changing the `block_id` parameter from type `BlockId` to something else. If we use `BlockId`, we should support all possible variations, including slot number.
> 
> I wonder if there is a sensible use case of wanting a blob by slot. If there is, then (by Murphy's Law) at least one user will ask for it.

I would prefer to keep `block_id` if we're already implementing most of the block_id variants, and if it doesn't add too much complexity / effort to implement look up by slot. There's already usage of looking up by slots in [blob explorer](https://github.com/Blobscan/blobscan-indexer/blob/ce89055945f6784a5cdbadb48312f6d66ad36bdf/src/index.js#L75) and [blob-indexer.rs](https://github.com/Blobscan/blobscan-indexer.rs/blob/master/src/beacon_chain/mod.rs#L59).

rkapka:
Let's keep it as is. Looking up by slot is currently supported in Prysm.

apis/beacon/blocks/blinded_blocks.yaml
```diff
@@ -32,6 +32,7 @@ post:
             - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Altair.SignedBeaconBlock"
             - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Bellatrix.SignedBlindedBeaconBlock"
             - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Capella.SignedBlindedBeaconBlock"
+            - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Deneb.SignedBlindedBeaconBlock"
```
rkapka, line 35: I think this is incorrect. We should publish blobs as well. So we need a `SignedBlindedBeaconBlockWithBlobs` type that wraps a `SignedBlindedBeaconBlock` and an array of `SignedBlobSidecar` (which is not defined in this PR), and use that type here.

jimmygchen: @rkapka you're right. This change is made in @realbigsean 's PR here https://github.com/ethereum/beacon-APIs/pull/302
Apologies for the messy diff, we're all kinda dependent on the [deneb rename branch](https://github.com/ethereum/beacon-APIs/pull/289) being merged, which should hopefully happen soon.

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,62 @@
...
+              data:
+                type: array
+                items:
+                  oneOf:
+                    - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlobSidecar"
+                minItems: 0
+                maxItems: 4
+        application/octet-stream:
+          schema:
+            description: "SSZ serialized block bytes. Use Accept header to choose this response type"
```
rolfyone, line 42: what's this returning?
My assumption is that its an ssz array of blobs in the order specified in the query?

eg. if i set `?indicies=1,0` do i get an ssz array of blobs, with blob 1 first, and blob 0 second?

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,62 @@
...
+              message: "Invalid block ID: current"
+    "404":
+      description: "Blob sidecar not found"
+      content:
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+            example:
+              code: 404
+              message: "Blob sidecar not found"
```
rolfyone, line 60: this is fairly reversed...

given the block id is in the path, then i'd expect a 404 to be the block itself not found...

400 would potentially be better for specific blobs not found?

Would a partial availability still be not found for blobs? so if i request `indicies=0,1`, and blob 1 is missing, do i return a 400?

types/primitive.yaml
```diff
@@ -123,9 +123,21 @@ LogsBloom:
...
   format: hex
   pattern: "^0x[a-fA-F0-9]{96}$"
   description: "A G1 curve point. Same as BLS standard \"is valid pubkey\" check but also allows `0x00..00` for point-at-infinity"
   example: "0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a"
+
+KZGProof:
+  type: string
+  format: hex
+  pattern: "^0x[a-fA-F0-9]{96}$"
+  description: "A KZG proof. Same check as `KZGCommitment`"
```
mcdee, line 143: Would be better to make this description explicit rather than refer to another type.

mcdee (changes requested):

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,62 @@
+get:
+  operationId: getBlobs
```
g11tech, line 2:
```suggestion
  operationId: getBlobsSidecar
```

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,62 @@
...
+  summary: Get blobs
+  description: |
+    Retrieves blobs for a given block id.
+    Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.
+    It is recommended that blob retrieval for finalized blocks should by done by slot else the underlying index might not be available.
```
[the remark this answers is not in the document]
g11tech: well lodestar resolves the block first and then uses root or slot to query, so for lodestar it works eitherway

apis/beacon/blobs/blobs.yaml
g11tech: i think if any of the indices is missing, the full api should fail with 400 and may be error indicating the missed index

rolfyone: I realised, its effectively an array return and what we generally do is return an empty array, not an error... so unless we need an error to indicate something was physically missing, following the pattern of empty array means nothing would likely be preferable from a consistency point of view...

apis/beacon/blobs/blobs.yaml
jimmygchen: Yeah I had that originally but dropped it. There was a suggestion a while ago to remove it, and I think it's a fair enough reason, although I also think `blob_sidecars` make sense too given it's the structure we return and consistent with the spec name. I don't have a strong opinion here though.

https://github.com/ethereum/beacon-APIs/pull/286#issuecomment-1379756410

apis/beacon/blobs/blobs.yaml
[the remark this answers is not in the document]
jimmygchen: Thanks, agreed! I'm also inclined to remove this as it isn't true for some beacon nodes (see [comment above](https://github.com/ethereum/beacon-APIs/pull/286#issuecomment-1495287100)) - will remove this sentence.

apis/beacon/blobs/blobs.yaml
jimmygchen: Thanks for raising this, great point - I think it would be good to be consistent with other endpoints that accepts array type query params (e.g. `getStateValidators), i.e. not guaranteeing order, unless there are specific reasons for guaranteeing the order when indices are specified?

From `getStateValidators`:
> If an index or public key does not
    match any known validator, no information will be returned but this will not cause an error.  There are no guarantees for the
    returned data in terms of ordering

apis/beacon/blobs/blobs.yaml
jimmygchen: Agree with the above on consistency - and only return blobs matching the indices, e.g. `indices=1` returns empty array if blob 1 is missing.

if the block itself is not found, then a 404  is returned like @rolfyone suggested. I'll update the message there.

types/primitive.yaml
jimmygchen: Updated in https://github.com/ethereum/beacon-APIs/pull/286/commits/4ccaaa6f018c596c6ebf1227943da952ed7cf29f, thanks!

apis/beacon/blobs/blobs.yaml
jimmygchen: Updated 404 description to "Block not found" in https://github.com/ethereum/beacon-APIs/pull/286/commits/4ccaaa6f018c596c6ebf1227943da952ed7cf29f

apis/beacon/blobs/blobs.yaml
g11tech: i think in previous version the comment made sense as blobs were sort of different from the sidecars aggregate object, but not blobSideCar is an independent object, i means its just not taxing on the mind to understand in a glance what they return.
but yes i wont strongly push it either but weakly in favor of change now

jimmygchen: Thanks @g11tech, I think that sounds reasonable to me, so: 
- operationId: getBlobSidecars
- path: `/eth/v1/beacon/blob_sidecars/{block_id}`

Any other thoughts on this @mcdee @rolfyone?

rolfyone: I'm impartial on this one :)

apis/beacon/blobs/blobs.yaml
rolfyone: i haven't looked at the SSZ, hopefully its easy to get the blob id from the object.. assuming thats fine...

CHANGES.md
```diff
@@ -15,6 +15,7 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#289](https://github.com/ethereum/beacon-APIs/pull/289) `/eth/v1/validator/blinded_blocks/{slot}` added `deneb` blinded block              |                                                  |                                                   |                                                    |                                                 |                                           |
 | [#289](https://github.com/ethereum/beacon-APIs/pull/289) `/eth/v1/validator/blocks/{slot}` added `deneb` block                              |                                                  |                                                   |                                                    |                                                 |                                           |
+| [#286](https://github.com/ethereum/beacon-APIs/pull/286) NEW `/eth/v1/beacon/blobs/{block_id}`                                              |                                                  |                                                   |                                                    |                                                 |                                           |
```
rolfyone, line 18: will want to update this if we're changing path...

eliasaadeh (approved):

apis/beacon/blobs/blobs.yaml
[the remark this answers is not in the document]
jimmygchen: Added in https://github.com/ethereum/beacon-APIs/pull/286/commits/27ac4b8e5486ab00efbef142b1ba3e024151d688

apis/beacon/blocks/blinded_blocks.yaml
jimmygchen: This has been addressed in #302

types/deneb/block.yaml
jimmygchen: Updated in https://github.com/ethereum/beacon-APIs/pull/286/commits/e3dfda2125df409e1fe25fa3ef6299bd5d6a2728. Thanks!

apis/beacon/blobs/blobs.yaml
jimmygchen: Added more details in description to specify the behaviour when `indices` is specified, and fixed the SSZ response description. Commit: 354ff9fa9b80ff2fd68bc5bfdabe2fe3f85b25f1.

@rolfyone what do you mean by getting the blob id from the object?

types/eip4844/state.yaml
```diff
@@ -0,0 +1,88 @@
+EIP4844:
+  BeaconState:
+    type: object
+    description: "The [`BeaconState`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#beaconstate) object from the Eth2.0 EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844`

types/eip4844/execution_payload.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - $ref: '#/EIP4844/ExecutionPayloadCommon'
+      - type: object
+        properties:
+          transactions:
+            $ref: '../capella/transactions.yaml#/Capella/Transactions'
+          withdrawals:
+            $ref: '../capella/withdrawals.yaml#/Capella/Withdrawals'
+
+  ExecutionPayloadHeader:
+    description: "The [`ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#executionpayloadheader) object from the CL EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/execution_payload.yaml
```diff
@@ -0,0 +1,56 @@
+EIP4844:
+  ExecutionPayloadCommon:
+    # An abstract object to collect the common fields between the ExecutionPayload and the ExecutionPayloadHeader objects.
+    type: object
+    description: "The [`ExecutionPayload`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#executionpayload) object from the CL EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+    allOf:
+      - $ref: '../altair/block.yaml#/Altair/BeaconBlockCommon'
+      - type: object
+        properties:
+          body:
+            $ref: '#/EIP4844/BlindedBeaconBlockBody'
+
+  SignedBlindedBeaconBlock:
+    type: object
+    description: "A variant of the the the [`SignedBeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#signedbeaconblock) object envelope from the CL EIP-4844 spec, which contains a `BlindedBeaconBlock` rather than a `BeaconBlock`."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+  BeaconBlockBody:
+    allOf:
+      - $ref: '#/EIP4844/BeaconBlockBodyCommon'
+      - type: object
+        properties:
+          execution_payload:
+            $ref: './execution_payload.yaml#/EIP4844/ExecutionPayload'
+
+  BeaconBlock:
+    description: "The [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/EIP4844/beacon-chain.md#beaconblock) object from the CL EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+    allOf:
+      - $ref: '../altair/block.yaml#/Altair/BeaconBlockCommon'
+      - type: object
+        properties:
+          body:
+            $ref: '#/EIP4844/BeaconBlockBody'
+
+  SignedBeaconBlock:
+    type: object
+    description: "The [`SignedBeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#signedbeaconblock) object envelope from the CL EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+    properties:
+      message:
+        $ref: "#/EIP4844/BeaconBlock"
+      signature:
+        $ref: "../primitive.yaml#/Signature"
+
+  BlindedBeaconBlockBody:
+    description: "A variant of the [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#beaconblockbody) object from the CL EIP-4844 spec, which contains a transactions root rather than a full transactions list."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+    allOf:
+      - $ref: '#/EIP4844/BeaconBlockBodyCommon'
+      - type: object
+        properties:
+          execution_payload_header:
+            $ref: './execution_payload.yaml#/EIP4844/ExecutionPayloadHeader'
+
+  BlindedBeaconBlock:
+    description: "A variant of the the [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#beaconblock) object from the CL EIP-4844 spec, which contains a `BlindedBeaconBlockBody` rather than a `BeaconBlockBody`."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

types/eip4844/block.yaml
```diff
@@ -0,0 +1,97 @@
+EIP4844:
+  BeaconBlockBodyCommon:
+    # An abstract object to collect the common fields between the BeaconBlockBody and the BlindedBeaconBlockBody objects
+    type: object
+    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/master/specs/eip4844/beacon-chain.md#beaconblockbody) object from the CL EIP-4844 spec."
```
rolfyone: link will need to update to `/deneb/` rather than `/eip4844/`

beacon-node-oapi.yaml
```diff
@@ -95,6 +95,8 @@ paths:
...
+    $ref: "./apis/beacon/blobs_sidecars/blobs_sidecar.yaml"
```
rolfyone: do both need to be plurals? is `blob_sidecars` accurate?

if the paths / files / object names can be consistent that'll make it easier to navigate

beacon-node-oapi.yaml
```diff
@@ -311,7 +315,7 @@ components:
     Bellatrix.SignedBlindedBeaconBlock:
       $ref: './types/bellatrix/block.yaml#/Bellatrix/SignedBlindedBeaconBlock'
     ConsensusVersion:
-      enum: [phase0, altair, bellatrix, capella]
+      enum: [phase0, altair, bellatrix, capella, eip4844]
```
rolfyone: I think i'll avoid mentioning it more unless we miss something, but they should all just reference `deneb` where `eip4844` currently referenced

jimmygchen: hey @rolfyone I think you might be looking at earlier commits (before latest master was merged). I'm not seeing any 4844 references on this branch, might have to change the file filter to all commits in the diff view? 
Almost got me though 😅

types/eip4844/state.yaml
jimmygchen: I think you updated all of them in the other PR! :)

apis/beacon/blobs/blobs.yaml
jimmygchen: Thanks for the inputs, I'll wait another few days before updating this to see if anyone else have an opinion on this.

beacon-node-oapi.yaml
rolfyone: oh! weird i'll check

apis/beacon/blobs/blobs.yaml
rolfyone: just wondering if the ordering is important, so if i query some weird order, do i need to process them in some specific order, and is that possible based on the return data if we're not returning in the same order...
I'm not sure just asking the stupid question to see if its going to cause problems when ppl get blobs back in a different order to the id's they passed...

beacon-node-oapi.yaml
rolfyone: looks right, im not sure how i got to an old commit :/

rolfyone:
So i think this is probably close, just whether namings are correct `blob_sidecars` v. `blobs_sidecars` and being consistent...

also my changes markdown comment about whatever the path ends up as :)

Then I think we should be right to merge...

beacon-node-oapi.yaml
jimmygchen: hmm I think this is on an old commit as well, I just did a global search for `blobs_` and not finding anything. I'll make sure the paths / file name are matching if we end up changing the paths 👍

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,62 @@
...
+            title: GetBlobsResponse
+            type: object
+            properties:
+              data:
+                type: array
+                items:
+                  oneOf:
+                    - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlobSidecar"
+                minItems: 0
+                maxItems: 4
```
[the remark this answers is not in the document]
jimmygchen: Looking at the latest discussions & changes in the PR, it looks like we may not need to change this value here. 

A new `MAX_BLOB_COMMITMENTS_PER_BLOCK` setting is introduced in the PR to increase the limit for the commitment list in a block, and we'll continue to use `MAX_BLOBS_PER_BLOCK` (4) for CL networking and apis - we should probably stick with this constant for the purpose of this endpoint. This value can be updated via a CL hard fork.

apis/beacon/blobs/blobs.yaml
jimmygchen: I'm not sure if the ordering is important tbh, but we do have `index` in each blob sidecar, which seems to be a more reliable property for user to depend on, and consumers should probably be encouraged to use this for ordering instead.

rolfyone:
Hope for one last merge to fix conflicts...

CHANGES.md
jimmygchen: Updated in a6d1c360991676c36cc5691a17d8ac1c17813c33

apis/beacon/blobs/blobs.yaml
jimmygchen: Updated to `blob_sidecars` in a6d1c360991676c36cc5691a17d8ac1c17813c33

rolfyone (approved):
LGTM

g11tech (approved):

mcdee (approved):

rolfyone (approved):
LGTM, thanks team!

PR #288, opened 2023-01-14
Add source param to getLightClientOptimisticUpdate

dapplion:
- Alternative approach to https://github.com/ethereum/beacon-APIs/pull/270

Route `getLightClientOptimisticUpdate` exposes a `LightClientOptimisticUpdate` produced with the SyncAggregate of the head block. This PR allows to customize this behaviour to support consumers that need a more "timely" SyncAggregate.

Proposed values of `source` param:
 - `block` (default): Get SyncAggregate from current's head block body, `body.sync_aggregate`

- `signedcontributionandproof_pool`: Construct SyncAggregate from aggregating `SignedContributionAndProof` objects in the node's pool received from the gossip topic `sync_committee_contribution_and_proof`. MUST aggregate messages with `message.contribution.slot == current_slot` and `message.contribution.beacon_block_root == head_root`.
  
- `synccommitteemessage_pool`: Construct SyncAggregate from aggregating `SyncCommitteeMessage` objects in the node's pool received from the gossip topics `sync_committee_{subnet_id}`. MUST aggregate messages with`slot == current_slot` and `beacon_block_root == head_root`. After receiving a `signatures` request, the node SHOULD subscribe to all sync committee subnets.

Since `source` param is optional and defaults to `block` I guess this is non-breaking change with the current rules of the repo right?

### DIFF
--- a/types/api.yaml
+++ b/types/api.yaml
@@ -81,3 +81,8 @@ Committee:
       description: "List of validator indices assigned to this committee"
       items:
         $ref: './primitive.yaml#/Uint64'
+
+SyncAggregateSource:
+  type: string
+  enum: ["block", "signedcontributionandproof_pool", "synccommitteemessage_pool"]
+  example: block
--- a/apis/beacon/light_client/optimistic_update.yaml
+++ b/apis/beacon/light_client/optimistic_update.yaml
@@ -8,6 +8,24 @@ get:
     Servers SHOULD provide results as defined in [`create_light_client_optimistic_update`](https://github.com/ethereum/consensus-specs/blob/v1.2.0-rc.3/specs/altair/light-client/full-node.md#create_light_client_optimistic_update).
   tags:
     - Beacon
+  parameters:
+    - name: source
+      in: query
+      default: block
+      description: |
+        Source to produce retrieve the SyncAggregate from. Valid values are:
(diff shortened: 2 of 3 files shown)


types/api.yaml
```diff
@@ -81,3 +81,8 @@ Committee:
       description: "List of validator indices assigned to this committee"
       items:
         $ref: './primitive.yaml#/Uint64'
+
+SyncAggregateSource:
+  type: string
+  enum: ["block", "signedcontributionandproof_pool", "synccommitteemessage_pool"]
```
rolfyone, line 87: these are really hard to read...

dapplion: Thanks! Picked an awful name to trigger suggestions :smile:

dapplion:
> Who are the right people to decide on this vs. the alternative you mentioned? I'm not really across the space, and would prefer the right people ultimately make the decision on this one, i'm happy to approve and merge if there's agreement...

Let's sit on it for 2 more weeks to collect input during interop

dapplion:
> I don't like the side effect of GET request to "MUST subscribe to topics and aggregate" which is a heavy operation.
>
> The functionality is already available using the validator namespace, and is more explicit (turn on subscription, then just ask for producing contributions whenever one has interest in it). This makes it explicit to express intent, instead of having nasty side effect to GET requests.

Agree with @etan-status 's points, so closing

PR #289, opened 2023-01-18
Introduce Deneb to remove EIP-4844 references

rolfyone:
Consensus-spec references may need to be fixed once deneb is renamed in that project.

related to https://github.com/ethereum/consensus-specs/issues/3207


mcdee (approved):

jimmygchen:
Hi @rolfyone thanks for this - is this PR ready to be merged?

types/capella/block.yaml
```diff
@@ -73,10 +73,6 @@ Capella:
         properties:
           execution_payload_header:
             $ref: './execution_payload.yaml#/Capella/ExecutionPayloadHeader'
-          bls_to_execution_changes:
```
Inphi, line 76: Why is this removed? Is this related to EIP-4844/deneb?

rolfyone:
> Hi @rolfyone thanks for this - is this PR ready to be merged?

won't be able to merge it until we've done the capella release...

types/capella/block.yaml
rolfyone: this seems like maybe a branch out of date issue ,but does look weird...

rolfyone: ok so this looks like a cleanup, it is in BeaconBlockBodyCommon, so it gets pulled into here when the `allOf` includes `BeaconBlockBodyCommon`

rolfyone:
Trying to get the capella release so this can move. I've asked in API's channel. hopefully this isn't completely stale!

rolfyone:
> any updates on this? Are we still waiting on a Capella release?

this is now unblocked, awaiting approval to merge to dev.

types/deneb/block.yaml
```diff
@@ -0,0 +1,97 @@
...
+  BeaconBlockBody:
+    allOf:
+      - $ref: '#/Deneb/BeaconBlockBodyCommon'
+      - type: object
+        properties:
+          execution_payload:
+            $ref: './execution_payload.yaml#/Deneb/ExecutionPayload'
+
+  BeaconBlock:
+    description: "The [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/Deneb/beacon-chain.md#beaconblock) object from the CL Deneb spec."
```
mcdee, line 55: URL should be lower-case D in Deneb for it to resolve.

types/deneb/state.yaml
```diff
@@ -0,0 +1,88 @@
...
+        type: array
+        description: "Fixed length of 8192 items"
+        items:
+          $ref: '../primitive.yaml#/Root'
+      state_roots:
+        type: array
+        description: "Fixed length of 8192 items"
+        items:
+          $ref: '../primitive.yaml#/Root'
+      historical_roots:
```
mcdee, line 26: Is this now a 0-length array in the spec, given that it has been superseded by `historical_summaries`?  Also, we're missing `historical_summaries` in this definition.

mcdee (changes requested):

types/deneb/state.yaml
rolfyone: It looks like we're missing historical_summaries in the 2.4.0 release from capella, so that's a gap I might raise an issue on...  This is consistent with the Capella definition in master, so we need to fix capella and deneb both.

rolfyone: https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md#beaconstate
I'm not sure it's a 0 length, but whatever it ends up, it does look like tech debt...

rolfyone: ok, was easy enough for capella, brought in change and moved to deneb also.

mcdee (approved):

michaelsproul (approved):

PR #290, opened 2023-01-20
Handle EL offline/unavailable in the Sync API

mehdi-aouadi:
When the execution client if offline/unavailable, it can't be properly reported by the sync API response.
Added an `el_offline` field to the `GetSyncingStatusResponse`.


apis/node/syncing.yaml
```diff
@@ -33,6 +36,7 @@
                       allOf:
                         - type: boolean
                         - description: "Set to true if the node is optimistically tracking head."
-
       "500":
         $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
+      "503":
+        $ref: '../../beacon-node-oapi.yaml#/components/responses/CurrentlySyncing'
```
rolfyone, line 42: Do we want a different response template here? It seems like a very explicit thing we're defining as a 503...

mehdi-aouadi: We could keep the 500 response only if that's what you mean

rolfyone: I just mean you're using the 'currently syncing' message, have a look at the result on the http server... it'll list 'the server is currently syncing', do we need to list 'the EL server is offline' or something for a better context?

mehdi-aouadi: Updated it with: The execution layer is unavailable

rolfyone:
There's a fairly long discussion in APIs channel on discord re. this ticket. I believe the preference would be to not return a 503, but instead have a `el_offline` or similar attribute that indicates the el is not responding... it does sound like a good suggestion from the outside, so posting here - thanks @mcdee for the suggestion.

rolfyone (dismissed):
LGTM, just PR comment now.

rolfyone:
oh - can we add an entry in changes.md to indicate that the endpoint has been updated? Cheers!

CHANGES.md
```diff
@@ -7,31 +7,32 @@ These endpoints have been added or updated since the last release.
...
+|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------|---------------------------------------------------|----------------------------------------------------|-------------------------------------------------|-------------------------------------------|
+| [#290](https://github.com/ethereum/beacon-APIs/pull/290) add `el_offline` to `GetSyncingStatusResponse`                                                              |                                                  |                                                   |                                                    |                                                 |                                           |
```
rolfyone, line 12: probably more meaningful to put 
`/eth/v1/node/syncing` added `el_offline` to response

mehdi-aouadi: Done

rolfyone (approved):
LGTM

arnetheduck:
is there a rationale for this field? ie what actually useful information does it bring, on top of `is_optimistic` (which already requires an "online" el)?

michaelsproul:
@arnetheduck The rationale for us in Lighthouse is:

- We don't want to import blocks optimistically when the EL errors, because optimistic import necessarily disables attestation and block production. Our concern is that a "bomb" block which makes the EL timeout or error across the whole network could pose a liveness risk if it throws all the CL clients into optimistic sync.
- Instead, we want to give a weak indication to any connected VC that "hey maybe you shouldn't use this node if you've got other options". If the EL is not responding, or has errored/timed out on the latest `newPayload` call, then we mark it as `el_offline: true`.
- The BN stays live (keeps producing attestations & blocks) while `el_offline: true`.

Credit to @paulhauner for the bomb block argument.

paulhauner:
I appreciate that Nimbus has decided that they feel it's safe to go ahead and optimistically import against the [optimistic sync spec](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#execution-engine-errors) (I mean that genuinely, since the spec may be over-restrictive now the merge transition has passed). However I don't think that there's been enough evidence presented about the liveness-safety of that approach to say that *everyone* should be optimistically importing on errors and that the API should only recognize that approach.

Deciding to optimistically import blocks with an offline/erroring EE is a non-trivial decision which, in my opinion, is neither objectively good or objectively bad. I see that giving CL clients the ability to differentiate between optimistic and offline allows for greater client diversity.

rolfyone:
> I appreciate that Nimbus has decided that they feel it's safe to go ahead and optimistically import against the [optimistic sync spec](https://github.com/ethereum/consensus-specs/blob/dev/sync/optimistic.md#execution-engine-errors) (I mean that genuinely, since the spec may be over-restrictive now the merge transition has passed)

Is this something we should consider raising an issue on?

michaelsproul:
> Is this something we should consider raising an issue on?

IMO it's ok if Nimbus BN just always returns `ef_offline: false`. It's perfectly compatible

PR #296, opened 2023-02-03
Add optional basic/bearer authentication

skylenet:
This will enable the "Authorize" button on Swagger UI. Note that authentication **is optional** and doesn't have to be set. 
But this is useful if you want to use the UI to interact with endpoints that are protected with Basic Auth or Bearer tokens. 

Example of how the change looks like: https://skylenet.github.io/beacon-APIs/?urls.primaryName=dev 

Just to clarify the following: 

```yaml
security:
  - {}  
  - BasicAuth: []
  - BearerAuth: []
```

`{}` means that no auth can be also used.

### DIFF
--- a/beacon-node-oapi.yaml
+++ b/beacon-node-oapi.yaml
@@ -195,8 +195,19 @@ paths:
   /eth/v1/events:
     $ref: "./apis/eventstream/index.yaml"
 
+security:
+  - {}
+  - BasicAuth: []
+  - BearerAuth: []
 
 components:
+  securitySchemes:
+    BasicAuth:
+      type: http
+      scheme: basic
+    BearerAuth:
+      type: http
+      scheme: bearer
   schemas:
     BeaconState:
       $ref: './types/state.yaml#/BeaconState'


ralexstokes (approved):

rolfyone:
The main question raised on this one was around what is the value it brings to including in the specification - can you elaborate on why it should be added?

My only thought is if you're using auth to connect from our spec page to your beacon-api directly which is using auth?

skylenet:
> The main question raised on this one was around what is the value it brings to including in the specification - can you elaborate on why it should be added?
> 
> My only thought is if you're using auth to connect from our spec page to your beacon-api directly which is using auth?

Yes, that's the main reason for adding it. I have a bunch of public endpoints that are protected via basic auth. Adding this optional authentication scheme would make it possible to use the deployed Swagger UI to interact with these endpoints.

rolfyone:
I guess the thing to be aware of, if we were to do this and you use the deployed API with basic auth to a HTTP endpoint, the implications...

I'm not assuming knowledge, so just to break it down...

When HTTP Auth is set, it's in a HTTP header. These headers are sent in clear text, and any proxy , routing service that handles the request to get it to your endpoint is going to have access to all the details, including basically your password. This would mean they can literally copy your http authentication header, and push requests to your servers with the credentials you've supplied.

In the validators-api, we required HTTPS because of this, as it's basically a security risk.  Our concern adding this to the swagger endpoint here is that people won't understand this security risk, and indadvertedly compromise their security by using our supplied site...

I think this is a fair representation of the discussion on discord, but will tag @mcdee and @sproul to see if they've anything to add.

skylenet:
Just adding some more context to my case: I'm terminating TLS on my servers (via Nginx) and proxy to a local beacon node. But I do agree with you that some users might not be aware of how HTTP and transport encryption works. 

On the other side: The current deployed Swagger UI at https://ethereum.github.io/beacon-APIs allows a user to configure a custom endpoint as the `server_url` but browsers won't allow you to do XHR requests to non HTTPS secured websites if your website is using HTTPS. You'll get an error like `Mixed Content: The page at 'https://ethereum.github.io/beacon-APIs/#/Beacon/getGenesis' was loaded over HTTPS, but requested an insecure resource 'http://your-website.org/eth/v1/beacon/genesis'. This request has been blocked; the content must be served over HTTPS.` So that's actually quite good. The browsers are already protecting the users from doing something stupid by accident.

So this means that the only way to use the current deployed website is to provide a HTTPS protected `server_url`. And if someone does this, I would argue that having some basic auth on it is better then having no auth, which would mean that the beacon API would be publicly reachable for everyone. (Unless you're IP whitelisting or something like that). 

But... if we're really concerned about users using the website to try interact with any beacon API, then we should maybe totally disable the "try out" feature on SwaggerUI or use  [Redoc](https://github.com/Redocly/redoc) which doesn't have that and fits probably better the "read only" mode for the spec.

rolfyone:
Good argument. I don't mind adding this to help your use-case, will just flag to @mcdee and @michaelsproul and take a vote

michaelsproul:
I'm still loosely against. Maintaining the beacon-APIs is hard enough, and expecting the spec to also function as an interactive client seems API like scope creep that will increase maintenance burden.

mcdee:
I agree with @michaelsproul let's keep the spec focused on being the spec.

rolfyone:
@skylenet it sounds like no real agreement on this one, so I can't really merge it. Did you want me to leave it open for now?

skylenet:
I agree that auth is not in scope of the spec. Closing the PR 🙏

PR #297, opened 2023-02-05
Remove array from attestation rewards response

michaelsproul:
There was an extra layer of unnecessary array nesting in the attestation rewards response:

```json
{
  "data": [
    {"ideal_rewards": [], "total_rewards": []}
  ]
}
```

The `data` list should always have 1 element, because the real lists are contained in `ideal_rewards` and `total_rewards`. Therefore this structure is more appropriate:

```json
{
  "data": {"ideal_rewards": [], "total_rewards": []}
}
```

This is already implemented by Lighthouse in https://github.com/sigp/lighthouse/pull/3822.


rolfyone (approved):
LGTM

Issue #300, opened 2023-02-08
Blob Signing APIs


realbigsean:
I started working on making a spec for this and had some general thoughts and questions:

### Blinded blobs by default

I would propose we only support blinded blob communication in block production because blobs aren't interpret-able by the VC and because they are large. It would also save us work with the blinded block flow for the builder API. Full blobs are still accessible [here](https://github.com/ethereum/beacon-APIs/pull/286) for purposes outside block production.

#### Separate requests 

- `GET /eth/v1/validator/blobs_sidecar/{slot}` with no query params
  - We could maybe still include necessary information to trigger block production (randao, graffiti) in case these requests come out of order but this seems ugly
- Having separate requests leaves open the possibility of inconsistency between block and blobs, so I think we'd have to do some sort of consistency check
-  Would maybe need to allow optionally requesting blobs by block root in the event of a consistency issue
- For maximal simplicity we could request the block as normal and follow up with a blobs request by root, but this would be slower

#### Combined request
- Have a single endpoint for block+blob production that returns something like this:
```
BlockContents {
    block: BeaconBlock,
    blobs: BlindedBlobsSidecar,
}
```
- Can we use the existing block endpoint or would this require a new endpoint?
- If it requires a new endpoint we may have to worry about switching to it at the fork, or making it backwards compatible

#### Separate requests

- A new endpoint that accepts a single signed blob, a blinded version of the singular blob gossip struct - this lets the BN start publishing to gossip most quickly
  - `POST /eth/v1/validator/blinded_blob` 
 
#### Combined request
- have a single block + blobs endpoint accepting some sort of a `SignedBlockContents` struct
- I think the main benefit here is it would mirror the combined `GET` endpoint if we go that route, but I don't see the benefit this adds otherwise.

mcdee:
From my current (vague) understanding of the blobs system there is some blob metadata embedded in the blocks, so the block and associated blobs are tightly coupled.  If that's the case then having the combined endpoint for both obtaining data to sign seems reasonable.  The `GET` endpoint could be backwards-compatible by making the blobs optional in the return data structure.

As the items are decoupled, or going to be decoupled, on the p2p layer it makes sense to keep the endpoints for them separate *unless* the beacon node is going to carry out its own verification of both items in tandem before broadcasting (_i.e._ if pre-gossip verification of the blob requires information in the block), in which case there wouldn't be any benefit to having them as separate endpoints.

realbigsean:
[...]

PR #302, opened 2023-02-13
Add blob signing endpoints

realbigsean:
Resolves https://github.com/ethereum/beacon-APIs/issues/300

### Notes 

- Built on top of https://github.com/ethereum/beacon-APIs/pull/289 so that should probably be merged first
- Moved `KZGProof` and `Blob` to primitives
- Opted for a combined `GET` request for block and blobs and a separate `POST` request for each signed blob

### Questions

- 202 vs 200 response code doesn’t really work the same anymore with these separate `POST` requests. I'm not sure how we should handle this. We could keep the existing meaning for each, it will just mean *most* requests return a 202 and only the final block/blob message for a slot leads to import and returns a 200. 
- There’s no spec definition of the `BlockContents` struct for SSZ serialization. This struct is solely for communication here and not used in consensus elsewhere. Should the SSZ definition be included here? Should we add some similar type's definition to the consensus spec? Can we describe this SSZ type in terms of spec objects? (e.g., `SignedBeaconBlock` with `BlobSidecars` appended?)
- Generally is it OK for the existing `/validator/block` endpoints to return a new type only for deneb (`BlockContents`) ?


types/deneb/block_contents.yaml
```diff
@@ -0,0 +1,18 @@
+Deneb:
+  BlockContents:
+    type: object
+    description: "The required object for block production according to the Deneb CL spec."
+    properties:
+      block:
+        $ref: "./block.yaml#/Deneb/BeaconBlock"
+      blob_sidecars:
```
jimmygchen, line 8: I wonder if it's worth being more explicit here to indicate this is `blinded`?

jimmygchen:
Probably need to add the new endpoint to the top level `CHANGES.md` as well.

jimmygchen:
The top level `validator-flow.md` has some explanation on the block proposing flow, maybe worth adding blob signing to the [`Block Proposing` section](https://github.com/ethereum/beacon-APIs/blob/master/validator-flow.md#block-proposing)?

beacon-node-oapi.yaml
```diff
@@ -89,6 +89,8 @@ paths:
     $ref: "./apis/beacon/blocks/blinded_blocks.yaml"
   /eth/v1/beacon/blocks:
     $ref: "./apis/beacon/blocks/blocks.yaml"
+  /eth/v1/beacon/blob_sidecar:
```
jimmygchen, line 92: do you think we should make this `/eth/v1/beacon/blinded_blob_sidecar`?

arnetheduck:
https://github.com/ethereum/consensus-specs/pull/3244 is now ready for review - it would be nice to leave this PR open for a bit until that has been merged and we've had time to evaluate the impact of blinded-only blobs.

apis/beacon/blobs/post_blobs.yaml
```diff
@@ -0,0 +1,54 @@
...
+  tags:
+    - Beacon
+    - ValidatorRequiredApi
+  summary: "Publish a signed blob."
+  operationId: "publishBlob"
+  description: |
+    Instructs the beacon node to broadcast a newly signed blob to the beacon network,
+    to be included in the beacon chain. A success response (20x) indicates that the blob
+    passed gossip validation and was successfully broadcast onto the network.
+    The beacon node is also expected to integrate the blob into state once it receives the
```
tbenr, line 11: The description here needs to be changed. Blobs are not expected to be integrated into state.

apis/beacon/blobs/post_blobs.yaml
```diff
@@ -0,0 +1,54 @@
...
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized blob sidecar bytes. Use content type header to indicate that SSZ data is contained in the request body."
+  responses:
+    "200":
+      description: "The blob was validated successfully and has been broadcast. It has also been integrated into the beacon node's database."
+    "202":
+      description: "The blob could not be integrated into the beacon node's database as it failed validation, but was successfully broadcast."
+    "400":
+      description: "The `SignedBlobSidecar` object is invalid and could not be broadcast"
```
tbenr, line 40: Shouldn't be `SignedBlindedBlobSidecar` instead of `SignedBlobSidecar`?

tbenr:
A curiosity. What's the rationale behind having separate `SignedBlindedBlobSidecar` instead of a single post with an array of `SignedBlindedBlobSidecar`?

I can image a more streamlined flow in VC which is (remote)sign and publish for each Sidecar. But I can still see benefits of having aggregated methods for remote-signing and publishing: it better scale with increasing number of blobs.

jimmygchen:
> A curiosity. What's the rationale behind having separate `SignedBlindedBlobSidecar` instead of a single post with an array of `SignedBlindedBlobSidecar`?
> 
> I can image a more streamlined flow in VC which is (remote)sign and publish for each Sidecar. But I can still see benefits of having aggregated methods for remote-signing and publishing: it better scale with increasing number of blobs.

I'm guessing it's for the same reason as separating blobs from the block request - so that it can be sent to the BN and lets it starts publishing to gossip more quickly. Although I wonder if it still provide the same benefit if we're sending blinded blobs between VC and BN. Does it still make sense to even separate block and blobs if we go with blinded blobs?

tbenr:
> Although I wonder if it still provide the same benefit if we're sending blinded blobs between VC and BN

On the proposer side, blobs are available immediately, all together. When VC post them all (blindly signed) to BN, bn can gossip them all together in parallel. The only problem i see is that the return code for the publish will (maybe) take in consideration partial failures. So we exchange a single call with more complex error handling\retry from the VC side.

Same consideration applies for remote signing (web3signer) all blinded blobs with one call.

realbigsean:
> > A curiosity. What's the rationale behind having separate `SignedBlindedBlobSidecar` instead of a single post with an array of `SignedBlindedBlobSidecar`?
> > I can image a more streamlined flow in VC which is (remote)sign and publish for each Sidecar. But I can still see benefits of having aggregated methods for remote-signing and publishing: it better scale with increasing number of blobs.
> 
> I'm guessing it's for the same reason as separating blobs from the block request - so that it can be sent to the BN and lets it starts publishing to gossip more quickly. Although I wonder if it still provide the same benefit if we're sending blinded blobs between VC and BN. Does it still make sense to even separate block and blobs if we go with blinded blobs?

Yeah, I honestly don't have a sense for whether a batch or individual requests would be faster and at what number of blobs. It made intuitive sense to me to view this endpoint as a "gossip publisher" in which case having it mirror the gossip topic's contents made sense. Open to changing this.

As far as how blinding impacts this, we can blind and cache in the beacon node per-blob, so I'm not sure I see how that impacts whether to publish them separately. Would keeping them together in blinding be a reduction in complexity or faster?

realbigsean:
> On the proposer side, blobs are available immediately, all together. When VC post them all (blindly signed) to BN, bn can gossip them all together in parallel. The only problem i see is that the return code for the publish will (maybe) take in consideration partial failures. So we exchange a single call with more complex error handling\retry from the VC side.

I agree the error handling gets more complex, the 200 vs 202 response codes end up having different meanings here

tbenr:
Another consideration is the "stickiness" required by blinded flows. I mean, when a VC can use multiple BNs, the publish calls must end on the same BN that handled the block/blobs production. Different BNs wont be able to unblind them. (I don't know if any clients has implemented a shared cache between different BN "clusters", teku didn't). So having multiple calls required to complete the entire flow become more exposed to eventual failover in the middle of the flow.

Maybe @mcdee might have opinions on these topics.

jimmygchen:
> Different BNs wont be able to unblind them. (I don't know if any clients has implemented a shared cache between different BN "clusters", teku didn't). > 

I think that's definitely a very strong point against separating the calls if a shared cache is required between BNs.

Although I'm not fully grasping why different BNs can't unblind them?
Let's say the VC submitted a signed block, and signed blinded sidecars all to different BNs, and each BN then sends them off to the relays individually, and then each receives the corresponding unblinded block / unblinded blobs back, wouldn't they be able to publish them separately? Am I missing something?

jimmygchen:
I think I confused this with the builder flow, you're right in the normal flow we'll need the blob cache (full blobs) to unblind them.

jimmygchen:
If we need a shared cache that might get too complicated, so the separate request + blinded blob signing combination may not be ideal.

that leaves us to 
1. separate request + full blob signing 
2. combined request + blind blob signing
3. combined request + full blob signing

not sure if there's any good reasons for full blob though, what do you think?

tbenr:
I thought a bit more about this and my current summary is this (let's keep the builder flow out for now)

- **blinded-full**: the main difference here is that the flow become stateless if everything is full. But as soon as we have one blinded object, than the flow must keep the talking with the same BN. So, to me, mixing a full block with blinded blobs doesn't make much sense. When we have full content (block and blobs) the VC can decide to publish each object to whatever BN it has configured (see sentry nodes feature in teku).

In case of blinded content, we can have everything blinded and VC have to stick to the BN that handled the content creation (teku has this stickiness logic implemented for the blinded block flow) but with the benefit of exchanging much less data.

- **separated-combined**: It is just matter of decide if worth to have the publish api accepting an array of blob sidecars and handle the complexity of partial errors or go for multiple single calls. In case of combined full blobs, the payload will be big BN will start publishing only when all of them has been streamed. TBH I don't know if having, let's say, 4 parallel tcp streams can give some latency benefits here. At the end I don't have a strong opinion.

tbenr:
The problem with the builder flow is the fallback. If the proposing BN decided to fallback to local EL, than VC have to publish to that one only. A solution to that would be to add some metadata in the block(content) production to signal the VC that a fallback has occurred, so it activates a sticky logic. In teku implementation, whenever a BN receives a publish request that requires an unblinding process it prints a warning and tries to unblind via builder anyway. Relay are able to handle multiple calls for the same block. But you'll receive an error if fallback occurred.

Pelonchas57:
> Resolves https://github.com/ethereum/beacon-APIs/issues/300
> 
> ### Notes 
> 
> - Built on top of https://github.com/ethereum/beacon-APIs/pull/289 so that should probably be merged first
> - Moved `KZGProof` and `Blob` to primitives
> - Opted for a combined `GET` request for block and blobs and a separate `POST` request for each signed blob
> 
> ### Questions
> 
> - 202 vs 200 response code doesn’t really work the same anymore with these separate `POST` requests. I'm not sure how we should handle this. We could keep the existing meaning for each, it will just mean *most* requests return a 202 and only the final block/blob message for a slot leads to import and returns a 200. 
> - There’s no spec definition of the `BlockContents` struct for SSZ serialization. This struct is solely for communication here and not used in consensus elsewhere. Should the SSZ definition be included here? Should we add some similar type's definition to the consensus spec? Can we describe this SSZ type in terms of spec objects? (e.g., `SignedBeaconBlock` with `BlobSidecars` appended?)
> - Generally is it OK for the existing `/validator/block` endpoints to return a new type only for deneb (`BlockContents`) ?

apis/beacon/blobs/post_blobs.yaml
g11tech: i guess description is talking about constructed object

jimmygchen:
Hey @tbenr,

Thanks for the explanation!

Just to make sure I understand you correctly (re builder flow) - in the case of fallback, the proposing BN would use the local EL to create a block and return it to VC, and VC would then sign and submit the full block to (any) BN. I don't quite get why it needs to be sticky if the flow became stateless?

With the builder flow, if we go with combined request to a BN, then it would be stateless as well - given the BN will receive the full blobs from the relayer. This would reduce the complexity and would be closer to today's flow? 

```mermaid
sequenceDiagram
    participant VC
    participant CL
    participant EL
    participant relay
    VC->>CL: get blinded block
    CL->>relay: get header
    alt successful getHeader response
        relay-->>CL: BuilderBid response containing blinded block and blinded BlobSidecars
        CL-->>VC: blinded block and blinded BlobSidecars
        Note over VC: sign block and each blinded BlobSidecars individually
        VC->>CL: submit signed blinded block and blobs (1 call)
        CL->>relay: submit signed blinded block and blobs (1 call)
        Note over relay: construct block and broadcast block and blobs
        relay-->>CL: revealed payload and full blob sidecars
    else request to relay failed / timed out, fallback to local EL
        CL->>EL: get full block (and full blobs)
        CL-->>VC: full block (and full blobs)
        Note over VC: sign block and each BlobSidecars individually
        VC->>CL: submit signed block and blobs (1 call)
    end
    Note over CL: construct block and broadcast block and blobs
```

tbenr:
@jimmygchen problem is that vc initiate the flow with a get blinded block and bn may respond with a blinded or full content. Even if it might be ok for json content type, for SSZ content we will have problems wince you need to know in advance what you are trying to decode. We currently use the an header (iirc) to distinguish between milestones but this won't be enough. We could solve it by using a Union or introducing another header.

realbigsean:
We spoke about this on the 4844 call today and it sounded like blinded by default would be preferable for the sake of reducing the amount of data transferred between the beacon node and validator client. Even though this leads to this new requirement that a block has to be published through the beacon node that constructed the block (this requirement already exists for the current blinded block flow). Wondering about how this impacts things like Vouch? @mcdee

We discussed a combined POST request for publishing signed blobs and the associated block together. Didn't seem like people thought having individual requests for each blob will really gain us anything performance-wise. Especially with blinded-by-default blobs which will result in a smaller message size and will mean the VC doesn't have to calculate each blob tree hash root to sign over. Individual requests add some complexity around when exactly a beacon node should import a locally produced block. If requests are separate, a node must wait until all components of the block are received, returning a 202 status for each message until the final component is received and it can import and return a 200 status. So on net sounded like the simplicity of the combined POST we preferred.

TL;DR I'm going to update this PR to a **blinded** and **combined** design

arnetheduck:
> Even though this leads to this new requirement that a block

This means functionality like [sentry nodes](https://nimbus.guide/validator-client-options.html#sentry-node-setup) will be constrained compared to the status quo - it also makes redundant setups / multiplexing several beacon nodes potentially of different client types harder - similar to vouch, the Nimbus validator client allows using each beacon node for specific tasks - one such setup is to use one set of nodes for production and another for publishing.

I see value in specifying both, like we have today - the non-blinded version semantically is a lot more simple because the beacon node doesn't have to store the "unblinded" blobs in some intermediate location for some undeterminate time, which means the implementation burden is low.

This is also an issue for the current blinded request: it does not clarify whether it is valid to ask for a blinded block and expect that EL-produced non-blinded blocks are made available through it (vs only giving builder blocks) - this has led to differences in implementation where some clients return blinded versions of unblinded blocks while others don't.

g11tech:
> In short: blinding the blob data would reduce the reliability of the network, unless the blinded data was available from an independent source. Even then, it would be preferable to provide unblinded information wherever possible to keep state within the proposing entity (_i.e._ the validator client).

As per my comprehension what you suggesting is:
1. Have blinded blob design between beacon and validator AND
2. Have an independent service for storing/retrieving/publishing blinded blobs by beacon 

I think this could work, even in the current implementation beacon is doing exactly 2 but could be nice to have it as external service given the challenge of home stakers with the increasing number/size of blobs

g11tech (commented):
looks good to me :+1 will modify lodestar impl to this and comeback to approve from my side

arnetheduck:
>  sake of reducing the amount of data transferred

I'm a bit curious about this point, and what we envision the problems are: we're intending to ship each block and blob redundantly around 8 times per node around the globe - comparatively, moving them between vc and bn shouldn't really matter that much.

+1 to what @mcdee points out about architectural simplicity, ie blinding the blobs requires keeping more state in the beacon node which indeed complicates the core implementation where no mev-boost is involved - the resilience argument here is pretty convincing, ie the non-blinded request is stateless and at the same time our insurance against complex block building failures.

realbigsean:
I think I'm convinced that there's values in having both the fully blinded and fully unblinded flow. So I'll update the PR here.  Are you on board with that @g11tech @tbenr ?

mcdee:
> As per my comprehension what you suggesting is:

>  1.  Have blinded blob design between beacon and validator AND
>   1. Have an independent service for storing/retrieving/publishing blinded blobs by beacon

For 2) it would be a service that provided unblinded blobs and would be independent of the beacon node.  This design would require the beacon node to fetch blobs from the blob service, which would make it possible for that service to be one of many things.  In the case of a validating infrastructure using Vouch, Vouch would fetch the blobs from the actual blob service and the beacon nodes would use Vouch as the blob service.  This would remove the state imposition on beacon nodes and re-enable the ability for a single validator client to use multiple beacon nodes for whatever reason (resilience, selection of best option, use of sentry node _etc._)

mcdee:
> I think I'm convinced that there's values in having both the fully blinded and fully unblinded flow.

This only really works for Vouch if the beacon node outsources the blob storage and fetches it over an API.  We have to take blinded execution payloads as a requirement, but a fully blinded flow means state (blob information) in a single beacon node.

mcdee:
> I find the "separate blob store" idea somewhat more difficult to work with - the VC-BN interaction is already more complex than it would have to be (compared to say .. web3signer which offers much of the functionality in a much simpler setup) - one more data source seems like a lot for the average user to have to manage, relative to potential benefits for complex setups.

Not sure what the "average validator" looks like.  A lot of at-home validators are already pushing the limits when it comes to the bandwidth they are using.  If they need to deal with blob traffic all of the time we're going to lose some of them.  Larger providers today commonly run multiple beacon nodes for resiliency or the ability to select the best data from each one, and beacon nodes subscribed to all subnets are already pushing a combined ~50Mb/s with ~100 peers; increasing this number is likely to result in providers reducing the number of beacon nodes they are running, which wouldn't be great for client diversity or general network health.

As for complexity of setups: a beacon node wouldn't be required to use an external blob service, but it should be possible for it to do so.  Similar to MEV relays today, a beacon node could have a single additional parameter `external-blob-server-url`.  If the parameter was set then the beacon node would not participate in the p2p blob channels, and instead would fetch blobs from the supplied URL when a block (blinded or otherwise) was requested by a validator client.  So by default it wouldn't change what exists today, but would give the ability to handle situations where home stakers don't have the bandwidth to handle the additional blob traffic, and service providers want the resilience that comes from being able to obtain the blob data independently from an individual beacon node.

realbigsean:
> > I think I'm convinced that there's values in having both the fully blinded and fully unblinded flow.
> 
> This only really works for Vouch if the beacon node outsources the blob storage and fetches it over an API. We have to take blinded execution payloads as a requirement, but a fully blinded flow means state (blob information) in a single beacon node.

In the fully blinded flow, the party responsible for payload production is also responsible for selecting the blobs to include in the block so this party will have the blob state. I'd imagine the reasons leading us to blind transactions generally in the builder flow would apply to blob data?

> As for complexity of setups: a beacon node wouldn't be required to use an external blob service, but it should be possible for it to do so. Similar to MEV relays today, a beacon node could have a single additional parameter external-blob-server-url. If the parameter was set then the beacon node would not participate in the p2p blob channels, and instead would fetch blobs from the supplied URL when a block (blinded or otherwise) was requested by a validator client. So by default it wouldn't change what exists today, but would give the ability to handle situations where home stakers don't have the bandwidth to handle the additional blob traffic, and service providers want the resilience that comes from being able to obtain the blob data independently from an individual beacon node.

You'll still need to download all blobs in order to verify the chain and in turn attest and choose which head to propose on top of. Otherwise we're in the territory of lazy validating. I'm not sure it's great outsourcing the data availability check generally it's sort of like staking with your VC pointed at eth2 infura.

mcdee:
> In the fully blinded flow, the party responsible for payload production is also responsible for selecting the blobs to include in the block so this party will have the blob state. I'd imagine the reasons leading us to blind transactions generally in the builder flow would apply to blob data?

Understood, thanks.

> You'll still need to download all blobs in order to verify the chain and in turn attest and choose which head to propose on top of. Otherwise we're in the territory of lazy validating. I'm not sure it's great outsourcing the data availability check generally it's sort of like staking with your VC pointed at eth2 infura.

This would only require the blobs that have been included in blocks, rather than all of the potential blobs floating around on p2p.  It also wouldn't require the beacon node to have to forward these with p2p gossip (assuming that the p2p for blobs is separate from that for blocks).

Basically, what I'm saying is that at current we have two options:
1. have enough bandwidth: you can validate
2. don't have enough bandwidth: you can't validate

And we're increasing the bandwidth requirement by a fair chunk so potentially pushing some people from category 1 to category 2.  By allowing an external service to supply blobs you can create a third option:
1. have lots of bandwidth: you can validate and participate in p2p blob traffic
2. have enough bandwidth: you can validate and use an external service for blobs
3. don't have enough bandwidth: you can't validate

I understand that there could be a concern of people who are in category 1 going for category 2, but that could be discouraged by a) not making it a default and b) highlighting the reliance on an external service as a result.

g11tech:
> I think I'm convinced that there's values in having both the fully blinded and fully unblinded flow. So I'll update the PR here. Are you on board with that @g11tech @tbenr ?

yes looks good to me :+1: 

Would be great if we can compel (by way of spec/api standatords) beacons  to also provide execution blocks+blocks via blinded route, so a validator can be freely configured by a way of flag to use blinded flows as much as possible, applicable specially in retail scenarios as well as with an external blob service as @mcdee has suggested (which imho is a good idea for keeping homestakers in fold)

tbenr:
I generally like the idea of having a separated "role" responsible for handling the blobs, with a clean API layer.

I actually see that a stand alone BS can't exist, it is more a "BN with blob capability" (to correctly participate in gossip and serve RPCs you need to follow the chain) which also expose blobs API.
On the other side we will have a "BN light" configured with `external-blob-server-url`.

The problem I see is related to p2p responsibility of BN. Spec now assumes that a BN have to participate in gossip and serve RPC query for blobs. And peers will downscore a node that is detected to not participate. If those become optional what would stop "category 1" to configure them as "category 2" to save bandwidth?
This means that we will have "attnets", "syncnets" and "blobnets" in enr. A node declaring "blobnets" is supposed to serve blobs RPCs otherwise not. (Or have to serve RPCs and act as a proxy (download\upload)?)

I need to think if there are additional complexity for clients to implement both "BN light" and "BN with blob capability" modes. Sync logic should be able to seamless use RPC methods or blobs API. This may work nicely.

realbigsean:
> This would only require the blobs that have been included in blocks, rather than all of the potential blobs floating around on p2p. It also wouldn't require the beacon node to have to forward these with p2p gossip (assuming that the p2p for blobs is separate from that for blocks).

Just to clarify the blobs in beacon p2p are there to communicate the validity of the candidate head, so they are necessary to attest. The blobs that are relevant to block production (blobs that are candidates for inclusion) are communicated in the execution layer's mempool. So whether the beacon is listening on gossip isn't relevant to these APIs. 

Since blobs are tightly coupled with blob transactions, whatever the source of the transactions is, more or less has to be the source of the blobs. Since the source of the transactions is currently established as the BN in the relevant APIs, I don't actually think it makes sense to separate this particular API. Unless we want to more broadly re-architect this to allow the VC to talk to something that looks like an EL (can provide a valid set of transactions and blobs).

So I think the discussion about an external blob service here might actually be out of scope.. I've opened [this issue](https://github.com/ethereum/consensus-specs/issues/3286) for discussion on whether we should support a class of beacon nodes that don't listen on gossip and can use `--external-blob-server-url`

realbigsean:
> Would be great if we can compel (by way of spec/api method/flag) beacons to also provide execution blocks+blobs via blinded route, so a validator can be freely configured ( for e.g. by a way of flag) to use blinded flows as much as possible :

I want to make sure, is this suggesting the same thing that was mentioned on the call? Have a single set of endpoints that could support both the blinded and unblinded flow? 

Specifically I think we're trying to address this scenario:

1. the beacon node is requested to make a blinded proposal
2. the beacon node can't reach the relay, so has to fall back to the local execution layer
3. the beacon node now has a full payload and blob and can return it to the VC but whether it *should* is ambiguous. 
    - In lighthouse we just cache the full payload and blobs in the BN - but this adds statefulness where it isn't strictly necessary and this can cause a failed proposal in combination with fallback between beacon nodes

But additionally if we are able to support both flows on a single set of endpoints we can deprecate the purely blinded endpoints. This also means we can remove any validator client config for "blindedness preference" it can just be inferred based on how the beacon node is configured.

To enable this we could:
1. add a new header similar to the `Eth-Consensus-Version` to denote blinded vs unblinded
2. add `{fork}-blinded ` variants to the existing `Eth-Consensus-Version` header

We would need to also explicitly allow the beacon node to respond with either the blinded or unblinded block type so long as the header is added. 

Open questions:
- Should this be added in new versions of the block endpoints (increment V2->V3 or V1->V2)? This + the fact that we are move from strictly blocks to a new block content type might be reason enough. I don't really have a strong opinion. Or should this be a new endpoint all together?
- If we go this route do we still support Deneb types on existing endpoints (`BlockContents` instead of blocks), both blinded and unblinded ?

g11tech:
> I want to make sure, is this suggesting the same thing that was mentioned on the call? Have a single set of endpoints that could support both the blinded and unblinded flow?

yes, but i am not sure about merging the blinded and unblinded endpoints and would prefer less "magic" on bn as it becomes hard to reason for what really happened.
Its not possible to get builder block contents on unblinded route.

so my suggestion was to for e.g. add a `query` param to produceBlindedBlock indicating its asking for execution/builder block.
similarly on publish block, again a query param to indicate what source was.

> 
> Specifically I think we're trying to address this scenario:
> 
>     1. the beacon node is requested to make a blinded proposal
> 
yes using execution source
>     2. the beacon node can't reach the relay, so has to fall back to the local execution layer
> 

In validator lodestar  race two calls on builder/execution and picks one which resolves first after a cutoff (3sec) or the more paying block if both resolves before the cutoff. Again i wouldn't prefer for beacon to make fallback decisions again for the purposes of validator logic enforcing any timings/cutoff constraints.

>     3. the beacon node now has a full payload and blob and can return it to the VC but whether it _should_ is ambiguous.

Again VC by the way of query flag for e.g. `block_source(builder|execution)` on the blinded route removes any ambiguity for the beacon.

>        
>        * In lighthouse we just cache the full payload and blobs in the BN - but this adds statefulness where it isn't strictly necessary and this can cause a failed proposal in combination with fallback between beacon nodes
> 
yes lodestar also implements similarly, but those who want to support fallback could request full contents on the normal route for the execution block use full blockcontents publish route. For builder block which is blinded route, it can use any BN to publish if connected to the same relay/builder. 
So at all times validator knows what action to take for fallback since it explicitly decides what to do. Validator SHOULD assume that using blinded route for an execution block will make the followup publishing sticky to that beacon. For 90% of the retail scenarios this flow is what will generally adopted as the probability of publishing failure after successfull block production is quite low. (i think)
 
> 
> But additionally if we are able to support both flows on a single set of endpoints we can deprecate the purely blinded endpoints. This also means we can remove any validator client config for "blindedness preference" it can just be inferred based on how the beacon node is configured.
> 

As reasoned above would prefer validator making a flow decision.

> To enable this we could:
> 
>     1. add a new header similar to the `Eth-Consensus-Version` to denote blinded vs unblinded
> 
>     2. add `{fork}-blinded ` variants to the existing `Eth-Consensus-Version` header
> 

if we allow validator to make a flow decision, we only need to add a block_source query param for both produceBlindedBlock and publishBlindedBlock for the BN to handle the request/response accordingly.

g11tech:
> In validator lodestar race two calls on builder/execution and picks one which resolves first after a cutoff (3sec) or the more paying block if both resolves before the cutoff. Again i wouldn't prefer for beacon to make fallback decisions again for the purposes of validator logic enforcing any timings/cutoff constraints.
> 

I forgot to mention that in lodestar we are also returning blockValue for both builder and execution (returns `0` pre getPayloadV2 availability i.e. pre capella) produceBlock/produceBlindedBlock 

and validator gets to make the choice as per a simple comparison as of now if both execution and builder blocks are produced within a cutoff (3sec)

As also observed in interop R&D channel discussion, a small overhaul of the produce/publish apis is needed for a comprehensive solution and for flexibility of various solutions for making block selection

g11tech:
> Interesting, we have the same logic but it exists in the beacon node not the validator client.
> 
got it!

> We could have the VC request a blinded/unblinded block preference via header and the beacon _should_ respond with the correct header but isn't strictly required to, and in the BN response we add a header to denote what the actual message type is (blinded/unblinded). With this model you could simply ignore anything unexpected in the VC (i.e. blinded request with unblinded response). The VC is still able to in parallel make one blinded and one unblinded request and include fallback logic

sounds good!

> I must have missed this, would you mind elaborating or linking to the discussion? Is there a general feeling the VC should be responsible for this? My general thought was type of thing would be handled in the BN and therefore wouldn't need to be considered in these APIs

https://discord.com/channels/595666850260713488/892088344438255616/1084096365572923392

jimmygchen:
> We could have the VC request a blinded/unblinded block preference via header and the beacon should respond with the correct header but isn't strictly required to, and in the BN response we add a header to denote what the actual message type is (blinded/unblinded). With this model you could simply ignore anything unexpected in the VC (i.e. blinded request with unblinded response). The VC is still able to in parallel make one blinded and one unblinded request and include fallback logic

If we're adding new behaviour on top of the endpoint, I think it's worth considering bumping up the endpoint version?

> add {fork}-blinded  variants to the existing Eth-Consensus-Version header

Probably worth considering using a different header so:
1.  we don't have to separately parse this string to determine the fork or block type
2.  existing code that depends on the consensus version header won't have to change

realbigsean:
In the interest of moving this forward, I made [this issue](https://github.com/ethereum/beacon-APIs/issues/309) about consolidating the blinded/unblinded flows and adding blinded/unblinded signalling headers.  The main reasons we would have wanted to include those changes here is to keep us from having to implement deneb types on two endpoints, or to only have to bump and endpoint version once. If that's reason enough to include the consolidation changes here, I'll close the other issue. But my feeling is even if we want to consolidate the blinded/unblinded endpoints for deneb we will still keep the blinded endpoint around for a bit and because of this will want to add deneb types to it anyways. If we're OK with not incrementing the endpoint versions for this PR then I think it's OK to keep the consolidation + headers approach as a separate discussion and set of changes independent of this.

types/deneb/block_contents.yaml
james-prysm: so for block content ( not blinded block content) the blob_sidecars is still blinded?

rkapka: > so for block content ( not blinded block content) the blob_sidecars is still blinded?

No. The property is of type `BlobSidecars`, not `BlindedBlobSidecars`.

types/deneb/execution_payload.yaml
rkapka, line 1: Do we need a new file when there are no changes between Capella and Deneb? I would rather reuse Capella definitions.

types/deneb/state.yaml
rkapka, line 1: Do we need a new file when there are no changes between Capella and Deneb? I would rather reuse Capella definitions.

jimmygchen: The Deneb state uses a different execution payload header (pointing to the Deneb version), not sure if there's a way to do this without introducing a new file

types/deneb/execution_payload.yaml
jimmygchen: There's a new field `excess_data_gas` in the `ExecutionPayloadCommon` object

rkapka: I didn't catch it. Thanks.

james-prysm:
> There’s no spec definition of the BlockContents struct for SSZ serialization. This struct is solely for communication here and not used in consensus elsewhere. Should the SSZ definition be included here? Should we add some similar type's definition to the consensus spec? Can we describe this SSZ type in terms of spec objects? (e.g., SignedBeaconBlock with BlobSidecars appended?)

was wondering if there were any continued discussions of this somewhere.

g11tech (dismissed):
lodestar has implemented the full contents signing flow

rolfyone:
Where are we at with this? Can we fix conflicts and think about merging?

types/deneb/block.yaml
```diff
@@ -52,7 +52,7 @@ Deneb:
             $ref: './execution_payload.yaml#/Deneb/ExecutionPayload'
 
   BeaconBlock:
-    description: "The [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/deneb/beacon-chain.md#beaconblock) object from the CL Deneb spec."
+    description: "The [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/Deneb/beacon-chain.md#beaconblock) object from the CL Deneb spec."
```
realbigsean, line 55:
```suggestion
    description: "The [`BeaconBlock`](https://github.com/ethereum/consensus-specs/blob/master/specs/deneb/beacon-chain.md#beaconblock) object from the CL Deneb spec."
```

realbigsean:
Fixed the conflicts @rolfyone 

I'm still interested it people have thoughts on this: https://github.com/ethereum/beacon-APIs/pull/302#issuecomment-1491182775 

But that's it from my POV

rolfyone:
add an entry in the changes for one :)

mcdee (dismissed):

rolfyone (approved):
LGTM, feel free to add a PR making changes closer to what you want, but at least this should be adequate as a start.

PR #303, opened 2023-02-14
Make fork choice validity enum lower case

tbenr:
All enums are lower case but this has been merged as UPPER CASE.


rolfyone (approved):
LGTM

PR #304, opened 2023-02-24
Added withdrawals endpoint to fetch withdrawals from a parent state

### LINKED ISSUE #301: Endpoint for expected withdrawals for a slot
Add an endpoint to fetch the expected withdrawals for some slot.

The endpoint path proposed is /eth/v1/beacon/states/{state_id}/withdrawals.

With the capella upgrade withdrawals are embedded in the payload attributes and currently there is no way to access the withdrawals outside the beacon node for validation. 

With this API, withdrawals in the payload of externally built blocks can be compared with the expected withdrawals to ensure that the two match.

rolfyone:
Added into a new route, `/eth/v1/builder`, as it's different to the existing beacon state fetch paradigm, and a new Tag for builders seems appropriate.

This endpoint is probably suited to allowing for returning SSZ via octet-stream also, so added that in. The idea would be similar to blocks, it would just return the data object as the SSZ result.

It would be possible to add withdrawals_root, but it'd be at the same level as finalized, so it'd be a little less clean. I've left that out for now.

Fixes #301.

### DIFF
--- a/apis/builder/states/expected_withdrawals.yaml
+++ b/apis/builder/states/expected_withdrawals.yaml
@@ -1,0 +1,61 @@
+get:
+  operationId: "getNextWithdrawals"
+  summary: "Get the withdrawals that are to be included for the block built on the specified state."
+  description: |
+    Get the withdrawals computed from the specified state, that will be included in the block 
+    that gets built on the specified state.
+  tags:
+    - Builder
+  parameters:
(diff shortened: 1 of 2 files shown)


michaelsproul (approved):
Looks good to me!

rkapka (approved):

PR #305, opened 2023-02-24
SSE for payload attributes

### LINKED ISSUE #244: SSE subscription for newPayload event to trigger block building
Block builders in the MEV-Boost ecosystem need a few bits of information to start block building as soon as possible, which is (a) at the newPayload event, or (b) after a timeout if the BN missed a slot, so block production is still triggered.

Currently a [custom BN fork](https://github.com/flashbots/prysm) is needed to trigger block production in time.

We'd like to propose adding another SSE event subscription to the beacon-APIs, which can be used to trigger block production.

The necessary payload fields are
- `slot`
- `blockHash`
- `timestamp`
- `prevRandao`
- `withdrawals` 

Current implementation [here](https://github.com/flashbots/prysm/blob/develop-boost/beacon-chain/blockchain/execution_engine.go#L291-L296)

I can open a PR with the changes to https://github.com/ethereum/beacon-APIs/blob/master/apis/eventstream/index.yaml but wanted to open the issue to discuss this first.

---

For reference, this is currently implemented in a custom BN fork with a [`notifyBuildBlock` method](https://github.com/flashbots/prysm/blob/develop-boost/beacon-chain/blockchain/execution_engine.go#L258)
which send `BuilderPayloadAttributes`, and is triggered on `onBlock` and other events and after a timeout has been reached (see `process_block.go`)

Edit 1: should the event be sent again if the likely parent changes due to new attestations?

michaelsproul:
Closes #244.

Define a new server sent event (SSE) for payload attributes.

The difficult part of specifying this event was deciding on an encoding for the payload attributes. Payload attributes presently only exist in the `execution-specs` repo where they are always encoded with `camelCase` idents and big-endian hex integers. We have two options for encoding them here:

1. Use the `execution-apis` encoding and break consistency with the rest of `beacon-apis`.
2. Define a new encoding for payload attributes here in order to maintain consistency.

There are pros and cons in favour of each approach. Option (1) potentially minimises the number of encodings that clients need to maintain for the payload attributes data structure, which could be good for implementation complexity. On the other hand, the break in consistency might make it hard for implementations to use a different encoding for _just these fields_. After discussions with @rolfyone we came to the conclusion that the consistency of option (2) is preferable, and not difficult to implement in either Lighthouse or Teku. **We would like feedback from other CL devs about their preference as well.**

The way the events are currently specified is quite loose, so I've added some extra descriptive text to convey the choice of encoding as clearly as possible. Perhaps in future it would be good to clean up the events API so that the returned events are given full schemas in their own right.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -32,6 +32,7 @@ get:
             - contribution_and_proof
             - light_client_finality_update
             - light_client_optimistic_update
+            - payload_attributes
   responses:
     "200":
       description: Opened SSE stream.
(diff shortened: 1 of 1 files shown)


rolfyone:
Looks good, didn't mark approved, will let others take a look first so any discussions can happen...

apis/eventstream/index.yaml
```diff
@@ -91,6 +92,32 @@ get:
...
+                - `version`: the identifier of the beacon hard fork at `proposal_slot`, e.g.
+                  `"bellatrix"`, `"capella"`.
+                - `proposal_slot`: the slot at which a block using these payload attributes may be
+                   built.
+                - `parent_block_root`: the beacon block root of the parent block to be built upon.
+                - `parent_block_number`: the execution block number of the parent block.
+                - `parent_block_hash`: the execution block hash of the parent block.
+                - `proposer_index`: the validator index of the proposer at `proposal_slot` on
+                   the chain identified by `parent_block_root`.
+                - `payload_attributes`: beacon-style encoding of `PayloadAttributesV<N>` as
```
rkapka, line 111: I doubt the term `beacon-style encoding` will be understandable. I would rephrase this as `Beacon API encoding`.

dapplion (commented):
Is the node expected to emit this event for the blocks its connected validators want to propose or for every slot? 

What should suggested_fee_recipient be set too? If the previous question is true then the response should be agnostic of who is connected to the bn so spec it to set to 0 address always

michaelsproul:
@dapplion In my PoC implementation I've added a flag to cause payload attributes to be sent with every fcU. Without that flag Lighthouse would just send payload attributes for the blocks it is proposing. I don't think it matters what the fee recipient is set to as the builders should be ignoring that anyway. IMO the BNs driving block builders should probably be distinct from the ones used by validators but if someone _did_ want to do this they could configure their real fee recipient in their BN config and use that for everything.

g11tech (approved):
@dapplion it should be pretty easy to add this event in the proposed (ssz json format) for on `prepareNextSlot` which we run in last 1/3 part of the slot to usually send a block building fcU (for next slot)
Adding a flag as suggested by @michaelsproul sounds reasonable. :+1:

apis/eventstream/index.yaml
michaelsproul: thanks for the suggestion, I've tried to improve the wording in this commit 64815e233e5ebbceb4465c9d0d39f02461a0c5d6

mcdee (approved):

apis/eventstream/index.yaml
```diff
@@ -91,6 +92,32 @@ get:
...
+
+                - `version`: the identifier of the beacon hard fork at `proposal_slot`, e.g.
+                  `"bellatrix"`, `"capella"`.
+                - `proposal_slot`: the slot at which a block using these payload attributes may be
+                   built.
+                - `parent_block_root`: the beacon block root of the parent block to be built upon.
+                - `parent_block_number`: the execution block number of the parent block.
+                - `parent_block_hash`: the execution block hash of the parent block.
+                - `proposer_index`: the validator index of the proposer at `proposal_slot` on
+                   the chain identified by `parent_block_root`.
```
ralexstokes, line 110: is there a reason we have all of this data, instead of just the minimum required data?

following the lead of the builder APIs, we can uniquely identify a proposal with the tuple `(proposal_slot, proposer_index, parent_block_hash)`

which suggests we can drop `parent_block_root` and `parent_block_number`

ralexstokes (commented):
looking good!

apis/eventstream/index.yaml
michaelsproul: @ralexstokes I figured the block number is useful for stateless dummy builders that just return empty payloads. We use such a builder to run blockprint. It wouldn't be hard to track the block number but it would require separately looking up data from a CL/EL.

The parent block root is potentially interesting to differentiate the case where an identical payload ends up in two competing beacon blocks. Although arguably this case _shouldn't_ be differentiated because the EL state is identical. Another reason to keep it is that it grounds the data in something CL-related, so if the receiver wants to look up more info from the beacon API they have a concrete reference point (a slot is ambiguous, and a payload hash isn't natively understood).

ralexstokes: makes sense, if everyone else thinks they pull their own weight then lets leave them in

dapplion:
> @dapplion In my PoC implementation I've added a flag to cause payload attributes to be sent with every fcU. Without that flag Lighthouse would just send payload attributes for the blocks it is proposing. I don't think it matters what the fee recipient is set to as the builders should be ignoring that anyway. IMO the BNs driving block builders should probably be distinct from the ones used by validators but if someone _did_ want to do this they could configure their real fee recipient in their BN config and use that for everything. In other words I think it would be overkill and less flexible to mandate setting it to 0.

Sounds reasonable to me. Please could you mention that in the spec?

michaelsproul:
Thanks @dapplion, added a comment just now in cb15cac95f36d9d2fd909a4395d0b8317500857b

dapplion (approved):

rolfyone (approved):
LGTM, and looks like general approval...

PR #311, opened 2023-04-16
Add 501 error response to submitSyncCommitteeSelections

nflaig:
Same as for `submitBeaconCommitteeSelections`,  `submitSyncCommitteeSelections` API is not implemented by a consensus client and should return 501.

https://github.com/ethereum/beacon-APIs/blob/f087fbf2764e657578a6c29bdf0261b36ee8db1e/apis/validator/beacon_committee_selections.yaml#L40-L41


rolfyone (approved):
LGTM

PR #321, opened 2023-05-29
Update publishBlindedBlockV2 request schema for Deneb

jimmygchen:
Updated the `publishBlindedBlockV2` to use the `Deneb.SignedBlindedBlockContents` schema for Deneb.

 I think the blob signing endpoints PR #302 was merged right after `publishBlindedBlockV2` was added in #317, and the `publishBlindedBlockV2` method was missing a schema change for Deneb.


rkapka:
Thanks for this! Can you also fix `publishBlockV2`? It uses `Deneb.SignedBeaconBlock` instead of `Deneb.SignedBlockContents`.

jimmygchen:
> Thanks for this! Can you also fix `publishBlockV2`? It uses `Deneb.SignedBeaconBlock` instead of `Deneb.SignedBlockContents`.

Good catch, I've updated `publishBlockV2`. Thanks!

rolfyone (approved):
Good catch - LGTM.

PR #326, opened 2023-06-13
Post release cleanup

rolfyone:


CHANGES.md
```diff
@@ -9,16 +9,6 @@ There are likely to be descriptions etc outside of the list below, but new query
 
 | Endpoint                                                                                                                                    | [Lighthouse](https://github.com/sigp/lighthouse) | [Lodestar](https://github.com/ChainSafe/lodestar) | [Nimbus](https://github.com/status-im/nimbus-eth2) | [Prysm](https://github.com/prysmaticlabs/prysm) | [Teku](https://github.com/ConsenSys/teku) |
 |---------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------|---------------------------------------------------|----------------------------------------------------|-------------------------------------------------|-------------------------------------------|
-| [#317](https://github.com/ethereum/beacon-APIs/pull/317) `POST /eth/v2/beacon/blocks` and `POST /eth/v2/beacon/blinded_blocks`              |                                                  |                                                   |                                                    |                                                 |                                           |
```
michaelsproul, line 12: Shouldn't these move to their own v2.4.1 section rather than being deleted?

rolfyone: i added them to the 2.4.1 release? want to create a list here as well? I figured once they were in notes there wasn't much point...

rolfyone: for reference: https://github.com/ethereum/beacon-APIs/releases/tag/v2.4.1

mcdee: That is going to be a bit of a pain for anyone who wants to know the changes between arbitrary versions, they'd have to look up the tags between them and read the file for each one.  Changelogs are usually presented with versions as headers to make life easier for those reading them.

michaelsproul: Yep, lets keep the changes here in the file. We already have them, seems unnecessary to delete them

rolfyone: ok, just going to be annoying to maintain potentially for me... but will give it a go

rolfyone:
I can adjust this if its not what you meant, just that the 2.4.1 isn't visible and i'd like to also see that, so merging, and will deal with it if the formatting isn't what you expected...

PR #331, opened 2023-06-23
Made eth-consensus-version required on new blocks endpoints

rolfyone:
When initially adding eth-consensus-version to interfaces it wasn't able to be required due to backwards compatibility issues.

Given these interfaces are new, we should set it as a required header. This will simplify the determination of whether we're reading a deneb blockContents or a pre-deneb signed block.

Potentially we deprecate the other interface once deneb comes around, but that's up for discussion, and I'm not putting that into this PR, just making this header required for the new interfaces.

### DIFF
--- a/apis/beacon/blocks/blinded_blocks.v2.yaml
+++ b/apis/beacon/blocks/blinded_blocks.v2.yaml
@@ -43,7 +43,7 @@ post:
         $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion'
       required: true
       name: Eth-Consensus-Version
-      description: "Version of the block being submitted, if using SSZ encoding."
+      description: "Version of the block being submitted."
   requestBody:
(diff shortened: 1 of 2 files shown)


apis/beacon/blocks/blinded_blocks.v2.yaml
```diff
@@ -41,7 +41,7 @@ post:
     - in: header
       schema:
         $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion'
-      required: false
+      required: true
       name: Eth-Consensus-Version
       description: "Version of the block being submitted, if using SSZ encoding."
```
jimmygchen, line 46: I'm assuming we want this always required for both SSZ & JSON? if so we can probably remove the "if using SSZ encoding" part.

rolfyone: yes if it's marked required it's required for all encodings..

PR #336, opened 2023-07-08
expand sync committee docs

ralexstokes:
I wasn't sure what the `validator_aggregates` were in the sync committee endpoint

@michaelsproul confirmed they are the sync subcommittee assignments and I have added this information to the API docs


rolfyone (approved):
LGTM

Issue #309, opened 2023-03-20
Consolodating blinded/unblinded flow


realbigsean:
Discussed here: https://github.com/ethereum/beacon-APIs/pull/302

General suggested approach so far: 

> add a new header similar to the Eth-Consensus-Version to denote blinded vs unblinded

> We could have the VC request a blinded/unblinded block preference via header and the beacon should respond with the correct header but isn't strictly required to, and in the BN response we add a header to denote what the actual message type is (blinded/unblinded). With this model you could simply ignore anything unexpected in the VC (i.e. blinded request with unblinded response). The VC is still able to in parallel make one blinded and one unblinded request and include fallback logic

Currently there isn't clarity in the spec around whether the VC or BN should handle things like fallback-to-local behavior.  Allowing the BN to choose the return type makes it clear that fallback in the BN is supported, but it seems that some prefer this to be disable-able. We will need to add `blockValue` to the beacon API to make sure fallback-to-local is fully supported in the VC.  And similarly this `shouldOverrideBuilder` flag will need to be made available to the VC eventually https://github.com/ethereum/execution-apis/pull/388. Would it make sense to move towards consensus on the question of "who picks the payload at the end of the day, the BN or VC"?  Being in-between on this makes the APIs more complicated and has created a lack of clarity around what the beacon node will do during the proposal flow. It'd be nice to iron this out in the spec.

So far the APIs seem to suggest it's the beacon node that should pick the payload (it has the `blockValue` information) but *should* this be how it works, or should validators choose? If validators should choose, should we think about separating the blinded API from the beacon node generally and instead move towards VC directly connecting to the builder API?

rolfyone:
I'm not sure there's an easy answer.
Teku has supported blinded api usage for a while, and there's really no reason the VC needs all the block data. Having said that, by supporting 2 sets of api, it means they're free to choose, so in instances like vouch where full blocks are required, they're not prevented from having that available. There's also SSZ vs non SSZ which enters into this whole complicated set of scenarios...

Being too prescriptive at the api level would force clients to do very specific things, and I'm not sure we need to be too closely prescribing how a client has to act in this regard...

Is there a concrete example of an actual issue we're trying to resolve? 

If we really did want to lock this down, the VC doesn't actually need the entire new block at all, so i'd probably be going a lot further with it, not just forcing the use of blinded or similar...

[...]

PR #339, opened 2023-07-16
Add block v3 endpoint.

mcdee:
As per #309 this adds a version 3 `block` endpoint that consolidates the current blinded and unblinded endpoints.  Features of this endpoint:

- a single endpoint to fetch blocks regardless of type
- allows the beacon node to decide (with help from the EL) if it should use MEV relays for execution payload
- provides the returned block's payload to the proposer
- metadata allows for simpler handling of decoding the returned data

### DIFF
--- a/apis/validator/block.v3.yaml
+++ b/apis/validator/block.v3.yaml
@@ -2,11 +2,16 @@ get:
   tags:
     - ValidatorRequiredApi
     - Validator
-  operationId: "produceBlockV2"
+  operationId: "produceBlockV3"
   summary: "Produce a new block, without signature."
-  deprecated: true
   description: |
-    Requests a beacon node to produce a valid block, which can then be signed by a validator.
+    Requests a beacon node to produce a valid block, which can then be signed by a validator. The
+    returned block may be blinded or unblinded, depending on the current state of the network as
+    decided by the execution and beacon nodes.
+
+    The beacon node must return an unblinded block if it obtains the execution payload from its
+    paired execution node. It must only return a blinded block if it obtains the execution payload
+    header from an MEV relay.
 
     Metadata in the response indicates the type of block produced, and the supported types of block
     will be added to as forks progress.
@@ -44,10 +49,14 @@ get:
       headers:
         Eth-Consensus-Version:
           $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+        Eth-Blinded-Payload:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Blinded-Payload'
+        Eth-Payload-Value:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Payload-Value'
       content:
         application/json:
           schema:
-            title: ProduceBlockV2Response
+            title: ProduceBlockV3Response
             type: object
             properties:
               version:
@@ -60,10 +69,12 @@ get:
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Altair.BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Bellatrix.BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Capella.BeaconBlock"
+                 - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Capella.BlindedBeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlockContents"
+                 - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlindedBlockContents"
         application/octet-stream:
           schema:
-            description: "SSZ serialized block bytes. Use Accept header to choose this response type, version string is sent in header `Eth-Consensus-Version`."
+            description: "SSZ serialized block or blinded block bytes. Use Accept header to choose this response type, version string is sent in header `Eth-Consensus-Version` and block type in `Eth-Blinded-Payload`."
     "400":
       description: "Invalid block production request"
       content:
--- a/apis/validator/block.v3.yaml
+++ b/apis/validator/block.v3.yaml
@@ -59,7 +59,7 @@ get:
             title: ProduceBlockV3Response
             type: object
             properties:
-              consensus-version:
+              version:
                 type: string
                 enum: [ phase0, altair, bellatrix, capella, deneb ]
                 example: "phase0"
--- a/apis/validator/block.v3.yaml
+++ b/apis/validator/block.v3.yaml
@@ -59,10 +59,16 @@ get:
             title: ProduceBlockV3Response
             type: object
             properties:
-              version:
+              consensus-version:
                 type: string
                 enum: [ phase0, altair, bellatrix, capella, deneb ]
                 example: "phase0"
+              blinded-payload:
+                type: boolean
+                example: false
+              payload-value:
+                type: string
+                example: "12345"
               data:
                 oneOf:
                  - $ref: '../../beacon-node-oapi.yaml#/components/schemas/BeaconBlock'
--- a/apis/validator/block.v3.yaml
+++ b/apis/validator/block.v3.yaml
@@ -49,10 +49,10 @@ get:
       headers:
         Eth-Consensus-Version:
           $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
-        Eth-Blinded-Payload:
-          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Blinded-Payload'
-        Eth-Payload-Value:
-          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Payload-Value'
+        Eth-Execution-Payload-Blinded:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Execution-Payload-Blinded'
+        Eth-Execution-Payload-Value:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Execution-Payload-Value'
       content:
         application/json:
           schema:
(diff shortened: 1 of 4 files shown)


beacon-node-oapi.yaml
```diff
@@ -421,3 +424,13 @@ components:
       required: true
       schema:
         $ref: '#/components/schemas/ConsensusVersion'
+    Eth-Blinded-Payload:
+      description: Required in response so client can deserialize returned json or ssz data to the correct object.
```
rkapka, line 428:
```suggestion
      description: Required in response so client can deserialize returned JSON or SSZ data to the correct object.
```

mcdee: The existing `Eth-Consensus-Version` uses lower-case so I stuck with it for consistency.

apis/validator/block.v3.yaml
```diff
@@ -0,0 +1,92 @@
...
+  responses:
+    "200":
+      description: Success response
+      headers:
+        Eth-Consensus-Version:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+        Eth-Blinded-Payload:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Blinded-Payload'
+        Eth-Payload-Value:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Payload-Value'
```
dapplion, line 55: Headers are required for SSZ returns. Should the added two headers be also present as JSON values to the `application/json` response type? Otherwise, why include version as property but not the payload value.

dapplion (commented):
Looks good! Minor nit on mirroring header values

apis/validator/block.v3.yaml
mcdee: Yeah, wasn't sure about that.  In the past we added the version to the JSON to allow for decoding, but it has pretty much been replaced with the header values now (as you don't need to do any mucking around with parsing streams to obtain the values from the headers).  I'm not averse to putting these new values in the JSON metadata, but not sure if they add significant value any more.

dapplion: > do we do something controversial and just return the json object not wrapped in a 'data' and have it just the json object or the ssz object determined by header?

It would definitely break the standard from other routes and prevent allowing to add more fields in the future. I would mirror the header values in the JSON type for overall consistency. It's easier to parse a JSON reply than headers for client side tooling that deals with JSON only

mcdee: @dapplion yeah I'm thinking the same.  If the metadata fields are in the JSON then it will allow parsing of the JSON separate from the HTTP response (e.g. if the JSON is saved and parsed later).  I'll update accordingly.

mcdee: Updated.  One question that rises from this is the naming of the fields: do we want to explicitly state which values are related to the execution payload, _i.e._

- `blinded-execution-payload` rather than `blinded-payload`
- `execution-payload-value` rather than `payload-value`

(and equivalent change for the header fields).  Obviously this makes them a little more verbose, but is safer if we ever end up with other payloads in the beacon block.  Thoughts?

rolfyone: if we went that way, i'd probably suggest something like
```
execution-payload-value
execution-payload-blinded
```
i don't mind either way on whether we expand it...

mcdee: Have updated to use these names, agree that they are a bit clearer.  I think that this looks like the best way forward; 
@dapplion @rolfyone agree/disagree?

apis/validator/block.v3.yaml
```diff
@@ -59,10 +59,16 @@ get:
             title: ProduceBlockV3Response
             type: object
             properties:
-              version:
+              consensus-version:
```
dapplion, line 62: Unless there's a strong need to namespace another `.*version.*` property would prefer to keep the previous name for consistency

rolfyone: we use 'version' in several endpoints in the metadata, should probably leave it at that I guess, but I'm not going to die on that hill...

mcdee: Okay put it back to just 'version'.

rolfyone (approved):
LGTM will just make sure others are happy...

dapplion (approved):
Looks good!

g11tech (changes requested):
1. ~~can't find `produceBlindedBlockV3` in PR, is this a mistake or intentional~~ 
  ahh realized this is replacement for both, so then can be add another query param like blockSource ("builder", "execution" or "any") 
2. can be also add `shouldOverrideBuilder` flag in response so that validator driven architectures of making the choice (like lodestar validator or may be vouch) can consume this info
3. can we also add the blockSource ("builder" or "execution") to the response apart from the query to reflect the selection here (especially in case of any), also gives a way to figure out that if the publishBlock call is binded to a particular beacon in case of blinded response

mcdee:
> 1. can't find `produceBlindedBlockV3` in PR, is this a mistake or intentional

Not sure to what this refers.  There is a `produceBlindedBlock` operation, but that is specifically for producing blinded blocks alone.  The operation ID of this call is `produceBlockV3`.

>  2. can be also add `shouldOverrideBuilder` flag in response so that validator driven architectures of making the choice (like lodestar validator or may be vouch) can consume this info

Beacon nodes will by default return the highest value block from either the relay(s) or the local payload.  However, if the beacon node's circuit breaker is triggered or the execution client specifies that the local payload should be used then the local payload will be returned regardless of value.  Does knowing why the full payload was returned (higher value / consensus circuit breaker / execution shouldOverrideBuilder) have any value here?

> 3. can we also add the blockSource ("builder" or "execution") to the response

This can be inferred from the returned payload:
- blinded => comes from the builder
- unblinded => comes from the execution node

mcdee:
1. I think that although the idea of having validator-controlled and beacon node-controlled selection of blocks as separate streams was nice in theory, in reality the beacon node is the entity that decides from where to fetch the execution payload.  The validator client ultimately has no choice in the matter; it can only use what the beacon node returns, regardless of what it asks for.  As such, I'm not sure what the purpose of the suggested flags will serve (specifically, in which situation would the validator client not ask for "any" as their block, given they want to maximize the value of the execution payload).

2. as @arnetheduck says in https://github.com/ethereum/beacon-APIs/issues/309#issuecomment-1498814540 we want the beacon node to be stateless as much as we can manage, which would require returning full blocks whenever possible

g11tech:
> 1. I think that although the idea of having validator-controlled and beacon node-controlled selection of blocks as separate streams was nice in theory, in reality the beacon node is the entity that decides from where to fetch the execution payload.  The validator client ultimately has no choice in the matter; it can only use what the beacon node returns, regardless of what it asks for.  As such, I'm not sure what the purpose of the suggested flags will serve (specifically, in which situation would the validator client not ask for "any" as their block, given they want to maximize the value of the execution payload).
> 
>     2. as @arnetheduck says in [Consolodating blinded/unblinded flow #309 (comment)](https://github.com/ethereum/beacon-APIs/issues/309#issuecomment-1498814540) we want the beacon node to be stateless as much as we can manage, which would require returning full blocks whenever possible

shouldn't these be left to implementation and support the implementation types. again lodestar can transition to this paradigm but should be nice to have and support implementation diversity.

mcdee:
Which flags are you after here?  The one that was under discussion was to tell the beacon node to not override use of a relay payload with a local payload (when it had been informed to either from its own circuit breaker or from the execution API `shouldOverrideBuilder` flag), but it seems to me that it makes more sense to make this a configuration option for the beacon node itself rather than a parameter passed each time a block is requested.

g11tech:
> Which flags are you after here? The one that was under discussion was to tell the beacon node to not override use of a relay payload with a local payload (when it had been informed to either from its own circuit breaker or from the execution API `shouldOverrideBuilder` flag), but it seems to me that it makes more sense to make this a configuration option for the beacon node itself rather than a parameter passed each time a block is requested.

lodestar current impl needs all the flags i mentioned before, but may be we will change our impl to push the block selection to beacon node.

but then atleast blockSource should be there in the response so that if the beacon node chooses blinded response even for executon block, (configured for e.g. via a startup flag for beacon node) then validator knows that the publish call is bound to the same beacon node.

beacon-node-oapi.yaml
```diff
@@ -421,3 +424,13 @@ components:
       required: true
       schema:
         $ref: '#/components/schemas/ConsensusVersion'
+    Eth-Execution-Payload-Blinded:
+      description: Required in response so client can deserialize returned json or ssz data to the correct object.
+      required: true
+      schema:
+        type: boolean
+    Eth-Execution-Payload-Value:
```
arnetheduck, line 432: Without the consensus value, there's no (reasonable) way for a VC to compare blocks for profit - should we add this here?

michaelsproul: Does Nimbus already compute this value during block proposal? Lighthouse sort of does, but approximates slightly (if there are slashable attestations we don't keep track of the rewards from both).

It would make VC block choosing more fair, as the heuristics Vouch uses currently seem to favour Prysm despite Prysm's blocks being less profitable than Lighthouse and Nimbus on average (according to my blockprint nodes, public dash coming soon).

rolfyone: could this be a new PR? this one is closed.

arnetheduck: > Does Nimbus already compute this value during block proposal? 

We do it for offline analysis for historical blocks (to pick out more detail than is needed for state transitions) but making it "online" would probably not be too hard - indeed, the VC currently has to do a lot of guesswork and it will likely always be wrong / skewed - further exacerbating the problem is that in that guesswork, it's not working with eth-denominated data usually but rather "fresh attestation count" and similar proxies which prevents the value from being compared with the execution profit (ie low cl+high el vs low cl+high el score)

arnetheduck: > could this be a new PR? this one is closed.

https://github.com/ethereum/beacon-APIs/pull/358

PR #341, opened 2023-07-23
Update syncing status requirements in getHealth

nflaig:
While implementing `syncing_status` query parameter (https://github.com/ethereum/beacon-APIs/pull/251), I noticed two issues with the current spec

1. Schema allows invalid http status codes (1-99), this will cause problems for most http servers
2. The API accepts user provided values but there is no error response for invalid values

### DIFF
--- a/apis/node/health.yaml
+++ b/apis/node/health.yaml
@@ -11,7 +11,7 @@ get:
       description: "Customize syncing status instead of default status code (206)"
       schema:
         type: integer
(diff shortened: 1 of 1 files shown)


rolfyone:
Looks reasonable, I'll see if i can get some feedback on it, thanks!

jimmygchen:
Looks good 👍

dapplion (approved):

PR #342, opened 2023-07-28
Fix casing of JSON metadata.

mcdee:
Casing of the JSON metadata for the v3 block API was in kebab case (as per headers) but should be in camel case as per all other JSON.


james-prysm:
looks good ! thanks

rolfyone (approved):

rolfyone:
this is only replacing `-` with `_` but the comment is casing... makes sense but i'll hold off merging until you confirm...

mcdee:
Yeah it's just the change from `-` to `_`.

Issue #348, opened 2023-09-11
Add Blob sidecar information to event stream


parithosh:
We use [Xatu](https://github.com/ethpandaops/xatu/) extensively for monitoring block/attestation arrival metrics on the beaconchain/testnets. This was extremely useful for us in the big block tests that help decide the values used for the size of blobs. Xatu gathers this data by [subscribing to the event stream](https://github.com/ethpandaops/xatu/blob/master/docs/sentry.md#beacon-api---event-stream), parsing the data, annotating it and forwarding it to a server for storage in a database. Extending the beaconAPI eventstream to include  events for blobs would greatly help in adding support downstream in Xatu. 

The events that would be nice to have are:
- blob arrival

Topics up for debate:
- would an event stream emit a block as valid before it gets blobs associated ? 
- when would we say a blob is available? when all its parts are received or when 2/3rd or whatever we need for reconstruction are received?

cc @dapplion (since he had some suggestions for a PR based on this issue)
cc @Savid for Xatu (can't tag sam for some reason)

mcdee:
I agree we should have a blob event that provides us with information on a blob.  At a first stab, I'd go for the following information in the event:

- block_root
- index
- slot
- versioned hash of blob
- KZG commitment

between them this should provide enough info for explorers and the like to index and reference them.

As to when the event stream emits information:
- blob should be emitted as soon as the blob is accepted from the P2P layer
- block should be emitted as soon as the block is accepted from the P2P layer
- head should be emitted once the head is updated in state

These would allow monitors to understand when the individual components of the block arrive, along with the time taken to update state.

dapplion:
Is it necessary to include the KZG commitment even if that data can be recovered from the block latter?

mcdee:
The thinking behind the KZG commitment is that you can use this information to see if the same blob is in multiple beacon blocks.  Not absolutely critical _per se_, but if the block to which the blob is attached is reorged it would be good to be able to track which block it turned up in again at the consensus layer.

dapplion:
> The thinking behind the KZG commitment is that you can use this information to see if the same blob is in multiple beacon blocks. Not absolutely critical _per se_, but if the block to which the blob is attached is reorged it would be good to be able to track which block it turned up in again at the consensus layer.

Would the versioned hash of blob by sufficient for this purpose?

dapplion:
- @mcdee let's move discussion to https://github.com/ethereum/beacon-APIs/pull/350

[...]

PR #350, opened 2023-09-11
Add blob_sidecar event

dapplion:
- Adds blob_sidecar event to SSE, see https://github.com/ethereum/beacon-APIs/issues/348 for rationale

**Discussion points**

- unsure to name the event `blob` or `blob_sidecar`. There won't be two different types of blob events so `blob` is shorter. But `blob_sidecar` matches the returned container summary.
- should the event include the kzg_commitment, the versioned hash or both? I tend to say that the version_hash is more useful since that's what the app layer is concerned about. Also one can derive the version hash, see [`kzg_commitment_to_versioned_hash`](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#kzg_commitment_to_versioned_hash)

Closes https://github.com/ethereum/beacon-APIs/issues/348


mcdee:
Rationale behind including the KZG commitment is that https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#beaconblockbody only contains the commitment.  So if (for example) you received a `blob` event but wanted to see if the same blob had shown up in an earlier (orphaned) block you could do so relatively simply.

Not sure about how useful this would be in the real world, but the commitment is the only way that I can see to link blobs to consensus blocks (beyond the obvious other information already in the event).

I would include the versioned hash as well, as that provides a similar function for the execution layer.

james-prysm:
> blob_sidecar matches the returned container summary.
casting my vote for blob_sidecar

rkapka (dismissed):
looks good to me

rolfyone (dismissed):
LGTM, just needs to be added to the change list.

rolfyone (approved):
LGTM

PR #353, opened 2023-09-15
Move state endpoint to `beacon` namespace

rkapka:
The [/eth/v2/debug/beacon/states/{state_id}](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Debug/getStateV2) endpoint is widely used by regular users and as such should probably not be present in the `debug` namespace. I think the original reason for putting it there was the large size of the response, but currently requesting all validators from `/eth/v1/beacon/states/head/validators` results in an even larger response.

This PR:
- creates a new `/eth/v1/beacon/states/{state_id}` endpoint (the debug version is `v2`, but since we are moving it to a new namespace, in my opinion versioning should start anew)
- deprecates the debug endpoint


rolfyone (commented):
LGTM, add this to the change list so we don't forget :)

rolfyone (approved):
LGTM

arnetheduck:
Uh, this was merged a bit quickly without time for feedback, but the condition we've discussed in the past for having the state endpoint in the official part of the api was to limit it to SSZ only - I think that change should be made, ie it's not feasible to create an expectation that client open up providing JSON for the state, which soon will reach GB sizes.

arnetheduck:
There was also discussion of fixing the flags encoding for json which is irregular with respect to its underlying intent.

rolfyone:
sorry @rkapka can you re-open a pull request for this to allow discussion to happen? I've reverted it from master after out of band discussions.

PR #357, opened 2023-09-19
Move state to beacon

rolfyone:
Re-opening PR from #353 , please discuss here any considerations about whether this should be moved.

From original submission from @rkapka 

The [/eth/v2/debug/beacon/states/{state_id}](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Debug/getStateV2) endpoint is widely used by regular users and as such should probably not be present in the debug namespace. I think the original reason for putting it there was the large size of the response, but currently requesting all validators from /eth/v1/beacon/states/head/validators results in an even larger response.

This PR:

creates a new /eth/v1/beacon/states/{state_id} endpoint (the debug version is v2, but since we are moving it to a new namespace, in my opinion versioning should start anew)
deprecates the debug endpoint


arnetheduck:
In JSON format, this API generates ~400mb presently - this is the main reason it has stayed in debug: it is inefficient to generate and transfer.

While the JSON format was convenient pre-mainnet to debug clients (hence the original namespace), currently this task has been superseded by SSZ-to-JSON converters and de facto, clients only use the SSZ version of this endpoint - using JSON is simply too slow for servers and clients alike and surrounding tooling and SSZ libraries for all imaginable languages have improved to the point that JSON is no longer useful to maintain.

As such, I'd like to see that as part of the move, we explicitly remove the JSON support entirely from the request, along with any types it references - this removes the expectation that JSON will continue to work.

Conveniently, this also solve an issue with defining a [canonical JSON spec](https://github.com/ethereum/consensus-specs/pull/3506) - it turns out that the beacon state is the only place where the beacon API deviates from the current iteration of the PR so the beacon API would naturally become canonical-json-compatible without further ado.

I'll also point out that mostly, this rename of the URL is strictly unnecessary and that the compatibility can be maintained if we continue to use `debug` in the URL but move it to the beacon section regardless - this pragmatic solution avoids the need for the entire community to update tooling, web pages, tutorials and so on and avoids the need of a transition period - everything that used the SSZ endpoint continues to work as before and the only loss is in aesthetics - this is a preferable approach in my mind.

beacon-node-oapi.yaml
```diff
@@ -65,6 +65,8 @@ paths:
 
   /eth/v1/beacon/genesis:
     $ref: "./apis/beacon/genesis.yaml"
+  /eth/v1/beacon/states/{state_id}:
+    $ref: "./apis/beacon/states/state.yaml"
```
arnetheduck, line 69:
```suggestion
  /eth/v1/debug/states/{state_id}:
    $ref: "./apis/beacon/states/state.yaml"
```

per comments elsewhere, this URL change is unnecessary and remains a cheap way to provide superior backwards compatibility.

rolfyone: its definitely not going to be `/eth/v1/debug/beacon/states/{state_id}` - that route was used, and Im not sure why we'd want a brand new route here `/eth/v1/debug/states/{state_id}`, can you clarify what you mean exactly?

just to be explicit, the current route that offers this is `/eth/v2/debug/beacon/states/{state_id}`, and the suggestion is that becomes `/eth/v1/beacon/states/{state_id}`

arnetheduck: the practical outcome would be that json support continues to live in the debug section of the spec while ssz becomes "enshrined" under the beacon section (but with the same url for backwards compatibility)

rolfyone:
> As such, I'd like to see that as part of the move, we explicitly remove the JSON support entirely from the request, along with any types it references - this removes the expectation that JSON will continue to work.

To me, this is not a move. If we want a new route with new functionality, we shouldn't be calling this a 'move'.

> Conveniently, this also solve an issue with defining a [canonical JSON spec](https://github.com/ethereum/consensus-specs/pull/3506) - it turns out that the beacon state is the only place where the beacon API deviates from the current iteration of the PR so the beacon API would naturally become canonical-json-compatible without further ado.

This is not really a primary concern, especially given it's currently only a proposed spec change.

> While the JSON format was convenient pre-mainnet to debug clients (hence the original namespace), currently this task has been superseded by SSZ-to-JSON converters and de facto, clients only use the SSZ version of this endpoint - using JSON is simply too slow for servers and clients alike and surrounding tooling and SSZ libraries for all imaginable languages have improved to the point that JSON is no longer useful to maintain.

On multiple occasions people have discussed json content from the debug/states endpoint, so I'm not convinced that this is an accurate assertion to make. While it may not be used directly by a CL, at the very least i've had researchers asking me directly about json data from states, and removing that option would mean they need to dump states and pretty print them via another mechanism.

arnetheduck:
>  While it may not be used directly by a CL, at the very least i've had researchers asking me directly about json data from states, and removing that option would mean they need to dump states and pretty print them via another mechanism.

keeping debug json support under the debug part of the spec solves this concern, though I suspect any working solution would satisfy the need.. indeed, what will likely happen once this endpoint is "blessed" is that clients will start asking for a flag to enable json support and at that point, it'll be easier to use a pretty printer since json endpoints won't be ubiquitous as they are today.

michaelsproul:
I'm going to switch my position from ambivalence to a vote against changing anything. I don't think it's worth the ecosystem churn, the bike-shedding, etc. I think the `/debug/` in the URL can be treated as purely cosmetic and clients can make their own choices about how to expose it. Lighthouse has no issue serving a few GB of JSON, and the state reconstruction itself takes only a few seconds on our tree-states branch (which will become default soon). Therefore I'm also against removing JSON: clients can handle it however they want, and users can watch their node OOM if they haven't allocated enough RAM.

rolfyone:
Ok closing this, i'd prefer not to move it than split, and some fair points raised above. Thanks for persisting so we could work it out all!

PR #362, opened 2023-09-26
make getBlobSidecars response title more consistent with the operationID

roberto-bayardo:


jimmygchen (approved):
Nice catch! 
I think I missed this one when I was renaming `/blobs` back to `/blob_sidecars`.

Issue #333, opened 2023-07-02
Support SSZ for validators endpoint


mcdee:
The `/eth/v1/beacon/states/<state_id>/validators` endpoint returns a lot of data.  At the time of writing, this endpoint returns around 370MB of JSON.  This requires a fair amount of CPU and memory resources on both the server, which has to convert from its internal data structures to JSON, and the client, which has to do the opposite.

SSZ is already used in a number of other areas where there is large amounts of data, such as the beacon state, and would reduce the burden on both server and client.  The SSZ representation of the validators at time of writing is around 112MB, less than 1/3 of the size.  Additionally, the CPU and memory required for encoding and decoding should be reduced due to the similarity of the native and SSZ formats.

The reason that this is an issue rather than a PR is that it would require work from client teams, most notably they would need to support that beacon node API validator object, which contains the spec validator object plus some additional fields.  As such, I would like to hear from client teams as to if this is something that they would or would not be interested in supporting.

dapplion:
Why not fetch the entire state via the debug endpoint as SSZ? Most of its SSZ serialized size is the validators list, and then its simple to produce a ValidatorResponse from the full state. Note that SSZ does not support strings so we would need to encode ValidatorStatus in some non-obvious way.

Given the current number of beacon nodes implementations I would prefer to offload data transformations to downstream tooling, and limit the beacon node's duty to expose all data necessary for such tools.

mcdee:
Yes that's an option, and one that I have played with (the size of state is broadly equivalent to the size of the validators response), but for the general-purpose consumer they won't know to do this.  Unless we deprecate the `/valdiators` endpoint entirely and point people to the `/state` endpoint, but if we were to do that we'd probably want to move the `/state` endpoint out of the `debug` namespace.

Agree that we'd need to define the validator status, but figure that would be part of the endpoint definition.

> Given the current number of beacon nodes implementations I would prefer to offload data transformations to downstream tooling, and limit the beacon node's duty to expose all data necessary for such tools.

This was certainly the initial intent (the `/validators` endpoint was a bit contentious because it didn't expose a simple spec object) but seems to have been diluted over time with the addition of various other endpoints such as rewards.  I don't have a problem with us still using that as a guiding principle for new endpoints, though.

dapplion:
A way I look at the debug / not distinction is:
- non-debug endpoints: can disrupt the node a bit but cannot kill it. These endpoints should be protected but are ok to be enabled by default on all users
[...]

PR #367, opened 2023-10-23
POST endpoints for `validators` and `validator_balances`

michaelsproul:
Some API users want to fetch large chunks of the validator set, and currently do so using the `id` param in the query string for `GET /eth/v1/beacon/states/{state_id}/validators`. However, query strings are usually subject to length limits on both the client and server side. For Lighthouse, we can't extend the length limit without forking our HTTP server (see https://github.com/sigp/lighthouse/issues/4333#issuecomment-1741929407).

To remedy this situation, this PR introduces a POST variant of the endpoint which is semantically identical, but passes the IDs and statuses in the request body. This is preferable to a GET request with body, as GETs with bodies [do not conform to HTTP standards](https://stackoverflow.com/questions/978061/http-get-with-request-body).

Currently, the only alternatives are sub-optimal:

- Make lots of small requests, which is slow on clients that load full states, particularly if done in parallel (see https://github.com/sigp/lighthouse/pull/4872).
- Make a single request for the _entire_ validator set. This has the downside of consuming a large amount of memory, and generating useless work on clients that are capable of loading just the relevant parts of a `BeaconState`.

I don't think this is a client-engineering problem, because designing a database around its ability to load partial states _is not_ likely to be space-efficient. The most space-efficient schemes known currently use state diffs, which don't (usually) provide the ability to be partially applied. Lighthouse's `tree-states` release track uses a variant of the state diffs described by Anton Nashatyrev [here](https://hackmd.io/G82DNSdvR5Osw2kg565lBA).

Hacking query string limits across clients is also unlikely to provide a long-term fix, as every new HTTP client that interacts with the beacon API may impose its own length-limit that breaks compatibility.


jclapis:
+1 to this, Rocket Pool uses this and runs into the limitation on `GET` request strings all the time. We have to limit our requests to 600 validators per call to get around the size limit, which ends up requiring around 50 parallel calls to get all of our validators in the set on Mainnet. When we call it for a state beyond the head, it does tend to slow things down. If we had a `POST` route where we could put all of the validators in, we could do it with a single call and probably remove a lot of the DB thrashing the clients do.

rolfyone:
I'd be happy for this, i do wonder if endpoints like this should support ssz or at least gzip compression, as the response would compress very well...

michaelsproul:
@rolfyone Agree we should pursue SSZ. I think there's broad support for it in https://github.com/ethereum/beacon-APIs/issues/333, just waiting for a PR to the spec.

cheatfate:
**GET** [/eth/v1/beacon/states/{state_id}/validator_balances](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidatorBalances) has same problem with `id` param used for filtering. So probably it should be also extended with `POST` call.

rolfyone:
can we add this to changelog then I'll approve and merge...

michaelsproul:
Updated the changelog and added a POST for `validator_balances` per @cheatfate's suggestion

dapplion:
Strict improvement, looking forward to adopt in Lodestar

dapplion (approved):
Strict improvement, looking forward to adopt in Lodestar

rkapka (approved):

rolfyone (approved):
LGTM - thanks everyone!

PR #369, opened 2023-10-31
Remove blob signing

realbigsean:
Changes for https://github.com/ethereum/consensus-specs/pull/3531

- update the `BlobSidecar` struct in the `GET` `blob_sidecars` endpoint
- remove `SignedBlobSidecar` type
- use `BlindedBeaconBlock` structs rather than `BlindedBlockContents` structs
- update `BlockContents` to wrap a `BeaconBlock`, an array of `KZGCommitment`'s an array of `KZGProof`'s, and an array of `Blob`'s

My thinking here is we want to maintain a stateless flow (the unblinded flow) so we will have to pass along blobs to the VC. And in this case I think we might as well pass along proofs so they don't need to be recomputed. In the stateful (blinded) flow we only need to pass along the object we're signing, so I think it can be kept maximally simple. The downside is it's less intuitive which types should be expected when on an endpoint like `/eth/v3/validator/blocks` and what the types represent exactly. 

The API changes here imply this block production flow:

**Stateless**
- BN gets payload, commitments, proofs, blobs from the EL
- BN assembles a block using the commitments and payload
- VC receives the block, blobs, and proofs from the BN
- VC Signs the block and returns everything
- BN computes the commitment inclusion proof, assembles the sidecars, publishes

**Stateful (blinded)**
- BN gets payload, commitments, proofs, blobs from the EL
- BN assembles a block using the commitments and payload
- BN caches payload, commitments, proofs, blobs
- VC receives the blinded block
- VC signs and returns the blinded block
- BN unblinds the block, computes the commitment inclusion proof, assembles the sidecars using cached proofs and blobs, publishes


types/deneb/blob_sidecar.yaml
```diff
@@ -6,83 +6,27 @@ Deneb:
     minItems: 0
     maxItems: 6
 
+  CommitmentInclusionProof:
```
mcdee, line 9: Given that we use `KZGCommitment` elsewhere, should we refer to this as `KZGCommitmentInclusionProof` or is this considered more generic?

dapplion: Updated field name to `kzg_commitment_inclusion_proof` https://github.com/ethereum/consensus-specs/pull/3531/commits/4a609cec574eca635a9e0980fe448089f8690ce1

apis/validator/block.v3.yaml
```diff
@@ -76,14 +76,14 @@ get:
...
               data:
                 oneOf:
-                 - $ref: '../../beacon-node-oapi.yaml#/components/schemas/BeaconBlock'
+                 - $ref: "../../beacon-node-oapi.yaml#/components/schemas/BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Altair.BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Bellatrix.BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Bellatrix.BlindedBeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Capella.BeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Capella.BlindedBeaconBlock"
                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlockContents"
```
rolfyone, line 85: remove this blockContents too?

realbigsean: This is where it's a little confusing and asymmetric. `BlockContents` is still used in the unblinded flow, because in order to keep it stateless we need to pass blobs along with the block back and forth. So this endpoint can return `Deneb.BlockContents` or `Deneb.BlindedBeaconBlock`

types/deneb/blob_sidecar.yaml
realbigsean: updated both the struct name and field name here: e9d053821a6e4dafe8651637322fb4ac4da01f6c

apis/validator/block.v3.yaml
g11tech: i would prefer a separate call by validator to pull blobs if it plans to reroute the publish since the data transfer is non trivial and will also resolve code spaghetti in the api, to address latency, validator call concurrently pull blobs while its signing the block

mcdee: That would not meet the requirement for statelessness, as it would be possible for the beacon node to exit between returning the data in this call and a separate request being made for the blobs.

realbigsean: >  i would prefer a separate call by validator to pull blobs if it plans to reroute the publish since the data transfer is non trivial and will also resolve code spaghetti in the api, to address latency, validator call concurrently pull blobs while its signing the block

I'm not sure this can be done without reintroducing the state requirement. Potentially we could set up an avenue for BNs to query ELs to check their local mempool for blobs matching a commitment, but I don't think this is really a great solution either. It might be better to mitigate the expense of transferring blob data by preferring SSZ transfer or preferring the blinded flow even when not using an external builder (this does put you at risk of missing a block proposal if you are using fallback between BNs however).

g11tech: beacon nodes are stateful, i am not sure why we are imposing the requirement of statefulness

g11tech (approved):

apis/validator/block.v3.yaml
realbigsean: I think it's been made a requirement to support both specifically in block production to support a greater variety of setups and make sure the proposal flow is generally more reliable

rolfyone: if this is the case we should probably delete the v2....

realbigsean: v2 supports both by way of having both the `/eth/v2/beacon/blocks` and `/eth/v2/beacon/blinded_blocks` endpoints, v3 just unifies the two

g11tech: ummm actually i also cant understand that why have v2 when we can extend v1 (its simply rename v2 to v1 in the codebase, there is no conceptual diff) and its still backward compatible... 

even for publish v2 could have been done away with by just extended v1 ....

types/deneb/block_contents.yaml
```diff
@@ -1,37 +1,36 @@
...
-
-  BlindedBlockContents:
-    type: object
-    description: "The required object for blinded block production according to the Deneb CL spec."
-    properties:
-      blinded_block:
-        $ref: "./block.yaml#/Deneb/BlindedBeaconBlock"
-      blinded_blob_sidecars:
-        $ref: "./blob_sidecar.yaml#/Deneb/BlindedBlobSidecars"
+      kzg_proofs:
```
g11tech, line 22: if we are assuming statelessness, are we requiring here for the beacon to build inclusion proofs for the blobs before transmitting them?

types/deneb/block_contents.yaml
```diff
@@ -1,37 +1,36 @@
...
-    type: object
-    description: "The required signed components of block production according to the Deneb CL spec."
-    properties:
-      signed_blinded_block:
-        $ref: "./block.yaml#/Deneb/SignedBlindedBeaconBlock"
-      signed_blinded_blob_sidecars:
-        $ref: "./blob_sidecar.yaml#/Deneb/SignedBlindedBlobSidecars"
-       
\ No newline at end of file
+      kzg_proofs:
```
g11tech, line 33: similar comment as above

apis/validator/block.v3.yaml
rolfyone: v2 was only created for deneb... if its causing us to keep a container we don't need, thats sub-optimal...

types/deneb/block_contents.yaml
```diff
@@ -1,37 +1,36 @@
 Deneb:
+  KZGProofs:
+    type: array
+    items:
+      $ref: '../primitive.yaml#/KZGProof'
+    minItems: 0
+    maxItems: 6
```
etan-status, line 7: Should use the theoretical max here, `MAX_BLOB_COMMITMENTS_PER_BLOCK`, for stable SSZ transport if the number changes in future

etan-status: Same for `Blobs`

apis/validator/block.v3.yaml
etan-status: theoretically, the VC doesn't even need access to the blobs. it can just sign the block and let the blob be handled by the BN, as in, creating the `BlobSidecar` from `SignedBeaconBlock` + `KzgProofs / Blobs`

types/deneb/block_contents.yaml
etan-status: The validator can compute them on its own, the inclusion proof can be computed solely from `block`.

apis/validator/block.v3.yaml
etan-status: Actually, it needs access, because the BN that they are sent to may not have the blobs anymore, if they were originally fetched from a different one.

types/deneb/block_contents.yaml
etan-status: ref: https://github.com/ethereum/builder-specs/blob/534e4f81276b8346d785ed9aba12c4c74b927ec6/specs/deneb/builder.md#blindedblobsbundle

apis/validator/block.v3.yaml
g11tech: right, thats the idea of statelessness

types/deneb/block_contents.yaml
g11tech: +1 considering ssz transport

types/deneb/block_contents.yaml
g11tech: correct however what i was clarifying that since validator publishes signed block contents, the intercepting beacon is would be building the proofs and publishing blocks  (which is fine or may be good as well that while validator is signing the beacon can build the proofs and patch them in as validator starts publish flow although just a matter of 20-30ms as such)

(this comment is about SignedBlockContents)

g11tech (commented):
i am onboard with the PR with a change that @etan-status proposed to keep the max length to MAX_BLOB_COMM... considering ssz transport encoding

apis/validator/block.v3.yaml
realbigsean: > v2 was only created for deneb... if its causing us to keep a container we don't need, thats sub-optimal...

right I think I missed your point. Agree we should deprecate V2

types/deneb/block_contents.yaml
realbigsean: I was thinking in implementation, a beacon node could start building the inclusion proofs and cache them when it builds a block. But should also have the ability to compute them on the fly (on a cache miss). So in the case where you publish through the BN building the block it'd be faster.

types/deneb/block_contents.yaml
realbigsean: good catch!

realbigsean: updated!

realbigsean:
Ok I think I've addressed the feedback so far

types/deneb/block_contents.yaml
g11tech: right, lodestar will soon be transition to SSZ now lol

apis/validator/block.v3.yaml
marioevz: Just wanted to mention that keeping the blobs in the unblinded flow response makes it really easy to update our testing tools for the blob p2p scenarios in testnets.

rolfyone:
sorry for the less interesting bit, but changelog entry :)

etan-status (approved):
Looks good, my understanding is as follows:

------------

Regular flow

1. BN1 -- BeaconBlockContents --> VC (with blobs + kzg_proofs)
2. VC -- SignedBeaconBlockContents --> BN2 --> libp2p with blobs + kzg_proofs as is, just re-upload)

BN1 may be different from BN2

------------

Blinded flow

1. MEV relay -- BlindedBeaconBlock --> BN --> VC
2. VC -- SignedBlindedBeaconBlock --> BN --> MEV relay
3. MEV relay -- SignedBeaconBlockContents --> BN --> libp2p

mcdee:
@etan-status in your blinded flow part 1, the MEV relay currently provides an execution payload rather than an entire blinded beacon block.

With the advent of deneb this should change to the execution payload plus an array of blinded blob info.  I *think* the blob info should be the blob roots plus the KZG commitments and KZG proofs, but would like to see if others on here concur.

etan-status:
ah, right. but from BN --> VC the executionpayload is then combined with the beacon block, so the the VC it looks the same, regardless.

Update:

1. MEV relay -- executionpayload --> BN -- BlindedBeaconBlock --> VC
2. VC -- SignedBlindedBeaconBlock --> BN --> MEV relay
3. MEV relay -- SignedBeaconBlockContents --> BN --> libp2p

etan-status:
No need for KZG proofs / blob roots. there is no more signature on the blobs.

etan-status:
And, in blinded flow, VC has to use the same MEV relay, so the same BN, to complete the flow. it is only in the unblinded flow where a different BN may be selected for publishing.

mcdee:
> No need for KZG proofs / blob roots. there is no more signature on the blobs.

Okay, so just the KZG commitments to place inside the block body.

etan-status:
the `hash_tree_root(kzg_commitments)` would be sufficient. otherwise, there is not really blinding because the BN can identify what blobs were included by the relay, if they also observed the mempool.

mcdee:
> And, in blinded flow, VC has to use the same MEV relay, so the same BN...

What's the restriction here?  Is the BN now holding state that must be passed to the MEV relay, or is it just the same condition at current that the BN knows which relay to talk to in order to carry out the unblinding?

etan-status:
The latter. The BN knows the identity of the MEV relay who sent the bid, and also maintains the validator registration with that relay.

mcdee:
> the `hash_tree_root(kzg_commitments)` would be sufficient. otherwise, there is not really blinding because the BN can identify what blobs were included by the relay, if they also observed the mempool.

But `BlindedBeaconBlock` has the full KZG commitments in its definition.

Also, blinding is really to avoid leaking bundles that could be picked apart, so not sure that knowing which blob transactions are in a block is an issue.

etan-status:
Sure, just saying, the protocol would work perfectly well even if `BlindedBeaconBlock` only contained `hash_tree_root(kzg_commitments)`. If blinding doesn't try to hide the blob transactions, that's fine (but also kinda pointless) to include the full `kzg_commitments` in `BlindedBeaconBlock`.

mcdee (approved):

rolfyone (approved):
LGTM

PR #370, opened 2023-11-03
Update Deneb for `v1.4.0-beta.4`

etan-status:
`BlobSidecar` are no longer signed, but instead are linked with the signed block via merkle proof. Also renamed `signed_blob_sidecars` and `signed_blinded_blob_sidecars` in `BlockContents` envelopes to match the new types (without signature).

- https://github.com/ethereum/consensus-specs/pull/3531


etan-status:
Builder API may change: 

- https://github.com/ethereum/builder-specs/pull/90

etan-status:
Alternative that uses `proofs` and `blobs` lists instead of `blobs_bundle`:

- https://github.com/ethereum/beacon-APIs/pull/369

PR #372, opened 2023-11-16
Clarify broadcast validation failure in response code for publishBlockV2

zilm13:
It's mentioned in description that in case of broadcast validation failure we should return 400, but error codes description like "The `SignedBlindedBeaconBlock` object is invalid" confuses a bit. I've changed it to be more clear


tbenr (approved):
LGTM

michaelsproul (approved):
Nice catch!

Issue #375, opened 2023-11-29
Add validator slashing information to event stream 


mehdi-aouadi:
The slashing events could help the validator determine whether at least one of its keys has been slashed and let it act accordingly (suspend VC requests, shut down, raise warnings...). This could be particularly useful to prevent massive slashing events when a VC runs multiple keys.
The massive slashing event that happened on November 14th 2023, at epoch [242546](https://beaconcha.in/epoch/242546) is a good example. It caused 99 validators to be slashed (99 ETH loss approximately). They were all newly spun up validators and this could have been prevented by stopping the VC which is running all the keys after the first slashing event.
This was initially raised by mariano.eth

The suggested event would be: 
- Event: `slashed_validator`
- Data: `{"pubkey": "0x...."}`

I'd like to have your opinion on this before raising a spec PR.

mcdee:
There are already tools to track slashing events based on the existing beacon chain APIs e.g. https://github.com/attestantio/esd  Not sure that having a separate event would make a significant difference in terms of ability to deploy such tools.

mehdi-aouadi:
> There are already tools to track slashing events based on the existing beacon chain APIs e.g. https://github.com/attestantio/esd Not sure that having a separate event would make a significant difference in terms of ability to deploy such tools.

Thanks for the heads-up I wasn't aware such tool exists. I was more thinking of a built in feature (that could complement the doppelganger detection). The difference here would be the ease of use plus IMO adding a separate event wouldn't do any harm

mehdi-aouadi:
> To an extent, I'm not sure how much the protocol should effectively subsidize these staking pools (BTCS in this case) which know 100% perfectly well the risk they take. Mass slashing events happen precisely when staking pools, or similarly typically professional Ethereum entities, become overconfident in their key and signing management and ability to juggle loading keys across multiple machines.
> 
> I'm not going to strongly oppose this as such; it's in isolation harmless and defensible, as is e.g., [prysmaticlabs/prysm#13194](https://github.com/prysmaticlabs/prysm/issues/13194) or [Consensys/teku#7707](https://github.com/Consensys/teku/issues/7707). However, it's an acceptance of allowing BTCS and its ilk to externalize effectively a cost of business on the ecosystem as a whole.
> 
> This slashing did not particularly hurt Ethereum as a whole. It hurt one particular node operator, and that's the exact gamble they took, and lost.
> 
> Longer-term, DVT will provide a better solution to this, and largely obviate this issue.

[...]

PR #376, opened 2023-12-01
Add proposer_slashing and attester_slashing to SSE

dapplion:
Add proposer_slashing and attester_slashing to SSE. There's a non-zero delay between slashable messages seen on gossip, and them being included on-chain. These new events allows consumers to learn about slashable events faster than just tracking blocks, specially in the case of large amounts of slashable messages.

### One relevant use-case

The total cost of a slashing event for large operators is highly dependant on their response time. If many keys are involved in an operational error it can take some time for those slashable messages to be included on chain.

There are monitoring tools available that stop your setup in response to slashable messages seen on-chain. For example https://github.com/attestantio/esd. Their response time could be minimized to zero slots by watching gossip directly.


mehdi-aouadi:
I think it would be better and simpler to consolidate the two events into a simple one:
- Event: `slashed_validator`
- Data: `{"pubkey": "0x...."}`

That would be sufficient for this purpose. I don't see the point of having all the current attester and proposer slashing messages payload, plus it's more usefull for the VC to have the public keys instead of the index (let the BN do the translation) as mentioned by @rolfyone [here](https://github.com/ethereum/beacon-APIs/issues/375#issuecomment-1832809778)

dapplion:
> I would have thought it'd be less work all round if there was just a stream of pubkeys or pubkeys and indices, rather than publishing the whole event for the 1 field... This would satisfy what's been asked for, just wondering if we're ever going to use the rest of that data... It's admittedly nice to just publish standard spec events, I do see that side of things...

Right the intent of this specification was to expose this data in the most generic way possible, instead of satisfy a very specific demand. However if we are confident that the full events will always be useless, just emitting pubkeys is more succint. Validators should already have a internal cached map of index -> pubkey, plus dealing with a few extra kB of data over the wire isn't that bad.

Do you see a big downside in complexity for having to parse the full events? Sending the full event also opens the possibility for trustless verification on the validator client side

g11tech:
> I'd be inclined to send the full spec struct, probably simpler in codebases to do this than creating a new struct with a subset of the data, and also sets us up if we ever move towards sending events as SSZ.

+1 

also its better to have it as separate events

g11tech (approved):

rkapka (approved):

rolfyone:
Lets add these to the change doc:)

james-prysm:
Biggest concern we have that still needs investigating is if this would clog up our event notifier based on its current design. may be a bigger task to add if its the case.

rolfyone:
> Biggest concern we have (prysm side) that still needs investigating is if this would clog up our event notifier based on its current design. may be a bigger task to add if its the case.

if we have that many slashings that this becomes an issue, that would be a concern in itself...

rolfyone (approved):
LGTM

PR #379, opened 2023-12-07
fix: typo in JSON payload

jeluard:
A trailing `,` prevents the payload to be valid JSON.
This is problematic for implementations relying on this data to perform automated validation tests.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -122,7 +122,7 @@ get:
                 ignored by block builders and most other API consumers.
               value: |
                 event: payload_attributes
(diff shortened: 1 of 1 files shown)


jeluard:
Thanks! Any chance this fix can be backported to existing releases?

rolfyone:
> Thanks! Any chance this fix can be backported to existing releases?

We'd have to manually hack the json from those releases... is it causing actual problems?

michaelsproul:
I feel like a new release would be a cleaner soln

rolfyone:
Can be done, ideally we can fix the outstanding PRs with v3 block before we do a release though.

jeluard:
Sure, we can wait for a new release.

Issue #109, opened 2020-10-26
Response code when aggregate can't be created


ajsutton:
`/eth/v1/validator/aggregate_attestation` may not be able to create an aggregate if the beacon node doesn't have any attestations with the specified attestation data root.  In that case, currently both Teku and Lighthouse are returning a 404 response code but it's not actually listed as one of the possible response codes in the OpenAPI docs.

Also, while 404 isn't an unreasonable option, it has led to a bunch of confusion about whether the URL being requested was wrong (typically 4xx response codes mean the caller did something wrong). In this case the call was valid but there was nothing to return.  I wonder if 204 No Content would be a better response code.

In any case, the OpenAPI docs should be updated to explicitly list what response code is used when no known attestations matching that data root are found.

mpetrunic:
We could return 400 with some error code/description

dapplion:
Not sure if this is still an issue? Lodestar returns 500 in the attestation data is not known.

The caller has agency to request an attestation data that it previously produced (correct spec), or request a random attestation data root. In the second case, it's the callers fault by requesting bad data. However I see @ajsutton point internal conditions of the beacon node can cause the aggregate to not be producible so should be a 500 code.

ajsutton:
500 is an internal server error so I wouldn't think that's appropriate for the case to just not having data to return. 500 response indicates a bug or similar fault in the beacon node.

It's also entirely within spec to request an aggregate from a beacon node based on data that you haven't previously sent to that specific node. You might have published the attestation via one node and then requested the aggregate from a different one (for any number of reasons).

dapplion:
Sometimes I wish we had a JSON RPC so we don't need to do semantical debates to choose a random status number lmao

ajsutton:
JSON-RPC has error codes too and they are even more poorly defined and lack standardisation. :) 

I'd say either 404 or 204 are the appropriate responses. 404 is more technically accurate given that the request for an aggregate matching the specified data and nothing matching that request was found.  Assuming the majority of clients are already returning 404 I'd just add that to the spec and be done with it. It's just unfortunate that 404 could mean "you typed the URL wrong" or "valid request but I don't have any data" but that's just the way the HTTP spec goes.

mcdee:
I think that 204 is more technically accurate in this case.  If the parameters were in the path a 404 would be reasonable, but the endpoint does indeed exist there is just no response to provide.

[...]

PR #380, opened 2023-12-08
Add note on no attestation_data_root

dapplion:
- Address a really old concern raised by https://github.com/ethereum/beacon-APIs/issues/109 where some implementations returned 404 if no aggregate can be built.


apis/validator/aggregate_attestation.yaml
```diff
@@ -7,6 +7,9 @@ get:
     A 503 error must be returned if the block identified by the response
     `beacon_block_root` is optimistic (i.e. the aggregated attestation attests
     to a block that has not been fully verified by an execution engine).
+
+    A 500 error must be returned if no attestation is available for the requested
+    `attestation_data_root`.
```
rolfyone, line 12: a 500 would be super weird, why are we going with that?

dapplion: The sane alternative is a 404, which is not ideal either.

rolfyone: It's more difinitive, and consistent with what we do when we have routes not found on the states endpoint... I'm happy with that.

rolfyone (approved):
LGTM

Issue #186, opened 2022-01-18
No endpoint to get sync committee contributions from pool


michaelsproul:
I noticed while browsing the API spec that there is no `GET /eth/v1/beacon/pool/sync_committees` to match the [POST](https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolSyncCommitteeSignatures) endpoint. For attestations and all other types of messages we define `GET` as well as `POST`.

For completeness I think it would be nice to add the sync committee endpoint

staccDOTsol:
Thank you for bringing this to my attention. I understand your concern regarding the lack of a `GET /eth/v1/beacon/pool/sync_committees` endpoint to retrieve sync committee contributions from the pool. While the current API specification does not include this specific endpoint, it is worth noting that the Ethereum Beacon Chain API is still under active development and subject to updates.

To ensure completeness and consistency, it would indeed be beneficial to have a corresponding `GET` endpoint for sync committee contributions. I will make a note of this suggestion and pass it along to the appropriate teams for consideration in future updates to the API specification.

In the meantime, if you have any further questions or need assistance with any other Ethereum-related topics, feel free to ask!

michaelsproul:
@staccDOTsol get out of here with your GPT word-salad mate

PR #381, opened 2023-12-08
Add getPoolSyncCommitteeSignatures route

dapplion:
- https://github.com/ethereum/beacon-APIs/issues/186 noted missing route to retrieve SyncCommitteeSignature from the pool

There hasn't been much demand for it, but there's no good reason to omit this route on purpose so adding for completeness

Closes https://github.com/ethereum/beacon-APIs/issues/186


rolfyone:
There doesn't seem to be much interest in adding this without a use-case..

tersec:
"no good reason to omit" isn't a reason to add. Oppose.

rolfyone:
closing, as no real takers for wanting this.

Issue #322, opened 2023-06-01
Change example endpoint


ralexstokes:
we should change this to something that is clearly just an example as many people expect there to be some hosted node behind this endpoint...

https://github.com/ethereum/beacon-APIs/blob/2a15cc71cf8473e1013713de98d0aacf8bf2ea05/beacon-node-oapi.yaml#L37

staccDOTsol:
I understand your concern about the example endpoint in the beacon-node-oapi.yaml file. It's important to clarify that the provided endpoint is indeed just an example and not a real hosted node. This is to demonstrate the structure and format of an API endpoint for educational purposes.

To avoid any confusion, we can modify the example endpoint to make it explicitly clear that it is not a functional endpoint. This will help users understand that they need to replace it with their own valid endpoint when implementing the API.

Thank you for bringing this to my attention, and I will make sure to update the example accordingly.

rolfyone:
If we used relatively common api ports we could default to localhost, but I don't think there's any kind of standard...
```
Teku: 5051
Lighthouse: 5052
prysm: 3500
nimbus: 5052
lodestar: 9596
```

We could potentially set the default to something like `http://localhost/` which will fail in the same way the current one does... 

It might be easier to just raise a PR with a suggestion that people can throw stones at?

PR #383, opened 2023-12-08
Use obvious example URL

dapplion:
Setting http://public-mainnet-node.ethereum.org/ as default is very deceiving as one could believe there's a node deployed there. Note that many networks deploy infra in there base URLs.

Setting to `http://localhost/` as a clear neutral URL that will break for sure. Not setting any specific port since every client uses a different one

Closes https://github.com/ethereum/beacon-APIs/issues/322


rolfyone (approved):
LGTM at least it wont try to resolve it etc...

Issue #366, opened 2023-10-21
Inconsistency in `/eth/v1/node/peers`


mcdee:
The `/eth/v1/node/peers` returns a `meta` element that contains a `count` item.  The schema for this item declares the value to be a number, which is inconsistent with the common format of providing numeric values as quoted strings.

A quick tour of the main beacon nodes shows the following for this value:

- lighthouse: returns the value unquoted
- nimbus: returns the value quoted
- prysm: does not return the `meta` element
- teku: returns the value unquoted

Options that I can think of:

1. patch the spec so that the `v1` endpoint returns a quoted value
2. raise bugs with the relevant beacon nodes to ensure this value is returned unquoted
3. deprecate the `v1` endpoint and create a `v2` endpoint with a quoted value
    1. retaining the `meta` element
    2. discarding the `meta` element (nowhere else do we return a value like this that can easily be calculated from the other returned information)
    3. moving the count value to the top level returned object where all other metadata lives

My inclination is for option 3 suboption 2, but I would be interested in hearing if anyone has a reason for this `meta` element to exist, and the `count` value within it.

rolfyone:
i wouldn't be against a v2 without the count - its pretty trivial to get the count from an array...

I guess that translates to 3 suboption 2 being my preference.

nflaig:
The `meta` element might have been a good idea if used consistently across all APIs but as it is right now, any top-level property should be considered metadata (`execution_optimistic`, `finalized`, etc.) and everything else (the actual data) must be part of `data` element.

As this API is an outlier and does not follow this pattern, I would also prefer option 3 suboption 2.

dapplion:
+1 to burn the meta

dapplion:
No one likes the meta element, but there's no appetite either for the effort to remove it 

- See discussion in https://github.com/ethereum/beacon-APIs/pull/384#issuecomment-1850805471

According to @mcdee review of clients seems that support of the meta element is poor. If no-one has complained should mean that no consumer is dependent on it.

To address this one-off oddity we could do a breaking change on route v1 and mark the meta property as optional. Then only Nimbus needs to fix the route to be compliant, which @arnetheduck is already keen on doing.

Thoughts?

PR #384, opened 2023-12-08
Add peers v2 without meta

dapplion:
- Suggested in https://github.com/ethereum/beacon-APIs/issues/366

Any consumer can derive the number of items by just doing `len(data)`. Creating a new route as a property is deleted from the response object

Closes https://github.com/ethereum/beacon-APIs/issues/366


rolfyone (commented):
we can probably deprecate the v1 peers if we do add this, and should add to changes list :)

arnetheduck:
This feels like a lot of bother for a small aesthetic nuisance in a fringe interface..

dapplion:
> This feels like a lot of bother for a small aesthetic nuisance in a fringe interface..

Also true, taking the counter argument there's really no harm in sticking with this oddity that already handled by all implementations.

nflaig:
It is an outlier and hence needs some special handling during parsing if you implement a generic http client/server. It's a minor thing but probably not wrong to get rid of technical debt.

rolfyone:
Sounds like maybe we close the issue as 'won't fix'?

Issue #378, opened 2023-12-07
Missing 404 for `/eth/v1/beacon/states/{state_id}/validator_balances` and similar


michaelsproul:
While implementing the new POST validators/validator_balances endpoints I realised that the endpoint specs are missing 404 responses for the case where the state is missing/not available.

Usually we signal that the state isn't known with 404, although I guess we could also return a 400?

It seems more like an oversight (on both GET and POST) to me.

PR #385, opened 2023-12-08
Add 404 response to states validator_balances

dapplion:
states validator_balances routes are missing the 404 response status that every other states/* route has to denote unknown state_id

- Noted in https://github.com/ethereum/beacon-APIs/issues/378

Closes https://github.com/ethereum/beacon-APIs/issues/378


rolfyone (approved):
LGTM

PR #386, opened 2023-12-11
Specify `builder_comparison_factor`

michaelsproul:
Generalisation and replacement for #377 based on the rough consensus achieved in the discussion there.

One new parameter is added for block building which can be used to encode several types of validator preferences. There is a trade-off between simplicity and expressiveness, and IMO this single parameter achieves a good middle ground.

I haven't done any renames of blinded -> builder in this PR as that seems like a separate issue, and the current spec is actually already unambiguous on this matter when it states:

> The beacon node must return an unblinded block if it obtains the execution payload from its paired execution node. It must only return a blinded block if it obtains the execution payload header from an MEV relay.


apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,38 @@ get:
...
+
+        * `builder_comparison_factor=0`: prefer the local payload unless an error makes it
+           unviable.
+        * `builder_comparison_factor=100`: default profit maximization mode; choose whichever
+          payload pays more.
+        * `builder_comparison_factor=2**64 - 1`: prefer the builder payload unless an error or
+        beacon node health check makes it unviable.
+
+        Servers should use saturating arithmetic or another technique to ensure that large values
+        of the `builder_comparison_factor` do not trigger overflows or errors.
```
rolfyone, line 71: It needs to be a natural number between 0-MAX_UINT64 inclusive, and if its not in that range we should give a 400 response?
Also, if it's not specified at all, then some kind of defaulting would be presumed, is there any guidance needed for consistency?

michaelsproul: Updated.

g11tech (approved):
lgtm

mcdee:
> Just a nit, but I prefer comparisons to be in addition to instead of relative. That is, instead of 120, I rather see the comparison be 20, as in 20% more than the local payload instead of 1.2 times the local payload

I don't see how you could request a locally-built block with the additive style configuration, given that the parameter is unsigned.  Relative seems more flexible.

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
...
+      required: false
+      description: |
+        Percentage multiplier to apply to the builder's payload value when choosing between a
+        builder payload header and a local payload. This parameter is only relevant if the beacon
+        node is connected to a builder, deems it safe to produce a builder payload, and receives
+        valid responses from both the builder endpoint _and_ the local execution node. When these
+        preconditions are met, the server MUST act as follows:
+
+        * if `local_payload_value >= builder_comparison_factor * builder_payload_value // 100`, then
+          return a full (unblinded) block containing the local payload.
```
tersec, line 57: `local` gets used in many ways -- my understanding is that the precise distinction is builder API or engine API?

It might not be local as such, even for engine API.

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
...
+
+        Servers must support the following values of the comparison factor which encode common
+        preferences:
+
+        * `builder_comparison_factor=0`: prefer the local payload unless an error makes it
+           unviable.
+        * `builder_comparison_factor=100`: default profit maximization mode; choose whichever
+          payload pays more.
+        * `builder_comparison_factor=2**64 - 1`: prefer the builder payload unless an error or
+        beacon node health check makes it unviable.
```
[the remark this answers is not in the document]
tersec: I'm not really opposed to this, but how is this this an improvement?

Also, if nothing else, the grammar's wonky now: "beacon node health check" lacks an article or other similarly-functioning word. Before, it shared `an` with `error`, but this removes that linkage.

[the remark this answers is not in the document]
dapplion: Why not just 
```
        * `builder_comparison_factor=2**64 - 1`: prefer the builder payload unless an error makes it unviable.
```
@michaelsproul why is a "beacon node health check" relevant here?

apis/validator/block.v3.yaml
michaelsproul: good point, clarified in e5b1045ecf5ed474a38ac857937dc593efee269c

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
...
+      in: query
+      required: false
+      description: |
+        Percentage multiplier to apply to the builder's payload value when choosing between a
+        builder payload header and a local payload. This parameter is only relevant if the beacon
+        node is connected to a builder, deems it safe to produce a builder payload, and receives
+        valid responses from both the builder endpoint _and_ the local execution node. When these
+        preconditions are met, the server MUST act as follows:
+
+        * if `local_payload_value >= builder_comparison_factor * builder_payload_value // 100`, then
```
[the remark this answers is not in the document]
michaelsproul: Yeah this is nicer, thanks! Updated in e5b1045ecf5ed474a38ac857937dc593efee269c.

michaelsproul:
> Locally built would be max uint. What's impossible is to request a builders block unconditionally which is part of the reason I like this better

@potuz The point of the multiplicative factor is to make a large variety of preferences expressible using a single param. Removing the ability to define the "prefer builder" case would be antithetical to that. A "builder" block also doesn't necessarily imply censorship. In a DVT setup you could e.g. sit a piece of software that uses the builder API in front of a local EL and share the blocks it produces with the entire cluster so that they achieve consensus on it.

rolfyone:
Is everyone happy with this now?

rolfyone (approved):
LGTM

rkapka (approved):

dapplion (approved):

tersec (approved):

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
         is a flag and does not take a value.
       schema: {}
       allowEmptyValue: true
+    - name: builder_comparison_factor
```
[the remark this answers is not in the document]
rolfyone: Given thats the whole point of the PR, i can't just merge that without letting people have a say so I'll have to wait for @michaelsproul to take a look.

rolfyone (approved):
LGTM

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
...
+
+        * if `exec_node_payload_value >= builder_comparison_factor * (builder_payload_value // 100)`,
+          then return a full (unblinded) block containing the execution node payload.
+        * otherwise, return a blinded block containing the builder payload header.
+
+        Servers must support the following values of the comparison factor which encode common
+        preferences:
+
+        * `builder_comparison_factor=0`: prefer the execution node payload unless an error makes it
+          unviable.
```
potuz, line 64: What does this mean? if the beacon health check makes it unviable, should we return the local payload or error out?

apis/validator/block.v3.yaml
```diff
@@ -43,6 +43,39 @@ get:
...
+        * `builder_comparison_factor=100`: default profit maximization mode; choose whichever
+          payload pays more.
+        * `builder_comparison_factor=2**64 - 1`: prefer the builder payload unless an error or
```
potuz, line 67: What does this mean? if the beacon health check makes it unviable, should we return the local payload or error out?

michaelsproul: (as discussed on Discord) local payload

PR #390, opened 2023-12-12
fix: remove deprecated project reference

jeluard:
Referenced project from ChainSafe has not been maintained since a couple years and is now relevant anymore.


rolfyone (approved):
Sounds fair - thanks for raising the PR.

PR #392, opened 2023-12-14
Fix broken link in deployed doc

jeluard:
The [Checkout validator flow](https://ethereum.github.io/beacon-APIs/validator-flow.md) link is broken on the deployed [doc](https://ethereum.github.io/beacon-APIs/#/).
This is due to a file not being copied to the right folder during the deployment phase.

Note that this will only impact the deploy procedure. The right file should be copied on the `gh-pages` branch to fix on the live website.


rolfyone (approved):
LGTM

PR #398, opened 2023-12-21
chore: fix references of builder's comparision factor to boost factor

g11tech:
the renaming of the `builder_comparision_factor` to `builder_boost_factor` was accepted in https://github.com/ethereum/beacon-APIs/pull/386 but all references weren't updated.

this PR just updates the references


rolfyone (approved):
tyvm, i had assumed the rename was complete!

PR #399, opened 2024-01-02
Add required field to object and request body definitions

nflaig:
**Motivation**

As mentioned in https://github.com/ethereum/beacon-APIs/pull/388#issuecomment-1864843080 I think it would be good practice to add the `required` field to object and request body definitions.

This main reasons for this are
- openapi spec compliance / technical correctness
- detect missing properties during schema validation by using [ajv](https://github.com/ajv-validator/ajv) or similar tooling
- allows to generate correct types via [openapi-generator](https://github.com/OpenAPITools/openapi-generator), properties are no longer optional (type safety)
- clarity on how to make properties optional (https://github.com/ethereum/beacon-APIs/pull/388)

**Description**
- Add required properties to object types
- Add required field to request bodies

### DIFF
--- a/apis/beacon/blob_sidecars/blob_sidecars.yaml
+++ b/apis/beacon/blob_sidecars/blob_sidecars.yaml
@@ -31,6 +31,7 @@ get:
           schema:
             title: GetBlobSidecarsResponse
             type: object
+            required: [data]
             properties:
               data:
                 $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Deneb.BlobSidecars"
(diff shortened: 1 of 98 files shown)


dapplion (approved):
OpenAPI spec clearly states that all properties of objects are optional by default (https://swagger.io/docs/specification/data-models/data-types/#object). All clients and consumers assume otherwise so we should align the spec with reality.

Reviewed all changes (long PR..) and are correct 👍

jeluard:
I could also confirm that this PR does the following:
* explicit all properties of all objects and types as `required` (openapi defaults to all properties optional)
* mark all `requestBody` as required (openapi defaults to not required)

Also note that `required` is already correctly set for `parameters` (either true or false) and `required: false` was already set for some optional `requestBody` (the default, but better be explicit probably).

Nice job @nflaig ! Great step towards improving API interoperability and making sure testing can be automated.

PR #400, opened 2024-01-08
Clarify that payload & block values are in Wei

### LINKED ISSUE #397: gwei vs wei in block v3
I found a discrepancy between Teku and Lighthouse's implementation of the block v3 endpoint. Lighthouse returns the consensus block value in gwei and the execution payload value in wei. Teku uses wei for both.

Arguably both are reasonable interpretations of the current spec, because we _usually_ use gwei on the consensus layer (we can't represent sub-gwei units on the CL). On the other hand, using wei for both is nice from a consistency point of view.

We should choose between gwei+wei, or wei everywhere and then update the descriptions of the block v3 headers accordingly. The current descriptions leave a lot to be desired:

> Eth-Execution-Payload-Value: Required in response so client can determine relative value of execution payloads.
> Eth-Consensus-Block-Value: Required in response so client can determine relative value of consensus blocks.

michaelsproul:
Require both execution payload value and consensus block value to be denominated in _wei_.

There are advantages to wei+gwei or wei+wei, and we just need to choose one. I've chosen to use the wei+wei option, because the consistency makes the API less confusing, and I think this trumps concerns about performance or unnecessary zero padding.

Closes #397

### DIFF
--- a/beacon-node-oapi.yaml
+++ b/beacon-node-oapi.yaml
@@ -420,12 +420,16 @@ components:
       schema:
         type: boolean
     Eth-Execution-Payload-Value:
-      description: Required in response so client can determine relative value of execution payloads.
+      description: |
+        Execution payload value in Wei. Required in response so client can determine relative value
+        of execution payloads.
       required: true
       schema:
         $ref: './types/primitive.yaml#/Wei'
     Eth-Consensus-Block-Value:
-      description: Required in response so client can determine relative value of consensus blocks.
+      description: |
+        Consensus rewards paid to the proposer for this block, in Wei. Required in response so
+        client can determine relative value of consensus blocks.
       required: true
       schema:
         $ref: './types/primitive.yaml#/Wei'


dapplion (approved):

michaelsproul:
Relatedly, the current spec uses decimal representation for both values. In Lighthouse were were using hex for the U256 and decimal for the u64. If everyone is OK with decimal I'll go ahead and fix our impl.

nflaig:
The values in the rewards API are denominated in _gwei_. Shouldn't we use the same unit across all APIs?

https://github.com/ethereum/beacon-APIs/blob/cfbe6677f6e9e936dfca8708467c7e0382c07b19/types/rewards.yaml#L48

mcdee:
This API has wei-level values.  I don't think that it makes sense to update everywhere, but it does make sense to be consistent within a single endpoint so all wei values here (as execution layer values are in wei precision).

nflaig (approved):
It would simplify the implementation, for example the rewards cache in Lodestar uses _gwei_ and we use it for both rewards API and payload / block value. This means an extra value conversion is required here which is error prone compared to just dealing with the same unit across the whole codebase.

I don't see an ideal solution here as breaking the existing rewards API is not great either, and should probably be a separate discussion.

Using _wei_ seems to align with execution layer values definitely in favor of using that here.

g11tech (approved):
lgtm

PR #401, opened 2024-01-08
Quote uint64 and int64 examples

dapplion:
Both uint64 and int64 are defined are quoted numbers, fix examples to align with their type.

https://github.com/ethereum/beacon-APIs/blob/4882aa0803b622b75bab286b285599d70b7a2429/types/primitive.yaml#L32-L34

https://github.com/ethereum/beacon-APIs/blob/4882aa0803b622b75bab286b285599d70b7a2429/types/primitive.yaml#L40-L42

For reference Lighthouse stable quotes all integers in the rewards API

https://github.com/sigp/lighthouse/blob/441fc1691b69f9edc4bbdc6665f3efab16265c9b/common/eth2/src/lighthouse/attestation_rewards.rs#L8-L27

### DIFF
--- a/types/rewards.yaml
+++ b/types/rewards.yaml
@@ -9,12 +9,12 @@ SyncCommitteeRewards:
       validator_index:
         allOf:
           - $ref: "./primitive.yaml#/Uint64"
-          - example: 0
+          - example: "0"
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM

Issue #349, opened 2023-09-11
SSE block event post deneb


dapplion:
The SSE block event is used currently to measure block propagation times. This data is critical to inform decisions like blob count and overall network health. In consensus clients blocks go through a few phases:

1. Arrival on gossip, before gossip validation
2. Completed gossip validation
3. Block imported (complete gossip, consensus, execution, and DA) validation
4. Block set as head

I believe both 2. and 3. are equally important event that convey different information. The API spec is vague such that you could interpret as having to emit the event on either 2. or 3.

https://github.com/ethereum/beacon-APIs/blob/8b52d949965d2294d3f0191c6b15644535a933fe/apis/eventstream/index.yaml#L50-L51

I propose to have two separate events properly defined:

- `block`: The node has imported a valid block into its fork-choice, received from P2P or API
- `block_gossip`: The node received a block from gossip that pass gossip validity conditions of the `beacon_block` topic

CC: @parithosh

mcdee:
What's the gap between 3) and 4), in general?  Can they be far enough apart that they warrant separate events?

I'd prefer to keep the `block` event as at item 2, as one of its prime uses is to track propagation of blocks through the network.

3 and 4 could be argued as both being part of the 'import to state' phase of the process, is there a requirement out there to know the timing of item 3?

dapplion:
> What's the gap between 3) and 4), in general? Can they be far enough apart that they warrant separate events?

Not all blocks end up being head, so for the same block hash event 4 may never be emitted. Also a block could be made head latter when there's enough LMD weight, so the gap between 3 and 4 ranges from 0 to infinity.

> I'd prefer to keep the block event as at item 2, as one of its prime uses is to track propagation of blocks through the network.

We would have to check with all consumers of this event, because emitting at 2 means you can potentially emit invalid blocks. Not sure if this can break explorers and other uses. As of today, at least Lighthouse and Lodestar emit the `block` event at step 3.

> 3 and 4 could be argued as both being part of the 'import to state' phase of the process, is there a requirement out there to know the timing of item 3?

Not a strong requirement in timing in general, but they convey different information: 2. is precise in timing but may emit invalid blocks. 3. is imprecise in timing but is guaranteed to emit valid blocks.

mcdee:
> Not a strong requirement in timing in general, but they convey different information: 2. is precise in timing but may emit invalid blocks. 3. is imprecise in timing but is guaranteed to emit valid blocks.

[...]

PR #403, opened 2024-01-09
Specify 'valid' gossip condition of SSE topics

dapplion:
Related to https://github.com/ethereum/beacon-APIs/issues/349, clarify what 'valid' means in the definition of all SSE topics except the block topic


rolfyone (approved):
LGTM

PR #404, opened 2024-01-09
Specify 'valid' condition for block SSE event

dapplion:
From discussion in https://github.com/ethereum/beacon-APIs/issues/349, specify to emit the `block` event only for blocks successfully imported on fork-choice. This condition includes: valid state transition (super-set of gossip rules) + valid/syncing execution + DA checks


mcdee:
I'm unsure of the point of the `block` event with this change.  Won't it be emitted at basically the same time as the `head` event?

(The only time we wouldn't have overlap would be if the imported block wasn't made head, but not sure of the use case in that situation, either.)

dapplion:
> Won't it be emitted at basically the same time as the head event?

Not necessarily, not all newly imported blocks become head and re-orgs cause previously imported blocks to become head. The head event is very relevant on its own and orthogonal to the block event.

The rules of this PR align with what Lodestar and Lighthouse are doing right now. I haven't look into other implementations but I suspect it will be similar.

rkapka:
Prysm does the same

rolfyone:
I'm happy to clarify that wording, I guess my one caveat remains this is still not a great way of measuring anything time sensitive given these messages may be the lowest priority to action on a node, and the latency from that prioritisation etc...

Within the scope of whats proposed, I think this is a good clarification.

rolfyone (approved):
LGTM

rolfyone:
@michaelsproul feedback?

michaelsproul (approved):
LGTM

PR #406, opened 2024-01-10
Make sure version matches

jeluard:
Light client has been introduced with `altair`. Make sure the relevant event examples reflect that.


rolfyone (approved):
LGTM - just examples, will treat as a chore.

PR #407, opened 2024-01-11
fix: remove unneeded allOf usages

jeluard:
A number of `Schema` and `Response` `$ref`s embed `description` and `example` in a way that is not supported by the openapi spec (relying on an incorrect `allOf` indirection).

Starting with [OpenAPI 3.1](https://github.com/OAI/OpenAPI-Specification/pull/2181) `description` can be overridden at the `$ref` level.
`example` can be defined at the `Response` level, fixing most issues. Some others were duplicates or irrelevant.

Note that some components were already leveraging this syntax, so this PR merely aligns on a single syntax. This also reduces the validation and error messages complexity.

I also validated that `example` still show up in the swagger UI after this change.

requires https://github.com/ethereum/beacon-APIs/pull/409

### DIFF
--- a/apis/node/syncing.yaml
+++ b/apis/node/syncing.yaml
@@ -20,24 +20,19 @@
                   required: [head_slot, sync_distance, is_syncing, is_optimistic, el_offline]
                   properties:
                     head_slot:
-                      allOf:
-                        - description: "Head slot node is trying to reach"
-                        - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+                      description: "Head slot node is trying to reach"
+                      $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
                     sync_distance:
-                      allOf:
-                        - description: "How many slots node needs to process to reach head. 0 if synced."
-                        - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+                      description: "How many slots node needs to process to reach head. 0 if synced."
+                      $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
                     is_syncing:
-                      allOf:
-                        - type: boolean
-                        - description: "Set to true if the node is syncing, false if the node is synced."
+                      type: boolean
+                      description: "Set to true if the node is syncing, false if the node is synced."
                     is_optimistic:
-                      allOf:
-                        - type: boolean
-                        - description: "Set to true if the node is optimistically tracking head."
+                      type: boolean
+                      description: "Set to true if the node is optimistically tracking head."
                     el_offline:
-                      allOf:
-                        - type: boolean
-                        - description: "Set to true if the execution client is offline."
+                      type: boolean
+                      description: "Set to true if the execution client is offline."
       "500":
         $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
--- a/types/altair/sync_committee.yaml
+++ b/types/altair/sync_committee.yaml
@@ -6,8 +6,7 @@ Altair:
       pubkeys:
         type: array
         items:
-          allOf:
-            - $ref: '../primitive.yaml#/Pubkey'
+          $ref: '../primitive.yaml#/Pubkey'
         minItems: 512
         maxItems: 512
       aggregate_pubkey:
@@ -33,12 +32,10 @@ Altair:
       sync_committee_indices:
         type: array
         items:
-          allOf:
-           - $ref: '../primitive.yaml#/Uint64'
+           $ref: '../primitive.yaml#/Uint64'
       until_epoch:
-        allOf:
-         - $ref: '../primitive.yaml#/Uint64'
-         - description: 'The final epoch (exclusive value) that the specified validator requires the subscription for.'
+        $ref: '../primitive.yaml#/Uint64'
+        description: 'The final epoch (exclusive value) that the specified validator requires the subscription for.'
          
   SignedContributionAndProof:
     type: object
@@ -54,9 +51,8 @@ Altair:
     required: [aggregator_index, selection_proof, contribution]
     properties:
       aggregator_index:
-        allOf:
-          - $ref: "../primitive.yaml#/Uint64"
-          - description: "Index of validator in validator registry."
+        $ref: "../primitive.yaml#/Uint64"
+        description: "Index of validator in validator registry."
       selection_proof:
         $ref: '../primitive.yaml#/Signature'
       contribution:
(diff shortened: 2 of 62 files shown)


jeluard:
Note that although validators are fine with thus syntax, strictly speaking `description` can be overridden in a `Reference Object` only starting OpenAPI 3.1.

Similarly `example` is not part of the `Reference Object`. So we are left with 2 options:
* switch all usages of `example` associated with the `Reference Object` to the previous `allOf` hack
* have `example` be part of the referenced `component` definition and get rid of per `$ref` examples
My preference would go to the second option, as it's unclear to me if such usages of `allOf` is sound.

jeluard:
I fixed the `example` by:
* pulling `example` one level up in `Response`s (the right location, some existing Responses were already doing that)
* removing some irrelevant ones (duplicates, dummy values, random enum)
* * leverage changes from https://github.com/ethereum/beacon-APIs/pull/408

Everything should be good now.

types/rewards.yaml
```diff
@@ -59,27 +59,21 @@ IdealAttestationRewards:
...
       description: "Ideal attester's reward for source vote in gwei"
-      example: 5000
     inclusion_delay:
       $ref: "./primitive.yaml#/Uint64"
       description: "Ideal attester's inclusion_delay reward in gwei (phase0 only)"
-      example: 5000
     inactivity:
       $ref: "./primitive.yaml#/Int64"
       description: "Ideal attester's inactivity penalty in gwei"
-      example: 5000
```
dapplion: Examples are important for the swagger UI render. Please rebase on top of merged https://github.com/ethereum/beacon-APIs/pull/401

nflaig (commented):
To make this easier to review and merge I think it should be split into three separate PRs
- update openapi version (this might require separate discussion)
- remove unneeded allOf usages
- correct example definitions / indentation fixes

Removing example values should only be done if those are set incorrectly, otherwise I would also be in favor of keeping them.

types/rewards.yaml
jeluard: Done

jeluard:
@nflaig @dapplion This PR now depends on https://github.com/ethereum/beacon-APIs/pull/409 for the openapi upgrade.
I removed all irrelevant `example` usages.

So we are left with what is required to remove the incorrect `allOf` usages:
* rely on the new `description` override feature
* pull up `example` in `Response`s

Note that some `allOf` were already useless.

types/p2p.yaml
```diff
@@ -8,16 +8,13 @@ NetworkIdentity:
...
-          - description: "Node's addresses on which eth2 RPC requests are served. [Read more](https://docs.libp2p.io/reference/glossary/#multiaddr)"
+        $ref: "./p2p.yaml#/Multiaddr"
+        description: "Node's addresses on which eth2 RPC requests are served. [Read more](https://docs.libp2p.io/reference/glossary/#multiaddr)"
     discovery_addresses:
       type: array
       items:
-        allOf:
-          - $ref: "./p2p.yaml#/Multiaddr"
-          - description: "Node's addresses on which is listening for discv5 requests. [Read more](https://docs.libp2p.io/reference/glossary/#multiaddr)"
-          - example: "/ip4/7.7.7.7/udp/30303/p2p/QmYyQSo1c1Ym7orWxLYvCrM2EmxFTANf8wXmmE7DWjhx5N"
```
nflaig, line 20: This example is kinda odd in the first place as is uses the default EL p2p port. Maybe the intend was to use a udp example here?

types/p2p.yaml
```diff
@@ -27,19 +24,14 @@ MetaData:
...
-      allOf:
-        - description: "Uint64 starting at 0 used to version the node's metadata. If any other field in the local MetaData changes, the node MUST increment seq_number by 1."
-        - $ref: "./primitive.yaml#/Uint64"
+      $ref: "./primitive.yaml#/Uint64"
+      description: "Uint64 starting at 0 used to version the node's metadata. If any other field in the local MetaData changes, the node MUST increment seq_number by 1."
     attnets:
-      allOf:
-        - description: "Bitvector representing the node's persistent attestation subnet subscriptions."
-        - $ref: "./primitive.yaml#/Hex"
-        - example: "0x0000000000000000"
```
nflaig, line 37: Need to be careful when removing examples, the default `Hex` example does not use the length you would expect for `attnets` whereas the override example does

types/p2p.yaml
```diff
@@ -27,19 +24,14 @@ MetaData:
...
+      $ref: "./primitive.yaml#/Hex"
+      description: "Bitvector representing the node's persistent attestation subnet subscriptions."
     syncnets:
-      allOf:
-        - description: "Bitvector representing the node's sync committee subnet subscriptions. This metadata is not present in phase0, but will be present in Altair."
-        - $ref: "./primitive.yaml#/Hex"
-        - example: "0x0f"
```
nflaig, line 42: Same as above, not in favor of removing override example here

types/primitive.yaml
```diff
@@ -18,10 +18,9 @@ Version:
   example: "Lighthouse/v0.1.5 (Linux x86_64)"
 
 GenesisTime:
-  allOf:
-    - $ref: "./primitive.yaml#/Uint64"
-    - example: "1590832934"
-    - description: "The genesis_time configured for the beacon node, which is the unix time in seconds at which the Eth2.0 chain began."
+  $ref: "./primitive.yaml#/Uint64"
+  description: "The genesis_time configured for the beacon node, which is the unix time in seconds at which the Eth2.0 chain began."
+  example: "1590832934"
```
nflaig, line 23: Should make sure to use same order of properties everywhere, or how did you determine if description vs example should be 2nd property? Looking at `ExecutionOptimistic` below, the original order was swapped

types/altair/sync_committee.yaml
```diff
@@ -61,42 +57,33 @@ Altair:
...
-        allOf:
-          - $ref: '../primitive.yaml#/Uint64'
-          - description: 'The index of the subcommittee that the contribution pertains to.'
+        $ref: '../primitive.yaml#/Uint64'
+        description: 'The index of the subcommittee that the contribution pertains to.'
       aggregation_bits:
-        allOf:
-          - description: 'A bit is set if a signature from the validator at the corresponding index in the subcommittee is present in the aggregate `signature`.'
-          - $ref: "../primitive.yaml#/Hex"
-          - example: "0x01"
```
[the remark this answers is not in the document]
jeluard: Reverted all example changes

types/primitive.yaml
jeluard: There is no consistent ordering in general. But that's probably for a different PR.
That said, I would be in favor of the following:
* `type` or `$ref`
* `description`
* `example`

types/altair/sync_committee.yaml
```diff
@@ -67,43 +63,35 @@ Altair:
...
+        $ref: '../primitive.yaml#/Uint64'
+        description: 'The index of the subcommittee that the contribution pertains to.'
       aggregation_bits:
-        allOf:
-          - description: 'A bit is set if a signature from the validator at the corresponding index in the subcommittee is present in the aggregate `signature`.'
-          - $ref: "../primitive.yaml#/Bitvector"
-          - example: "0xffffffffffffffffffffffffffffffff"
+        description: 'A bit is set if a signature from the validator at the corresponding index in the subcommittee is present in the aggregate `signature`.'
+        $ref: "../primitive.yaml#/Bitvector"
+        example: "0xffffffffffffffffffffffffffffffff"
```
nflaig, line 77: does not follow ordering discussed in https://github.com/ethereum/beacon-APIs/pull/407#discussion_r1450450306

apis/node/syncing.yaml
nflaig, line 1: could apply consistent ordering of properties in this file

types/p2p.yaml
```diff
@@ -28,19 +26,16 @@ MetaData:
...
+      $ref: "./primitive.yaml#/Uint64"
+      description: "Uint64 starting at 0 used to version the node's metadata. If any other field in the local MetaData changes, the node MUST increment seq_number by 1."
     attnets:
-      allOf:
-        - description: "Bitvector representing the node's persistent attestation subnet subscriptions."
-        - $ref: "./primitive.yaml#/Bitvector"
-        - example: "0x0000000000000000"
+      description: "Bitvector representing the node's persistent attestation subnet subscriptions."
+      $ref: "./primitive.yaml#/Bitvector"
+      example: "0x0000000000000000"
```
nflaig, line 34: ordering

types/p2p.yaml
```diff
@@ -28,19 +26,16 @@ MetaData:
...
     syncnets:
-      allOf:
-        - description: "Bitvector representing the node's sync committee subnet subscriptions. This metadata is not present in phase0, but will be present in Altair."
-        - $ref: "./primitive.yaml#/Bitvector"
-        - example: "0x0f"
+      description: "Bitvector representing the node's sync committee subnet subscriptions. This metadata is not present in phase0, but will be present in Altair."
+      $ref: "./primitive.yaml#/Bitvector"
+      example: "0x0f"
```
nflaig, line 38: ordering

types/primitive.yaml
```diff
@@ -18,10 +18,9 @@ Version:
   example: "Lighthouse/v0.1.5 (Linux x86_64)"
 
 GenesisTime:
-  allOf:
-    - $ref: "./primitive.yaml#/Uint64"
-    - example: "1590832934"
-    - description: "The genesis_time configured for the beacon node, which is the unix time in seconds at which the Eth2.0 chain began."
+  $ref: "./primitive.yaml#/Uint64"
+  example: "1590832934"
+  description: "The genesis_time configured for the beacon node, which is the unix time in seconds at which the Eth2.0 chain began."
```
nflaig, line 23: ordering

types/primitive.yaml
```diff
@@ -42,21 +41,18 @@ Int64:
...
+  description: "The block root that this response is dependent on."
 
 ExecutionOptimistic:
-  allOf:
-    - type: boolean
-    - example: false
-    - description: "True if the response references an unverified execution payload. Optimistic information may be invalidated at a later time. If the field is not present, assume the False value."
+  type: boolean
+  example: false
+  description: "True if the response references an unverified execution payload. Optimistic information may be invalidated at a later time. If the field is not present, assume the False value."
```
nflaig, line 50: ordering

types/primitive.yaml
```diff
@@ -42,21 +41,18 @@ Int64:
...
 
 Finalized:
-  allOf:
-    - type: boolean
-    - example: false
-    - description: "True if the response references the finalized history of the chain, as determined by fork choice. If the field is not present, additional calls are necessary to compare the epoch of the requested information with the finalized checkpoint."
+  type: boolean
+  example: false
+  description: "True if the response references the finalized history of the chain, as determined by fork choice. If the field is not present, additional calls are necessary to compare the epoch of the requested information with the finalized checkpoint."
```
nflaig, line 55: ordering

nflaig (approved):
lgtm, just a few nits regarding consistent ordering of properties. 

This fixes error examples when working on spec in dev mode (without bundling step)

types/altair/sync_committee.yaml
jeluard: The idea is to keep the order unchanged in this PR (should be the case now) and have another one introducing a consistent order with associated CI and command to fix it.

nflaig (approved):
Looks good to me, as discussed in https://github.com/ethereum/beacon-APIs/pull/407/files#r1452482046, previous ordering is kept to keep changes as minimal as possible

rolfyone:
This runs ok, are we ok to merge?

jeluard:
@rolfyone fine with me!

dapplion (approved):
LGTM

Issue #402, opened 2024-01-08
Derive OpenAPI type schema from pyspec


dapplion:
Thanks to [ethereum/consensus-specs/pull/3506](https://github.com/ethereum/consensus-specs/pull/3506) we have canonical mapping of SSZ -> JSON, so it is possible to derive OpenAPI schemas for all the spec types.

Since the vast majority of API types are spec types auto-deriving the types would reduce future fork development and maintenance. Plus it paves the way to SSZ-ing most API routes.

We can define API-only extra types in SSZ syntax, such as:

```markdown
// extra_types/bellatrix.md

## Containers

### Registration

From the Builder API specification

```python
class ValidatorRegistration(Container):
    fee_recipient: ExecutionAddress  # Address to receive fees from the block
    gas_limit: uint64  # Preferred gas limit of validator
    timestamp: uint64  # Unix timestamp of registration
    pubkey: BLSPubkey  # BLS public key of validator

class SignedValidatorRegistration(Container):
    message: ValidatorRegistration
    signature: BLSSignature

```

### Current status

I've created a simple python script to translate the pyspec to OpenAPI schemas https://pypi.org/project/pyspec2openapi. Here is a demo integration into the beacon-APIs repo https://github.com/dapplion/eth2.0-APIs/pull/1

The **diff** is quite large since it merges all types into a single output file. 

One can compare the bundled outputs with

```
git checkout master
swagger-cli bundle ./beacon-node-oapi.yaml -r -t yaml -o ./deploy/beacon-node-oapi-master.yaml
git checkout dapplion/derive-spec-types
swagger-cli bundle ./beacon-node-oapi.yaml -r -t yaml -o ./deploy/beacon-node-oapi-derive.yaml
diff -l -u deploy/beacon-node-oapi-master.yaml deploy/beacon-node-oapi-derive.yaml | colordiff | more -R 
```

The diff is also quite substantial due to some key characteristics 
- Usage of allOf to define block types. Generated spec does not de-duplicate keys, resulting in less indentations
```
  BeaconBlock:
    allOf:
      - $ref: '.../BeaconBlockCommon'
      - $ref: '#/Deneb/BeaconBlockBody'
```
- Lots of comments per-property and per type not present in the generated specs
https://github.com/ethereum/beacon-APIs/blob/4882aa0803b622b75bab286b285599d70b7a2429/types/block.yaml#L24

### Next steps

[...]

PR #408, opened 2024-01-11
Remove generic Hex primitive type

nflaig:
**Motivation**

The Beacon api spec currently uses an unspec'd hex type for properties which could use spec'd types instead. Using a spec type allows to enforce stricter validation via pattern and provides better information for consumers of the api spec.

This cleanup also removes some extra diff when deriving types from the consensus spec in the future as proposed in https://github.com/ethereum/beacon-APIs/issues/402.

**Description**

- removes generic Hex primitive type
- update properties that referenced it with more specific type
- update some example values
- stricter pattern for Bitlist and Bitvector


dapplion (dismissed):

rolfyone (approved):
LGTM

PR #409, opened 2024-01-12
chore: upgrade to latest openapi

jeluard:
Starting with [OpenAPI 3.1](https://spec.openapis.org/oas/latest.html#fixed-fields-18) `description` can be [overridden](https://github.com/OAI/OpenAPI-Specification/pull/2181) at the `$ref` level.

Migrating to this stable version allows to fix an incorrect usage of `allOf` attempting to overcome the previous limitation.

**with openapi 3.0**

```yaml
type: object
properties:
 root:
  allOf:
   - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
   - description: HashTreeRoot of BeaconState object
```

**with openapi 3.1**

```yaml
type: object
properties:
 root:
    $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Root'
    description: HashTreeRoot of BeaconState object
```

The migration to openapi 3.1 is fairly smooth and only involves replacing the now deprecated `nullable` attribute with [null type](https://github.com/OAI/OpenAPI-Specification/releases/tag/3.1.0-rc0)

required by https://github.com/ethereum/beacon-APIs/pull/407


rolfyone (approved):
LGTM

PR #410, opened 2024-01-12
fix: correctly merge enums

jeluard:
Enums are currently merged using `allOf`. This is incorrect and leads to impossible validation as `allOf` does `intersection` (i.e value must be present in both enum sets to be correct), not `union` (i.e. value must be present in one of both sets, what we want here).

While there is no good built-in way to create [extensible enum](https://github.com/OAI/OpenAPI-Specification/issues/1552) in openapi, using [oneOf](https://github.com/OAI/OpenAPI-Specification/issues/1320#issuecomment-322536342) is considered an acceptable workaround.

Note that while merged values appear in the swagger UI, this is actually a [bug](https://github.com/swagger-api/swagger-js/issues/2472)


rolfyone (approved):
LGTM

PR #426, opened 2024-02-18
chore: Rewards apis no longer experimental

rolfyone:


nflaig (approved):
LGTM

beacon-node-oapi.yaml
```diff
@@ -58,8 +58,6 @@ tags:
       [Checkout validator flow](./validator-flow.md) to learn how to use this api.
   - name: Rewards
     description: Endpoints to query rewards and penalties for validators.
-  - name: Experimental
```
ensi321, line 61: @rolfyone I know this has been merged but imo we should keep the `Experimental` tag and its description just in case any future endpoints require it

rolfyone: We'll definitely re-use it, trivial to re-add. no point keeping a mess around imo...

nflaig: It also keeps an empty section around in the api explorer (at the bottom)

arnetheduck:
Are there test cases for the  response values rewards API?

PR #428, opened 2024-02-18
Remove leftover block.yaml file

nflaig:
This API has been effectively removed in https://github.com/ethereum/beacon-APIs/pull/218 but file was leftover.


rolfyone (approved):
LGTM

PR #432, opened 2024-03-01
Update Lodestar column in changelog

nflaig:
Similar to [keymanager client support table](https://github.com/ethereum/keymanager-APIs?tab=readme-ov-file#client-support) it might be useful for downstream consumers to know which APIs are supported by each client.

Hoping this PR encourages other teams to fill out this information as well :)


rolfyone (approved):
LGTM

PR #438, opened 2024-03-29
Removed default builder boost

fredriksvantes:
Removed "default" boost wording where the default value should be a fixed to 100 on clients as this has been changed to 90 on multiple clients.


mcdee:
We can't just remove the default, it's already part of the released spec and it's possible that validator clients are sending in requests without this value and expecting it to be 100.

Given that at this point there are client releases that have altered this value I think the best bet would be to state that if the parameter is not supplied then it defaults to whatever value is configured in the beacon node.  Although that doesn't adhere to the spec as-is, and so is still a spec change, I think it is the closest that we will get when taking reality into account.

rolfyone:
Given that we're just removing the 'default' term, this PR seems reasonable in description?

rolfyone (approved):

PR #440, opened 2024-04-07
Return 404 error if no sync committee contribution is available

nflaig:
Clarify that 404 status code (error) should be returned if no sync committee contribution is available for the requested `beacon_block_root`.

See related PR https://github.com/ethereum/beacon-APIs/pull/380 that added the same note for no aggregated attestation available.


apis/validator/sync_committee_contribution.yaml
```diff
@@ -10,6 +10,9 @@ get:
     A 503 error must be returned if the block identified by the response
     `beacon_block_root` is optimistic (i.e. the sync committee contribution
     refers to a block that has not been fully verified by an execution engine).
+
+    A 404 error must be returned if no sync committee contribution is available
+    for the requested `beacon_block_root`.
```
rolfyone, line 15: is the implication here if the block denoted by beacon_block_root is not present?

Are there any other reason we'd not see a contribution?

eg. if i request sync committee contribution from 12 months ago, and only 6 months blocks are stored, then i'd expect 404...

nflaig: > is the implication here if the block denoted by beacon_block_root is not present?

The two conditions I see where a 404 should be returned is if
- block as denoted by beacon_block_root is not known or not a descendant of the finalized root, as determined by fork choice
- or if beacon node did not prepare a contribution for beacon_block_root, this might be implementation specific but _might_ happen if validator client did not [submit a sync committee signature](https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolSyncCommitteeSignatures) earlier in the slot

> eg. if i request sync committee contribution from 12 months ago, and only 6 months blocks are stored, then i'd expect 404...

This API is part of validator APIs which are not intended to serve historical data so I am not sure this is a case we need to consider here. At least in Lodestar we serve this data from a cache which prunes any data from previous slots based on clock slot.

But yes, that would result in a 404 as well..

Unless I am missing something there is no equivalent to [getAggregatedAttestation](https://ethereum.github.io/beacon-APIs/#/Validator/getAggregatedAttestation), so would have the request the whole block to get sync committee contributions from 12 months ago.

rolfyone: Yes i think we'd need the block. Ok I think this makes sense, will flag to review on discord.

mcdee (approved):

PR #442, opened 2024-04-12
Clarify use of 206 status code for node health.

mcdee:
The 206 return code for the node health endpoint does not take into account the situation where the consensus node may be synced by the execution node is optimistic.  Clarify that this situation should return the 206 status code.

### DIFF
--- a/apis/node/health.yaml
+++ b/apis/node/health.yaml
@@ -17,7 +17,7 @@ get:
     "200":
       description: Node is ready
     "206":
-      description: Node is syncing but can serve incomplete data
+      description: Either the consensus node is syncing, or its execution node is optimistic, so data served may be incorrect
     "400":
       description: Invalid syncing status code
     "503":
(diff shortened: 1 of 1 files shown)


nisdas (approved):

apis/node/health.yaml
```diff
@@ -17,7 +17,7 @@ get:
     "200":
       description: Node is ready
     "206":
-      description: Node is syncing but can serve incomplete data
+      description: Either the consensus node is syncing, or its execution node is optimistic, so data served may be incorrect
```
nflaig, line 20: The spec does not use the term "consensus node" while "beacon node" is used 85 times

```suggestion
      description: Either the beacon node is syncing, or its execution node is optimistic, so data served may be incorrect
```

Probably personal preference but in my mind beacon node is also more accurate, a consensus node might suggest that it also includes a validator client as a beacon node by itself can't participate in consensus

nflaig (approved):

rkapka (approved):

PR #447, opened 2024-05-09
Add getAggregatedAttestationV2 to support EIP-7549

### LINKED ISSUE #445: Update attestation related APIs to be compatible with EIP-7549
attestations for `Electra` require an additional `committee_bits` field.

I think these APIs are effected.
`/eth/v1/beacon/pool/attestations`
`/eth/v1/beacon/blocks/{block_id}/attestations`
`/eth/v1/validator/aggregate_attestation`

ensi321:
`AttestationData` post-electra is committee-indifferent, and thus, the same `attestation_data_root` might be shared across different committees. 

An aggregator is only interested in attestations in its corresponding committee. `getAggregatedAttestationV1` might aggregate attestations from multiple committees since `committee_index` is not in its params. 

This PR adds `getAggregatedAttestationV2` to aggregate attestations given `attestation_data_root` + `slot` + `committee_index`.

Part of #445

### DIFF
--- a/apis/validator/aggregate_attestation.v2.yaml
+++ b/apis/validator/aggregate_attestation.v2.yaml
@@ -1,0 +1,50 @@
+get:
+  operationId: "getAggregatedAttestationV2"
+  summary: "Get aggregated attestation"
+  description: |
+    Aggregates all attestations matching given attestation data root, slot and committee index.
+
+    A 503 error must be returned if the block identified by the response
+    `beacon_block_root` is optimistic (i.e. the aggregated attestation attests
+    to a block that has not been fully verified by an execution engine).
+
+    A 404 error must be returned if no attestation is available for the requested
+    `attestation_data_root`.
+  tags:
+    - ValidatorRequiredApi
+    - Validator
+  parameters:
+    - name: attestation_data_root
+      in: query
+      required: true
+      description: "HashTreeRoot of AttestationData that validator wants aggregated"
+      schema:
+        $ref: '../../beacon-node-oapi.yaml#/components/schemas/Root'
+    - name: slot
+      in: query
+      required: true
+      schema:
+        $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+    - name: committee_index
+      in: query
+      required: true
+      schema:
+        $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+  responses:
+    "200":
+      description: "Returns aggregated `Attestation` object with same `AttestationData` root, slot and committee index."
+      content:
+        application/json:
+          schema:
+            title: GetAggregatedAttestationV2Response
+            type: object
+            required: [data]
+            properties:
+              data:
+                $ref: '../../beacon-node-oapi.yaml#/components/schemas/Electra.Attestation'
+    "400":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
+    "404":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/NotFound'
+    "500":
+      $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
--- a/apis/validator/aggregate_attestation.v2.yaml
+++ b/apis/validator/aggregate_attestation.v2.yaml
@@ -33,12 +33,15 @@ get:
   responses:
     "200":
       description: "Returns aggregated `Attestation` object with same `AttestationData` root, slot and committee index."
+      headers:
+        Eth-Consensus-Version:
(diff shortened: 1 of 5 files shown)


apis/validator/aggregate_attestation.v2.yaml
```diff
@@ -0,0 +1,50 @@
...
+      description: "Returns aggregated `Attestation` object with same `AttestationData` root, slot and committee index."
+      content:
+        application/json:
+          schema:
+            title: GetAggregatedAttestationV2Response
+            type: object
+            required: [data]
+            properties:
+              data:
+                $ref: '../../beacon-node-oapi.yaml#/components/schemas/Electra.Attestation'
```
nflaig, line 44: we likely wanna use anyOf phase0 and electra attestation to be able to switch to the api pre-electra

apis/validator/aggregate_attestation.v2.yaml
```diff
@@ -0,0 +1,50 @@
...
+  responses:
+    "200":
+      description: "Returns aggregated `Attestation` object with same `AttestationData` root, slot and committee index."
+      content:
+        application/json:
+          schema:
+            title: GetAggregatedAttestationV2Response
+            type: object
+            required: [data]
+            properties:
```
nflaig, line 42: let's make this api fork aware by adding version meta

rkapka:
Adding to @nflaig 's comments, if we make committee index required then this endpoint will not be backwards compatible and validator clients will have to be aware of both.

nflaig (dismissed):

types/electra/attestation.yaml
```diff
@@ -0,0 +1,18 @@
+Electra:
+  Attestation:
+    type: object
+    description: "The [`Attestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.2/specs/electra/beacon-chain.md#attestation) object from the CL spec."
```
nflaig, line 4:
```suggestion
    description: "The [`Attestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/electra/beacon-chain.md#attestation) object from the CL spec."
```

nflaig (approved):

types/electra/attestation.yaml
```diff
@@ -0,0 +1,18 @@
...
+  Attestation:
+    type: object
+    description: "The [`Attestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.3/specs/electra/beacon-chain.md#attestation) object from the CL spec."
+    required: [aggregation_bits, data, signature, committee_bits]
+    properties:
+      aggregation_bits:
+        $ref: "../primitive.yaml#/BitList"
+        description: "Attester aggregation bits."
+      data:
+        $ref: "../attestation.yaml#/AttestationData"
```
rkapka, line 11: Since you added a description to every other field, can you add one to `data` too?

rkapka (approved):

types/electra/attestation.yaml
nflaig: The default description of `AttestationData` should be sufficient as it has not been modified in electra

https://github.com/ethereum/beacon-APIs/blob/672e03e25ace85a3bacaea553fbf374f4f844435/types/attestation.yaml#L46-L48

tersec (approved):

types/electra/attestation.yaml
rolfyone: I'm happy with that default description...

rolfyone (approved):

apis/validator/aggregate_attestation.v2.yaml
```diff
@@ -0,0 +1,59 @@
...
+      schema:
+        $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+    - name: committee_index
+      in: query
+      required: true
+      schema:
+        $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+  responses:
+    "200":
+      description: "Returns aggregated `Attestation` object with same `AttestationData` root, slot and committee index."
```
nflaig, line 35: Noticed the description will not be correct post-electra as `index` in `AttestationData` will be set to 0 and not match the  value passed in `committee_index` query param.

ensi321: @nflaig good catch. Will update now

ensi321: > Noticed the description will not be correct post-electra as `index` in `AttestationData` will be set to 0 and not match the value passed in `committee_index` query param.

I think we can interpret this as:
```
Returns aggregated `Attestation` object with the same root and slot in `Attestation.data`, and same committee index in `Attestation`. 
```

I think this description still holds true. Just that the committee index is now in `Attestation` and not `AttestationData`

PR #448, opened 2024-05-24
Update attestation APIs to support EIP-7549

mehdi-aouadi:
Following the Electra containers updates:
- Update the `/block` and `/pool` attestation GET APIs to handle the new Electra attestations
- Add `Electra.Attestation` schema
- Extract the `AttestationData` schema used by both `Attestation` and `Electra.Attestation`
- Update the attester slashing API
- Update the `IndexedAttestation` schema

Partially fixes https://github.com/ethereum/beacon-APIs/issues/445
Related to https://github.com/ethereum/beacon-APIs/pull/447


apis/beacon/blocks/attestations.yaml
```diff
@@ -13,6 +13,9 @@ get:
   responses:
     "200":
       description: Success
+      headers:
+        Eth-Consensus-Version:
```
nflaig, line 17: I think we have to bump all apis to v2 since we now require the version header to be set

apis/beacon/blocks/attestations.yaml
```diff
@@ -27,7 +30,9 @@ get:
               data:
                 type: array
                 items:
-                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
+                  oneOf:
```
nflaig, line 33:
```suggestion
                  anyOf:
```

apis/beacon/blocks/attestations.yaml
```diff
@@ -27,7 +30,9 @@ get:
               data:
```
nflaig, line 30: json response also needs the version

types/attestation.yaml
```diff
@@ -37,25 +37,8 @@ PendingAttestation:
...
-    index:
-      $ref: "./primitive.yaml#/Uint64"
-    beacon_block_root:
-      $ref: './primitive.yaml#/Root'
-      description: "LMD GHOST vote."
-    source:
-      $ref: "./misc.yaml#/Checkpoint"
-    target:
-      $ref: "./misc.yaml#/Checkpoint"
+      $ref: "./primitive.yaml#/Uint64"
```
nflaig, line 44: missing new line here causes strange diff in line 44

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
...
+        description: "Attesting validator indices"
+        items:
+          $ref: "../primitive.yaml#/Uint64"
+      signature:
+        $ref: '../primitive.yaml#/Signature'
+        description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
+      data:
+        $ref: '../attestation_data.yaml#/AttestationData'
+
+  Attestation:
```
nflaig, line 18: this conflicts with the other open PR https://github.com/ethereum/beacon-APIs/pull/447, might wanna get consensus there on how we go about updating all the APIs and then rebase this branch

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: I thought adding a header won't break backward compatibility but I'm not against bumping this API to v2 too since we're doing that for the other related ones

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: Not sure about this, IMU a block will contain only one type of attestations and the API should a list of one of them

types/attestation.yaml
mehdi-aouadi: Actually I moved the `AttestationData` to a separate schema file since it's shared by both Attestation and `Electra.Attestation`. No need to add a new line then

types/electra/attestation.yaml
mehdi-aouadi: Absolutely, I had in mind to rebase this one once the #447 gets merged

apis/beacon/blocks/attestations.yaml
nflaig: ah right this is an array, `oneOf` should be correct, it only caused issue in other apis and invalidated the openapi spec (https://github.com/ethereum/beacon-APIs/pull/421) but those were just single object cases

mehdi-aouadi: I'm still confused about the usage of `oneOf` and `anyOf` for array items tbh, checked the spec and still not sure about it: In one hand there is the [`oneOf`](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#oneof) which states that the data should be valid against **one** of the specified schemas and in the other hand the [mixed typed array](https://swagger.io/docs/specification/data-models/data-types/#array) using `oneOf` which could contain both types

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: wdym here? the version in in the header

nflaig: All existing apis that have the version in the header also return it as part of json
https://github.com/ethereum/beacon-APIs/blob/fb718b65d39b62ffd4ee6c95d1d6fb1458c08958/apis/beacon/blocks/block.v2.yaml#L27

apis/beacon/blocks/attestations.yaml
nflaig: Current approach seems correct to me, we need to apply the `oneOf` / `anyOf` on the array itself, not the items. Confirmed this results in the desired behavior with a [simple example](https://gist.github.com/nflaig/cd1cf5a03f9da6bcb055d639ddc8a6dd) using the schema validator we use.

apis/beacon/blocks/attestations.yaml
```diff
@@ -25,9 +28,13 @@ get:
               finalized:
                 $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Finalized"
               data:
-                type: array
-                items:
-                  $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
+                oneOf:
```
nflaig, line 31: here it does not make a difference if we use `anyOf` now instead which I think we should prefer to avoid [previous issue](https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fethereum.github.io%2Fbeacon-APIs%2Freleases%2Fv2.4.2%2Fbeacon-node-oapi.json)

nflaig:
What about attestation event, should it be part of this PR?

https://github.com/ethereum/beacon-APIs/blob/fb718b65d39b62ffd4ee6c95d1d6fb1458c08958/apis/eventstream/index.yaml#L64-L68

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: Bumped up both APIs to V2 since the version is required in the data and that would be a breaking change

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: Great thanks! Used `anyOf` everywhere now

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: I see, added the version as a property too. Thanks

apis/beacon/blocks/attestations.yaml
mehdi-aouadi: done

mehdi-aouadi:
> What about attestation event, should it be part of this PR?
> 
> https://github.com/ethereum/beacon-APIs/blob/fb718b65d39b62ffd4ee6c95d1d6fb1458c08958/apis/eventstream/index.yaml#L64-L68

I considered the event stream events out of the scope of this PR. Might be addressed in a separate PR imo

apis/beacon/pool/attester_slashings.v2.yaml
```diff
@@ -0,0 +1,62 @@
...
+    "500":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
+
+post:
+  operationId: submitPoolAttesterSlashingsV2
+  summary: Submit AttesterSlashing object to node's pool
+  description: Submits AttesterSlashing object to node's pool and if passes validation node MUST broadcast it to network.
+  tags:
+    - Beacon
+  requestBody:
```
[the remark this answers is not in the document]
mehdi-aouadi: Added

apis/beacon/pool/attestations.yaml
```diff
@@ -18,6 +18,9 @@ get:
   responses:
     "200":
       description: Successful response
+      headers:
+        Eth-Consensus-Version:
```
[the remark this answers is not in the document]
mehdi-aouadi: np, I pushed some commits after requesting a new review that's why...

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
+Electra:
+  IndexedAttestation:
+    type: object
+    description: "The [`IndexedAttestation`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#indexedattestation) object from the CL spec."
```
rolfyone, line 4: i think we'd be better off locking it into a spec version, even if it's alpha... it'll mean we need to update it at some point but even currently we'll need to move it off of dev...

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
...
+          $ref: "../primitive.yaml#/Uint64"
+      signature:
+        $ref: '../primitive.yaml#/Signature'
+        description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
+      data:
+        $ref: '../attestation_data.yaml#/AttestationData'
+
+  Attestation:
+    type: object
+    description: "The [`Attestation`](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#attestation) object from the CL spec."
```
rolfyone, line 20: same everywhere really, lock to the alpha.2 for these instead of `/dev`

rolfyone:
Yep I agree with new version for interfaces given they're a new required field (breaking) and headers.
minor comment about references to `/dev` versions.

types/electra/attestation.yaml
mehdi-aouadi: done, used the alpha version for now

types/electra/attestation.yaml
mehdi-aouadi: done, used the alpha version for now

rolfyone: great - thats better - we'll just need to periodically update it.

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
...
+    If an attestation is validated successfully the node MUST publish that attestation on the appropriate subnet.
+
+    If one or more attestations fail validation the node MUST return a 400 error with details of which attestations have failed, and why.
+  parameters:
+    - in: header
+      schema:
+        $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion'
+      required: true
+      name: Eth-Consensus-Version
+      description: "Version of the block being submitted."
```
rkapka, line 70:
```suggestion
      description: "Version of the attestation being submitted."
```

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
...
+            code: 400
+            message: "Invalid slot: current"
+    "500":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
+
+post:
+  operationId: submitPoolAttestationsV2
+  summary: Submit Attestation objects to node
+  description: |
+    Submits Attestation objects to the node.  Each attestation in the request body is processed individually.
```
rkapka, line 59:
```suggestion
    Submits Attestation objects to the node. Each attestation in the request body is processed individually.
```

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
...
+    
+    If an attestation is validated successfully the node MUST publish that attestation on the appropriate subnet.
```
rkapka, line 61:
```suggestion
    If an attestation is validated successfully, the node MUST publish that attestation on the appropriate subnet.
```

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+    If one or more attestations fail validation the node MUST return a 400 error with details of which attestations have failed, and why.
```
rkapka, line 63:
```suggestion
    If one or more attestations fail validation, the node MUST return a 400 error with details of which attestations have failed, and why.
```

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
...
+          anyOf:
+            - type: array
+              items:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
+            - type: array
+              items:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Electra.Attestation'
+  responses:
+    "200":
+      description: Attestations are stored in pool and broadcast on appropriate subnet
```
rkapka, line 88:
```suggestion
      description: Attestations are stored in pool and broadcast on the appropriate subnet
```

apis/beacon/pool/attester_slashings.v2.yaml
```diff
@@ -0,0 +1,69 @@
...
+                  - type: array
+                    items:
+                      $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Electra.AttesterSlashing'
+    "500":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
+
+post:
+  operationId: submitPoolAttesterSlashingsV2
+  summary: Submit AttesterSlashing object to node's pool
+  description: Submits AttesterSlashing object to node's pool and if passes validation node MUST broadcast it to network.
```
rkapka, line 38:
```suggestion
  description: Submits AttesterSlashing object to node's pool. Upon successful validation the node MUST broadcast it to the network.
```

apis/beacon/pool/attester_slashings.v2.yaml
```diff
@@ -0,0 +1,69 @@
...
+  parameters:
+    - in: header
+      schema:
+        $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion'
+      required: true
+      name: Eth-Consensus-Version
+      description: "Version of the block being submitted."
```
rkapka, line 45:
```suggestion
      description: "Version of the attester slashing being submitted."
```

apis/beacon/pool/attestations.v2.yaml
```diff
@@ -0,0 +1,96 @@
+get:
+  operationId: getPoolAttestationsV2
+  summary: Get Attestations from operations pool
```
rkapka, line 3:
```suggestion
  summary: Get attestations from operations pool
```

apis/beacon/pool/attester_slashings.v2.yaml
```diff
@@ -0,0 +1,69 @@
+get:
+  operationId: getPoolAttesterSlashingsV2
+  summary: Get AttesterSlashings from operations pool
```
rkapka, line 3:
```suggestion
  summary: Get attester slashings from operations pool
```

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
...
+    properties:
+      attesting_indices:
+        type: array
+        maxItems: 131072
+        description: "Attesting validator indices"
+        items:
+          $ref: "../primitive.yaml#/Uint64"
+      signature:
+        $ref: '../primitive.yaml#/Signature'
+        description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
```
rkapka, line 14: What does "validator of the attestation mean"?

mehdi-aouadi: IMU it means that it comes from the corresponding `Attestation` whose signature was created by the validator which generated it

rkapka (approved):

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
...
+    type: object
+    description: "The [`Attestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.2/specs/electra/beacon-chain.md#attestation) object from the CL spec."
+    required: [aggregation_bits, signature, committee_bits, data]
+    properties:
+      aggregation_bits:
+        $ref: "../primitive.yaml#/BitList"
+        description: "Attester aggregation bits."
+      data:
+        $ref: '../attestation_data.yaml#/AttestationData'
+      committee_bits:
```
rkapka, line 28: Let's move this field to the end as per https://github.com/ethereum/consensus-specs/pull/3768

rkapka (changes requested):

types/electra/attestation.yaml
```diff
@@ -0,0 +1,33 @@
+Electra:
+  IndexedAttestation:
+    type: object
+    description: "The [`IndexedAttestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.2/specs/electra/beacon-chain.md#indexedattestation) object from the CL spec."
+    properties:
```
nflaig, line 5: `required` properties filed is missing here

nflaig (commented):
need to resolve conflicts since we merged https://github.com/ethereum/beacon-APIs/pull/447

types/electra/attestation.yaml
mehdi-aouadi: Rebased the branch on to of master, the order should be good now

types/electra/attestation.yaml
mehdi-aouadi: Good catch, they were missing in the existing `IndexedAttestaion`, added them there too

types/electra/attestation.yaml
```diff
@@ -8,11 +25,11 @@ Electra:
...
+        $ref: '../attestation_data.yaml#/AttestationData'
       signature:
-        $ref: "../primitive.yaml#/Signature"
+        $ref: '../primitive.yaml#/Signature'
         description: "BLS aggregate signature."
       committee_bits:
         $ref: "../primitive.yaml#/Bitvector"
         description: "Committee bits."
-        example: "0x0000000000000000000000000000000000000000000000000000000000000001"
+        example: "0x000000000000000000000000000000000000000000000000000"
```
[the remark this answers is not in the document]
mehdi-aouadi: My bad, rebasing issue...

beacon-node-oapi.yaml
```diff
@@ -373,8 +383,6 @@ components:
       $ref: './types/primitive.yaml#/Blob'
     Deneb.BlobSidecars:
       $ref: './types/deneb/blob_sidecar.yaml#/Deneb/BlobSidecars'
-    Electra.Attestation:
```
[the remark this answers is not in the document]
mehdi-aouadi: My bad rebasing issue. Moved the `ElectraAttesterSlashing` here too

types/electra/attestation.yaml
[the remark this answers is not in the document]
nflaig: > Put bad the initial example

this initial one was wrong as well, put correct length now

nflaig (approved):

nflaig (approved):

dapplion (approved):

rkapka (approved):

Emmanuel0899 (approved):

PR #452, opened 2024-06-10
Validator identities endpoint.

mcdee:
In support of #449.


rolfyone:
This looks good. Will cross-post and seek more opinions.

types/api.yaml
```diff
@@ -45,6 +45,20 @@ ValidatorBalanceResponse:
...
+  type: object
+  required: [index, pubkey, activation_epoch]
+  properties:
+    index:
+      $ref: './primitive.yaml#/Uint64'
+      description: "Index of validator in validator registry."
+    pubkey:
+      $ref: './primitive.yaml#/Pubkey'
+      description: "Public key of validator."
+    activation_epoch:
```
dapplion, line 58: Why add the `activation_epoch`? It's a curious in-between of no epoch and all epochs in the validator record

rkapka:
I support this. Our validator client currently needs to grab the full validator to get the index

nflaig:
> I support this. Our validator client currently needs to grab the full validator to get the index

Curious, do you track the validator status on the VC side? This has a lot of benefits, e.g. not querying duties for exited validators, not registering to builder etc. but it would still require to query the full validator object if we don't include all information in this API to recalculate the status, or include the status itself here, preferably in a format that can be ssz-serialized.

rkapka:
> Curious, do you track the validator status on the VC side?

No, we don't. I agree it would allow for certain optimizations, but I am not sure if it's worth the extra complexity.

rolfyone:
I think i prefer the idea that this is a simple index lookup that would allow us to continue to use it in a post index being unique world, and for the rest of the data you have the indices you could look up further info...

rolfyone:
we can definitely add this to changes table though...

apis/beacon/states/validator_identities.yaml
```diff
@@ -0,0 +1,69 @@
+post:
+  operationId: "postStateValidatorIdentities"
```
nflaig, line 2: The http method used is a technical detail imo and we don't need it as part of operation id, it might be a bit confusing semantically

```suggestion
  operationId: "getStateValidatorIdentities"
```

nflaig: For context, we use the `operationId` as method in our api client and error messages

mcdee: We do seem to use the HTTP method elsewhere in the operation ID, notably we already have `getStateValidators` and `postStateValidators` in the beacon API.  Given that these methods have different signatures (data in body vs in request)  I'd be quite wary in changing it in this way.

rolfyone: I think in this instance i'd prefer the `post`... i do think sometimes we call them more in line with the intent, but it also seems very minor...
This way (post) is arguably the same as the current validators endpoint so there's that...

nflaig: yeah, let's keep it `post` then but should avoid adopting this convention in other routes if possible (outside of validators)

apis/beacon/states/validator_identities.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+     Depending on `Accept` header data can be returned either as JSON or as bytes serialized by SSZ.
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+  requestBody:
+    description: "An array of either hex encoded public keys (any bytes48 with 0x prefix) or validator indices"
```
nflaig, line 19: This would make it slightly clearer that it can be a mixed array (from https://github.com/ethereum/beacon-APIs/pull/453)

```suggestion
    description: "An array of values, with each value either a hex encoded public key (any bytes48 with 0x prefix) or a validator index."
```

apis/beacon/states/validator_identities.yaml
```diff
@@ -0,0 +1,69 @@
...
+    "404":
+      description: "State not found"
+      content:
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+          example:
+            code: 404
+            message: "State not found"
+    "500":
```
nflaig, line 68: would be good to add 406 here as well but depends on https://github.com/ethereum/beacon-APIs/pull/457 being merged fist

types/api.yaml
dapplion: Index re-use won't come until years down the line if ever. Is it better to do a v2 in 2030 when that happens?

rolfyone: i was thinking there's no harm having it?

apis/beacon/states/validator_identities.yaml
mcdee: Added, thanks!

nflaig (approved):

rolfyone (approved):

apis/beacon/states/validator_identities.yaml
```diff
@@ -0,0 +1,71 @@
...
+  operationId: "postStateValidatorIdentities"
+  summary: "Get validator identities from state"
+  description: |
+    Returns filterable list of validators identities.
+
+    Identities will be returned for all indices or public keys that match known validators. If an index or public key does not
+    match any known validator, no identity will be returned but this will not cause an error. There are no guarantees for the
+    returned data in terms of ordering.
+
+     Depending on `Accept` header data can be returned either as JSON or as bytes serialized by SSZ.
```
nflaig, line 11: was wondering if we wanna state this here, seems a bit redundant considering we have a more general note about this now since https://github.com/ethereum/beacon-APIs/pull/457 and it's already implied by the response content type + the Accept header is mentioned there as well.

The reason to add the note is that it seems consistent with other endpoint that support SSZ, however could consider cleaning those up a bit

rolfyone: i'd be happy to just merge and cleanup as a separate task...

PR #453, opened 2024-06-12
Clarify requirement for validator balances body

mcdee:
Looking at the `validator_balances` endpoint there is some ambiguity around what to do in some cases.  Specifically, for POST to `/eth/v1/beacon/states/head/validator_balances`:

With no body:

- lighthouse returns an error
- lodestar returns an error
- nimbus returns an error
- prysm returns an error
- teku returns all balances

With an empty array in the body:

- lighthouse returns no balances
- lodestar returns no balances
- nimbus returns all balances
- prysm returns all balances
- teku returns all balances

I think that the best balance of attempting to retain existing functionality whilst firming up the spec is to state that no body will return an error, and an empty array will return all balances.  This also ensures that users have a way to obtain all balances easily..

### DIFF
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -79,7 +79,7 @@ post:
   description: |
     Returns filterable list of validators balances.
 
-    Balances will be returned for all indices or public key that match known validators. If an index or public key does not
+    Balances will be returned for all indices or public keys that match known validators. If an index or public key does not
     match any known validator, no balance will be returned but this will not cause an error. There are no guarantees for the
     returned data in terms of ordering; the index is returned for each balance, and can be used to confirm for which inputs a
     response has been returned.
@@ -90,8 +90,11 @@ post:
       in: path
       $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
   requestBody:
-    description: "An array of either hex encoded public key (any bytes48 with 0x prefix) or validator index"
-    required: false
+    description: |
+      Either hex encoded public key (any bytes48 with 0x prefix) or validator index.
+
+      If the supplied list is empty (i.e. the body is `[]`) then balances will be returned for all validators.
+    required: true
     content:
       application/json:
         schema:


nflaig:
> an empty array will return all balances

Agreed, this behavior is likely for the best. And just yesterday I noticed few interop issues between clients when using `postStateValidators`.

E.g. Prysm VC sends an empty `statuses` array which leads to some clients not returning any data. We might wanna add the same note there as well.

Empty array should equal omitting the property and not apply any filtering as otherwise it will always return no data.

nflaig (dismissed):

dapplion (commented):
It's a bit weird but feels like the best solution given current behaviour

rkapka (dismissed):

rolfyone:
I'm not sure I 100% agree that 'no body' equates to error. It's logically the same as an empty array... We can throw an error, but it's a bit weird IMO. 

In real terms you haven't requested anything specifically in both cases, and the default is 'return all data'.

nflaig:
> I'm not sure I 100% agree that 'no body' equates to error. It's logically the same as an empty array.

We might wanna keep the body as `required : false`, we don't really gain anything if server throws an error if body is omitted, or just leave it up to the server implementation to decide.

This is also much more user friendly if you wanna do a quick curl as you can just do
```
curl -X POST http://localhost:9596/eth/v1/beacon/states/head/validator_balances
```
no extra headers and  `-d "[]"` required

rolfyone:
> > I'm not sure I 100% agree that 'no body' equates to error. It's logically the same as an empty array.
> 
> We might wanna keep the body as `required : false`, we don't really gain anything if server throws an error if body is omitted, or just leave it up to the server implementation to decide.
> 
> This is also much more user friendly if you wanna do a quick curl as you can just do
> 
> ```
> curl -X POST http://localhost:9596/eth/v1/beacon/states/head/validator_balances
> ```
> 
> no extra headers and `-d "[]"` required

It's handy. technically adding required probably could be argued as breaking but given that most clients already error i'm not going to make that argument :) I'm fairly sure that accepting it as equivalent to an empty array is about 1-2 lines of code if people would just do that...
Teku generally has tests for empty array of a post as well as the empty body as equivalence for endpoints that query in that way.

nflaig:
> Teku generally has tests for empty array of a post as well as the empty body as equivalence for endpoints that query in that way.

Have updated Lodestar to behave the same as there is no downside if the server accepts an empty body for routes that only apply it to filter the response, but the client will still send an empty array if no filtering is applied.

rolfyone:
> Have updated Lodestar to behave the same as there is no downside if the server accepts an empty body for routes that only apply it to filter the response, but the client will still send an empty array if no filtering is applied.

this sounds like basically what we do - we always do send empty array if required, as technically that is correct, but no real down side to allowing no body - ta for update.

apis/beacon/states/validator_balances.yaml
```diff
@@ -90,8 +90,11 @@ post:
       in: path
       $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
   requestBody:
-    description: "An array of either hex encoded public key (any bytes48 with 0x prefix) or validator index"
-    required: false
+    description: |
+      Either hex encoded public key (any bytes48 with 0x prefix) or validator index.
+
+      If the supplied list is empty (i.e. the body is `[]`) then balances will be returned for all validators.
```
rolfyone, line 96: Shall we update to `(i.e. the body is [] or the body is empty)` given the discussion in the comments? I don't mind leaving it undefined unless people are going to rely on that behaviour, but it does seem cleaner than an error...

nflaig: in favor of this, also need to make sure body is not required by the schema

mcdee: I'll update accordingly.

mcdee:
Latest update tweaks the wording to consider both `[]` and an empty body as "return all validators".

nflaig (approved):

rolfyone (approved):

PR #459, opened 2024-07-17
Add `eventstreamV2`

ensi321:
We need to communicate fork version during `attestation` and `attester_slashing` since it is modified by EIP-7549.

For example, `"aggregation_bits":"0x01"` will mean different validators in pre and post-EIP-7549.

This PR adds `eventstreamV2` which has `version` in `attestation` and `attester_slashing`

Related to #445


nflaig:
I am in favor of bumping the whole api to v2 instead of just single topics ([previous discussion](https://discord.com/channels/595666850260713488/710831784991916072/1253318472692269066)) but I think we should use this opportunity to make other topics more forward compatible to not have to bump the api again at some (unknown) point in the future.

Further topics we could wrap inside version container are
- `voluntary_exit`: SignedVoluntaryExit (phase0 container of spec)
- `proposer_slashing`: ProposerSlashing  (phase0 container of spec)
- `bls_to_execution_change`: SignedBLSToExecutionChange (capella container of spec)
- `contribution_and_proof`: SignedContributionAndProof (altair container of spec)

In the worst case if those never happen to change in the future, it would at least make the format of events more consistent

rolfyone:
i think if we're wanting to go this way we want to put version on every type...

One option would be to just add `versioned_attestation` to event stream v1... It lets us version out the attestations with no change to the eventstream, and just change the single message we're wanting to update. in future we can do the same at any point with any other event message... This is probably a MVP type solution if we updated V1.

nflaig:
> One option would be to just add versioned_attestation to event stream v1

If we do that, does the `attestation` topic just handle phase0 attestations then?

> This lets us not duplicate our eventstream framework

yeah, we need more feedback from others on the implementation complexity here, might not be as trivial as bumping other apis, it's also a bit more effort on Lodestar side to do this.

rolfyone:
> If we do that, does the `attestation` topic just handle phase0 attestations then?

yes i would expect that it would just handle phase0, and you'd need a new topic for the versioned attestation format...

rolfyone:
quick update - currently its looking like we may be able to avoid this change, I've raised a question in discord.

nflaig:
Not in favor of a v2 api due to high code duplication and maintenance overhead of the server-side implementation

rolfyone:
closing as not planned at the moment.

PR #463, opened 2024-08-07
Add workflow to create pre-releases

nflaig:
**Motivation**

This is a proposal for pre-releases of the beacon api spec similar to what's done on the consensus spec. This allows clients to run spec tests earlier in the development cycle.

**Description**

A pre-release can be created by simply pushing a new tag with the suffix `-alpha.x`. The CD will create a github release and upload the bundled spec files.

Pre-releases will not be listed in `index.html` and not show up on the explorer.


rolfyone (approved):
LGTM

ensi321 (approved):

PR #466, opened 2024-08-19
Remove deprecated produce block apis

nflaig:
This removes previously deprecated produce blocks apis
- `GET /eth/v2/validator/blocks/{slot}`
- `GET /eth/v1/validator/blinded_blocks/{slot}`

Those were replaced / consolidated by `GET /eth/v3/validator/blocks/{slot}` during Capella and part of spec release for Deneb.

Implementations should be able to safely remove the apis after Electra fork.


rkapka (approved):

rolfyone:
I do agree with the sentiment, but we probably want to hold off merging this until electra or they'll no longer be listed...

I know at least teku is still calling v2 in several deneb releases, so if other clients are not publishing those endpoints that could be a potential problem.

I guess equally if we said 2.5.0 is what is expected in deneb, this would be able to be merged... We don't really have a great way of articulating that maybe?

tersec:
In favor of removing.

rolfyone:
As discussed on discord, if they were deprecated for the deneb release we can remove in electra. Updated tag.

rolfyone (approved):
LGTM

PR #472, opened 2024-09-16
Update submitPoolAttestationsV2 endpoint

eserilev:
If the decision is made to introduce a `SingleAttestation` object via https://github.com/ethereum/consensus-specs/pull/3900 it may make more sense for `submitPoolAttestationsV2` to accept a list of `SingleAttestations` instead of a list of `Attestation` objects.


apis/beacon/pool/attestations.v2.yaml
```diff
@@ -79,10 +79,7 @@ post:
           anyOf:
             - type: array
               items:
-                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
```
nflaig, line 82: so the idea is to always submit SingleAttestation even pre-Electra? in this case should remove the `anyOf`

types/attestation.yaml
```diff
@@ -15,6 +15,23 @@ IndexedAttestation:
       $ref: './primitive.yaml#/Signature'
       description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
 
+SingleAttestation:
+  type: object
+  description: "The [`SingleAttestation`](https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/electra/beacon-chain.md#singleattestation) object from the CL spec."
```
nflaig, line 20: this results in a 404, need to wait for upstream PR to be part of a pre-release, or update it later on

nflaig (commented):
When submitting an attestation via api or receiving one on p2p we also emit an event. We could consider adding a new `single_attestation` event to accommodate https://github.com/ethereum/consensus-specs/pull/3900 or convert to Attestation and use existing event, for backward compatibility, we would likely need to emit both events.

https://github.com/ethereum/beacon-APIs/blob/ca2f9a8cbff300f458bc27d72bb0e4152936b318/apis/eventstream/index.yaml#L64-L65

apis/beacon/pool/attestations.v2.yaml
eserilev: removed

The idea here is that this endpoint should only be used post electra. Since the v2 endpoint is fairly new I thought it'd be ok to repurpose it for post electra usage. Lighthouse is only using the v2 endpoint post electra, but maybe other clients would prefer a v3 endpoint?

Scutua (approved):

dapplion:
I want to echo @eserilev concerns of the non-simplicity of this change here too, see
- https://github.com/ethereum/consensus-specs/pull/3900#issuecomment-2353688733

apis/beacon/pool/attestations.v2.yaml
nflaig: I don't think we need a v3, this endpoint is effectively only used in devnets right now and would be updated for the next devnet and I don't think anyone is running mixed client setups yet.

In Lodestar we also start using the attestation v2 apis post-electra but what I meant is that after electra is live on mainnet we can be sure that all nodes implement the v2 apis and the v1 apis become unusable since those only support phase0 attestations, this means we could completely remove them from the codebase. We still wanna support earlier forks though, e.g. our sim tests do a fork transitions from phase0 to latest fork, but for that the v2 apis need to be backward compatible.

We should clarify what type of attestation should be submitted pre- and post-electra, submitting `SingleAttestation` should also work for earlier forks, so I don't think this is an issue.

nflaig: Since https://github.com/ethereum/consensus-specs/pull/3900 now also updates the honest validator spec, wouldn't it make the most sense to submit whatever attestation format the validator works with, i.e. `Attestation` for pre-electra and `SingleAttestation` post-electra. This means we don't need to do any transformation before gossiping the attestations.

types/attestation.yaml
```diff
@@ -15,6 +15,23 @@ IndexedAttestation:
       $ref: './primitive.yaml#/Signature'
       description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
 
+SingleAttestation:
```
nflaig, line 18: we might wanna put this into electra namespace as no namespace types are either phase0 or unspec'd types

Mahmoudhpp (approved):

apis/eventstream/index.yaml
```diff
@@ -66,6 +67,11 @@ get:
               value: |
                 event: attestation
                 data: {"aggregation_bits":"0x01", "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505", "data":{"slot":"1", "index":"1", "beacon_block_root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "source":{"epoch":"1", "root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"}, "target":{"epoch":"1", "root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"}}}
+            single_attestation:
```
arnetheduck, line 70: I wonder if it's not better to simply add the validator index as a stand-alone field here, so that both index and committee list are available to consumers - ideally, this would be introduced in a way that doesn't require downstream tooling to change.

nflaig: > ideally, this would be introduced in a way that doesn't require downstream tooling to change.

No matter if we go with https://github.com/ethereum/consensus-specs/pull/3900 or not, there will be an impact on downstream consumers since the `Attestation` is modified in electra.

For `single_attestation` it would be best to start emitting it even pre-electra (by converting `phase0.Attestation` to `SingleAttestation`) so that consumers can switch to the new topic earlier.

Overall, it seems like the cleanest solution would be to also go with https://github.com/ethereum/consensus-specs/pull/3787 and `AggregatedAttestation` (as noted in https://github.com/ethereum/consensus-specs/pull/3900#issuecomment-2383952335). We could then add two new topics `onchain_attestation` and `aggregate_attestation` and completely deprecate `attestation` post-electra.

This seems cleaner from a consumer point of view as well, you can always subscribe to multiple topics if you want to track all kind of attestations but as it is right now, the `attestation` topic always emits all 3 different types of attestations which is not great imo.

apis/eventstream/index.yaml
```diff
@@ -66,6 +67,11 @@ get:
...
+              description: The node has received a SingleAttestation (from P2P or API) that passes validation rules of the `beacon_attestation_{subnet_id}` topic
+              value: |
+                event: single_attestation
+                data: {"committee_index":"1", "attester_index": "1", "signature":"0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505", "data":{"slot":"1", "index":"1", "beacon_block_root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "source":{"epoch":"1", "root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"}, "target":{"epoch":"1", "root":"0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"}}}
```
nflaig, line 74: It would be better to put the `signature` last to match [spec container](https://github.com/ethereum/consensus-specs/pull/3900/files#diff-ecc9c1ba13f4c2bf31208a56061f23431b62a644cbb6db6ea1480a32976f2fe3R264-R269)

apis/eventstream/index.yaml
eserilev: As far as I understand it, we need access to the state in order to convert between `SingleAttestation` and `Attestation`. So emitting an `Attestation` event for this new endpoint seems to add complexity

nflaig: > access to the state in order to convert between SingleAttestation to Attestation

you only need access to shuffling but this is already required to validate `Attestation`

types/electra/attestation.yaml
```diff
@@ -15,7 +15,22 @@ Electra:
       signature:
         $ref: '../primitive.yaml#/Signature'
         description: "The BLS signature of the `IndexedAttestation`, created by the validator of the attestation."
-
+  SingleAttestation:
+    type: object
+    description: "The [`SingleAttestation`](https://github.com/ethereum/consensus-specs/blob/v1.3.0/specs/electra/beacon-chain.md#singleattestation) object from the CL spec."
```
nflaig, line 20:
```suggestion
    description: "The [`SingleAttestation`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-alpha.9/specs/electra/beacon-chain.md#singleattestation) object from the CL spec."
```

nflaig (approved):
LGTM

Mahmoudhpp (approved):

rkapka (approved):
LGTM

rolfyone (approved):
LGTM

michaelsproul (approved):
LGTM

PR #477, opened 2024-10-08
Update blinded block to contain execution requests instead of root

nflaig:
Latest changes from https://github.com/ethereum/builder-specs/pull/101 pass `execution_requests` instead of just the root in the blinded block, see PR for rationale.


ralexstokes (approved):
we should do this

ensi321 (approved):

mcdee (approved):

PR #462, opened 2024-08-05
Update CI tooling to latest versions

nflaig:
Update github actions / workflows to use latest versions and swaps out deprecated actions with maintained ones. This also simplifies a few upload / bundling steps.

I have tested all workflows on a forked repo, everything seems to work as before.

PR #486, opened 2024-12-12
chore(ci): update actions/upload|download-artifact from v3 to v4

skylenet:
Due to deprecation notice: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -27,7 +27,7 @@ jobs:
           file: "beacon-node-oapi.json"
           target: "deploy/releases/${{ matrix.release }}/beacon-node-oapi.json"
       - name: Save releases (artifact)
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: releases
           retention-days: 7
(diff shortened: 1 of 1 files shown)


nflaig:
Just bumping the version is not sufficient, we need to adapt the workflows to work with v4 as there are behavior changes.

Last time I tried upgrading to v4 was in https://github.com/ethereum/beacon-APIs/pull/462 but I didn't get it to work and had done a lot of other changes so I skipped upgrading those. They changed how files are copied and how to deal with existing directories, iirc it throws an error if a directory already exists. There is a `overwrite` flag (see [inputs](https://github.com/actions/upload-artifact?tab=readme-ov-file#inputs)) but the problem there was it would overwrite the whole folder and since we run this as a matrx it would only keep the last release.
https://github.com/ethereum/beacon-APIs/blob/1cf1adae5bd944108fb6522000d2d392967729ba/.github/workflows/deploy.yaml#L21

There is probably a way to tweak this (ie. by using different folders) to make it work but this needs to be tested in a forked repo first.

nflaig:
https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md might be helpful

nflaig:
I opened a PR https://github.com/ethereum/beacon-APIs/pull/487 which updates the workflows to work with v4

PR #441, opened 2024-04-11
Add metadata fields to getBlobSidecars response

nflaig:
As noted in https://github.com/ethereum/beacon-APIs/issues/439 and after further [discussion on discord](https://discord.com/channels/595666850260713488/1227937241427611658), the rough consensus seems to be to add those metadata fields without bumping the api version, but to make them optional for now until all clients have updated their implementation.

Closes https://github.com/ethereum/beacon-APIs/issues/439

PR #488, opened 2025-01-21
Add Electra blob sidecars type to support max blobs from EIP-7691

nflaig:
Adds new `BlobSidecars` type to Electra with max items of 9 to match `MAX_BLOBS_PER_BLOCK_ELECTRA` from EIP-7691.

In addition this PR makes the metadata fields added in https://github.com/ethereum/beacon-APIs/pull/441 and the `Eth-Consensus-Version` header required. Implementations should make sure to support this at latest in their Electra release.


tersec (approved):

mcdee:
The change to the existing endpoint to now require `Eth-Consensus-Version` is technically breaking, it's worth doing a quick check of the endpoint on existing beacon nodes to confirm if they supply this, and if not to raise an issue so that they are aware of this and can address it ahead of time.

But in general LGTM.

nflaig:
> The change to the existing endpoint to now require Eth-Consensus-Version is technically breaking, it's worth doing a quick check of the endpoint on existing beacon nodes to confirm if they supply this

I updated the PR description to highlight this, Lodestar already sets this header in the stable release.

mcdee:
Sure, but it's still a breaking change to an existing endpoint which is unexpected and may not be picked up by client teams without a nudge.

nflaig:
From references of the previous PR, looks like at least Prysm and Teku should be fine as well, so there is only Nimbus, Lighthouse and Grandine left to check

rkapka (approved):
Prysm does already return the metadata

Tumas:
Grandine returns the metadata

rolfyone (approved):
LGTM

types/electra/blob_sidecar.yaml
```diff
@@ -0,0 +1,7 @@
+Electra:
+  BlobSidecars:
+    type: array
+    items:
+      $ref: "../deneb/blob_sidecar.yaml#/Deneb/BlobSidecar"
+    minItems: 0
+    maxItems: 9
```
dapplion, line 7: This API returns JSON, why is it relevant this maxItems value? Even for SSZ, in it's serialized form the max value is not relevant

nflaig: It returns both JSON and SSZ
https://github.com/ethereum/beacon-APIs/blob/aa1be259e2d8a64f451dd304913e0310779f6f80/apis/beacon/blob_sidecars/blob_sidecars.yaml#L50-L52

> Even for SSZ, in it's serialized form the max value is not relevant

I quickly mentioned this [in discord](https://discord.com/channels/595666850260713488/710831784991916072/1331233019519832065), we could just use `MAX_BLOB_COMMITMENTS_PER_BLOCK` which is 4096 as max items which is what we use in Lodestar for the list limit.

tersec: It seems better to discard invalid data as quickly as possible, and anything past `MAX_BLOBS_PER_BLOCK_ELECTRA` is invalid. No need to build in capacity in the rest of the system to handle it, either on the client or server end, for theoretical possibilities which even if working in isolation aren't useful as part of a broader system.

dapplion: You are talking about a consumer of the REST API, not a p2p network with adversarial peers. I think `MAX_BLOB_COMMITMENTS_PER_BLOCK` is fine and we don't have to deal with forks on this type anymore

tersec: Unclear that https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/getBlobSidecars will even be useful/salient in Fulu; the beacon API might end up with a new `getColumnSidecars` or similar endpoint. A similar discussion has been happening in the req/resp area, where the v3 Fulu-only getBlobsByRange was removed in lieu of keeping only the column one.

rolfyone:
@tersec @dapplion is the above conversation resolved? I don't want to merge until there's consensus...

james-prysm (approved):

types/electra/blob_sidecar.yaml
nflaig: we can always change this to `MAX_BLOB_COMMITMENTS_PER_BLOCK` in a follow-up PR, I don't feel strongly about this as both approaches work

nflaig:
> @tersec @dapplion is the above conversation resolved? I don't want to merge until there's consensus...

can be changed in a follow-up PR if someone feels strongly about it, discussion has stopped for a week now

PR #489, opened 2025-01-22
Add `custody_group_count` to metadata in `node/identity` response

jimmygchen:
Add `custody_group_count` to metadata.

This field is part of `MetaDataV3` used by PeerDAS and will be available from the Fulu fork.

https://github.com/ethereum/consensus-specs/blob/e79ef816aef05e34eff0b53311cf50c13fb67fc9/specs/fulu/p2p-interface.md?plain=1#L159

### DIFF
--- a/types/p2p.yaml
+++ b/types/p2p.yaml
@@ -36,6 +36,9 @@ MetaData:
       description: "Bitvector representing the node's sync committee subnet subscriptions. This metadata is not present in phase0, but will be present in Altair."
       $ref: "./primitive.yaml#/Bitvector"
       example: "0x0f"
+    custody_group_count:
+      description: "Uint64 representing the node's custody group count. The metadata is present from the Fulu fork."
+      $ref: "./primitive.yaml#/Uint64"
 
 Peer:
   type: object


nalepae (approved):

nflaig (dismissed):
I guess we did the same for `syncnets` in https://github.com/ethereum/beacon-APIs/pull/150 to add it as an optional field.

The problem with this approach is that it makes returning a ssz response more complicated, although as per spec this api only supports json.

I would be in favor of bumping the api version if we add new properties to `data` object but this seems reasonable as well since we used the same approach before.

michaelsproul (approved):

nflaig (approved):

rolfyone (approved):
LGTM

dapplion:
> The problem with this approach is that it makes returning a ssz response more complicated, although as per spec this api only supports json.

For SSZ you could set both values to zero

nflaig:
> For SSZ you could set both values to zero if pre-v3

That works if all clients do this but without version bump of the api you don't really know what to expect and might get different result based on client version.

My comment was more general that we shouldn't add optional fields, `/node` apis already don't work well with SSZ anyways.

agnxsh (approved):
lgtm

PR #490, opened 2025-01-27
Add EIP-7805 (FOCIL) endpoints

ensi321:


types/validator.yaml
```diff
@@ -111,3 +111,20 @@ SignedAggregateAndProof:
...
+      $ref: "./primitive.yaml#/Pubkey"
+    validator_index:
+      $ref: "./primitive.yaml#/Uint64"
+      description: "Index of validator in validator registry."
+    slot:
+      $ref: "./primitive.yaml#/Uint64"
+      description: "The slot at which the validator must propose an inclusion list." 
+    inclusion_list_committee_root:
+      $ref: "./primitive.yaml#/Root"
+      description: "The root of inclusion list committee that the duty is part of"
```
nflaig, line 130: might wanna rephrase to

```suggestion
      description: "The root of inclusion list committee that the validator is part of." 
```

types/focil/inclusion_list.yaml
```diff
@@ -0,0 +1,24 @@
+Focil:
+  InclusionList:
```
rolfyone, line 2: we'll want a fork name rather than a EIP name for the api, but it's probably premature to know that I guess...

nflaig: I think for now we should keep it as EIP name (or number) but it would be good if we can merge new endpoints before those are officially part of a hard fork. I added a new tag "Draft" here to separate those in the api explorer, but maybe even a EIP specific tag would make sense.

rolfyone: In teku i just use `Experimental` tag but draft also works, that'd be fine...

nflaig: That's what I [was using](https://github.com/ethereum/beacon-APIs/pull/490/commits/b82eaf152555e4717f8e721adb9c372ae296da18) but it had a different purpose previously when it was used [for the rewards apis](https://github.com/ethereum/beacon-APIs/pull/426/files#diff-962ea34dc6ec0e65d4e1d6c8682e97c547e6e0d2f1168925210ff7ee47ef6ce0L62). I think either one works, if we do this for more EIPs it might even be good to use a separate tag for each. This would allow to disable spec tests based on tag

rolfyone: its been super problematic merging with feature flags sometimes, that's my hesitation.

nflaig: Yeah we probably won't be merging this any time soon since the CL spec isn't even merged yet but it would be nice if we can come up with an approach to allow merging this before it's confirmed as part of a hard fork.

I pushed a few changes group files / types under EIP specific namespace and used the EIP as tag as well, this would allow us to have more than one EIP in "Draft" stage and be part of the api explorer.

apis/validator/inclusion_list.yaml
```diff
@@ -0,0 +1,62 @@
...
+      description: Success response
+      content:
+        application/json:
+          schema:
+            title: ProduceInclusionListResponse
+            type: object
+            required: [data]
+            properties:
+              data:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Bellatrix.Transactions"
```
gfukushima, line 25: is this meant to return `EIP7805.InclusionList`? maybe I'm reading things wrong but this seems to expect the transactions of the IL in the data property whereas description says this will request BN to produce an IL.

nflaig: depends how you define what a inclusion list is, technically it's just a list of transactions which is the return type of this api and also the engine-api

the other return type could be `InclusionList` (ie. wrap it in container) but this would return useless and even confusing data, eg. `validator_index` isn't even set by beacon node, so beacon node would just set a random value (maybe `0`) and then the validator client is responsible for overriding it

we can think about if we wanna wrap this in a container, and maybe even make the api fork-aware by adding version metadata, this might be better for future forward compatibility

ensi321: > is this meant to return `EIP7805.InclusionList`? maybe I'm reading things wrong but this seems to expect the transactions of the IL in the data property whereas description says this will request BN to produce an IL.

Yes. As @nflaig mentioned, IL can mean 
1) A list of IL transactions. It is used here, [execution api spec](https://github.com/ethereum/execution-apis/blob/e70099c8408affe43bf227d76ce9c33ec9ff6882/src/engine/experimental/eip7805.md#engine_getinclusionlistv1) and anything as far as EL is concerned 
2) IL container as defined in consensus specs. Basically 1 + any other useful consensus info


Here we only return IL transactions because any additional consensus info (slot, validator index, ILC root) are all available in VC. No need to grab them from BN

jihoonsong:
@ensi321 Does this PR require some updates for Heze rebase?

apis/eventstream/index.yaml
```diff
@@ -147,7 +148,8 @@ get:
...
+                data: {"version": "eip7805", "data": {"proposer_index": "123", "proposal_slot": "10", "parent_block_number": "9", "parent_block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "parent_block_hash": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "payload_attributes": {"timestamp": "123456", "prev_randao": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "suggested_fee_recipient": "0x0000000000000000000000000000000000000000", "withdrawals": [{"index": "5", "validator_index": "10", "address": "0x0000000000000000000000000000000000000000", "amount": "15640"}], "parent_beacon_block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "inclusion_list_transactions": ["0x03f88f0780843b9aca008506fc23ac00830186a09400000000000000000000000000000000000001008080c001e1a0010657f37554c781402a22917dee2f75def7ab966d7b770905398eba3c44401401a0840650aa8f74d2b07f40067dc33b715078d73422f01da17abdbd11e02bbdfda9a04b2260f6022bf53eadb337b3e59514936f7317d872defb891a708ee279bdca90"]}}}
```
bharath-123: Can we also return the `inclusion_list_bits`? would it make sense to return that for the builder to add the correct list of bits in the bid

nflaig: why do you need that, you have literally the `inclusion_list_transactions`, aren't the `inclusion_list_bits` in the bid supposed to signal which ILs your payload has/will satisfy? cc @jihoonsong

jihoonsong: A builder listens to this event for their payload building and constructs a bid. So the builder expects to have all the information they need for a bid from this event.

However, `inclusion_list_bits` indicates which *ILs* a builder considered. This is what the builder cannot derive from `inclusion_list_transactions`.

Correct me if I'm wrong @bharath-123

nflaig: > why do you need that, you have literally the inclusion_list_transactions

I got confused, this is not a list of inclusion lists, it's a deduplicated list of transactions the payload needs to include to satisfy the ILs, so you cannot derive `inclusion_list_bits` from that

bharath-123: the `inclusion_list_bits` are the indices of the commitee members whose ILs were considered. its hard for a builder to derive that from just the txs.

jihoonsong: Following the discussion in [FOCIL breakout #36](https://github.com/ethereum/pm/issues/2121#issuecomment-4732268993), I've opened a [PR](https://github.com/ethereum/consensus-specs/pull/5371) that removes the IL bitlist from bid. cc @bharath-123

types/duty.yaml
```diff
@@ -65,3 +65,20 @@ PtcDuty:
...
+        $ref: "./primitive.yaml#/Pubkey"
+      validator_index:
+        $ref: "./primitive.yaml#/Uint64"
+        description: "Index of validator in validator registry."
+      slot:
+        $ref: "./primitive.yaml#/Uint64"
+        description: "The slot at which the validator must propose an inclusion list."
+      inclusion_list_committee_root:
+        $ref: "./primitive.yaml#/Root"
+        description: "The root of inclusion list committee that the validator is part of."
```
conache, line 84: Now that [consensus-specs#5513](https://github.com/ethereum/consensus-specs/pull/5513) replaced `inclusion_list_committee_root` with `dependent_root` on the `InclusionList` container, and [consensus-specs#5544](https://github.com/ethereum/consensus-specs/pull/5544) removed committee roots from the IL flow entirely, I think that `inclusion_list_committee_root` can also be dropped here

types/heze/inclusion_list.yaml
```diff
@@ -0,0 +1,24 @@
...
+    type: object
+    description: "The [`InclusionList`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/heze/beacon-chain.md#inclusionlist) object from the CL Heze spec."
+    required: [slot, validator_index, inclusion_list_committee_root, transactions]
+    properties:
+      slot:
+        $ref: "../primitive.yaml#/Uint64"
+      validator_index:
+        $ref: "../primitive.yaml#/Uint64"
+      inclusion_list_committee_root:
+        $ref: "../primitive.yaml#/Root"
```
conache, line 12:
```suggestion
    required: [slot, validator_index, dependent_root, transactions]
    properties:
      slot:
        $ref: "../primitive.yaml#/Uint64"
      validator_index:
        $ref: "../primitive.yaml#/Uint64"
      dependent_root:
        $ref: "../primitive.yaml#/Root"
```
Aligning with the change added in [consensus-specs#5513](https://github.com/ethereum/consensus-specs/pull/5513), which replaced this field on the `InclusionList` container

Issue #393, opened 2023-12-14
Invalid schema


jeluard:
If you look at the bottom right corner of the deployed [doc](https://ethereum.github.io/beacon-APIs/), an `INVALID` link shows up.

Browsing this [link](https://validator.swagger.io/validator/debug?url=https%3A%2F%2Fethereum.github.io%2Fbeacon-APIs%2Freleases%2Fv2.4.2%2Fbeacon-node-oapi.json) shows up a bunch of errors.
Is this known? Are we interested in fixing those?

rolfyone:
Last time this happened I did a sweep of fixes, I'm happy if someone wants to fix them :)

nflaig:
Resolved by https://github.com/ethereum/beacon-APIs/pull/421

PR #492, opened 2025-01-31
Change anyOf to oneOf for attestation submissions.

mcdee:
Add clarifying text as to when to supply `Attestation` and when `SingleAttestation`

### DIFF
--- a/apis/beacon/pool/attestations.v2.yaml
+++ b/apis/beacon/pool/attestations.v2.yaml
@@ -33,7 +33,7 @@ get:
                 enum: [phase0, altair, bellatrix, capella, deneb, electra]
                 example: "phase0"
               data:
-                anyOf:
+                oneOf:
                   - type: array
                     items:
                       $ref: '../../../beacon-node-oapi.yaml#/components/schemas/Attestation'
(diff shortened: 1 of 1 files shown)


apis/beacon/pool/attestations.v2.yaml
```diff
@@ -33,7 +33,7 @@ get:
                 enum: [phase0, altair, bellatrix, capella, deneb, electra]
                 example: "phase0"
               data:
-                anyOf:
+                oneOf:
```
nflaig, line 36: I don't really see the difference between any and one of, it's mostly semantics but the usage of `oneOf` is problematic as it can cause an invalid oapi spec, see https://github.com/ethereum/beacon-APIs/pull/421 and related issue https://github.com/ethereum/beacon-APIs/issues/393.

The issue is that if you pass a value that matches multiple of the schemas then it fails when `oneOf` is used while it passes with `anyOf`. This causes false positives when running api spec tests and some schema parsers don't accept invalid oapi specs so we had to patch it locally before running our tests previously.

Now in this specific case I think it's not a issue since the both types are different enough and `phase0.Attestation` will not fit the schema of `SingleAttestation` (and vice versa) but this is something to keep in mind.

rolfyone: To me this is just documentation, so if a change to oneOf is a problem I'm fine with leaving it as anyOf. neither really enforces `at slot X , type Y must be used`.

mcdee: The main part of this is the clarifying text, if people want to keep it as `anyOf` then fair enough, although as mentioned there isn't an overlap situation here so `anyOf` should work.

nflaig: I tested usage of `oneOf` in our spec tests and as expected it does not cause issues due to the difference in schemas, I am ok with keeping this change as it's seem to help in terms of semantics here

nflaig (approved):
LGTM - but might wanna rephrase the PR title a little

PR #494, opened 2025-02-04
Deprecate deposit snapshot api

nflaig:
Due to changes introduced by [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110) this api will no longer be relevant after the Electra hard fork once `state.eth1_deposit_index == state.deposit_requests_start_index` is reached.

Deprecating this api as part of the Electra release will allow us to remove it in the Fulu release.


tersec (approved):

philknows (approved):

rolfyone (dismissed):
LGTM

rolfyone:
the one thing is we should probably put in changelog? can then list it'll be removed for fulu fork release..

nflaig:
> the one thing is we should probably put in changelog? can then list it'll be removed for fulu fork release..

good idea to get more visibility on this, we've missed some more changelog entries https://github.com/ethereum/beacon-APIs/pull/496

rolfyone (approved):

PR #496, opened 2025-02-06
Add changelog entries for deprecated v1 block publishing apis

nflaig:
We've deprecated those in https://github.com/ethereum/beacon-APIs/pull/467 but as suggested by @rolfyone it makes sense to also note this in the changelog.


rolfyone (approved):
LGTM

PR #501, opened 2025-02-11
Show most recent fork as default example in api explorer

nflaig:
There was [some discussion](https://discord.com/channels/595666850260713488/710831784991916072/1337447612273655808) related to this on discord that right now we use phase0 as default fork in examples and also render the oldest fork type in preview in api explorer.

This PR changes the order of types in `anyOf` to list the most current fork first as it's more relevant for users / developers looking at the explorer.

Technically these changes don't make a difference, eg. when running spec tests or similar.


mcdee (approved):

tersec (approved):

rolfyone (approved):
LGTM

PR #503, opened 2025-02-14
Apply oneOf to submit attestations api as intended

nflaig:
The intend of https://github.com/ethereum/beacon-APIs/pull/492 was to apply `oneOf` to `submitPoolAttestationsV2` but it was applied to `getPoolAttestationsV2` which causes issues as outlined here https://github.com/ethereum/beacon-APIs/pull/492#discussion_r1937508072.

Alternative would just be to use `anyOf` everywhere


mcdee:
Oops, thanks for the catch.

rolfyone (approved):
LGTM

PR #504, opened 2025-02-14
Update attestation event example for electra

nflaig:
Added `committee_bits` and reordered fields to match [Attestation](https://github.com/ethereum/consensus-specs/blob/a9e40ace97aa39ff9756bf88e27559d16697acb4/specs/electra/beacon-chain.md?plain=1#L350-L354) container from electra spec.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -66,7 +66,7 @@ get:
               description: The node has received an Attestation (from P2P or API) that passes validation rules of the `beacon_attestation_{subnet_id}` topic
               value: |
                 event: attestation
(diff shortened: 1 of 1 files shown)


apis/eventstream/index.yaml
```diff
@@ -66,7 +66,7 @@ get:
               description: The node has received an Attestation (from P2P or API) that passes validation rules of the `beacon_attestation_{subnet_id}` topic
```
nflaig, line 66: Once electra goes live, Lodestar will only emit this event for aggregate attestations either received via `beacon_aggregate_and_proof` gossip topic or the attestations part of the block when importing.

Curious how other client use the event but the description might be misleading since for unaggregated (single) attestations we have `single_attestation` now.

ensi321: The description mentions `Attestation` container which implies aggregated attestations and block attestations and excludes single attestations. So it is kind of unambiguous in my opinion.

ensi321 (approved):

rolfyone (approved):
LGTM

PR #505, opened 2025-02-14
Added viewport meta

dinitheth:
**Add viewport meta tag for mobile support and ensures the page scales correctly on mobile devices.**


nflaig (commented):
Thanks for the PR! could you please share some images how this renders before and after your change on mobile?

PR #507, opened 2025-02-18
Run CI build step if commit is pushed to release branch

nflaig:
Noticed CI build step was not triggered when the release workflow opened the PR here https://github.com/ethereum/beacon-APIs/pull/506 but we need that job to succeed to pass required merge checks.

This PR should solve this by running the job when the release commit is pushed to the branch.


rolfyone (approved):
LGTM

PR #508, opened 2025-02-23
Add SSZ support to register validator api

nflaig:
To accommodate https://github.com/ethereum/builder-specs/pull/110, since the endpoint is mostly just a pass through of the request from the validator client to the builder, this allows to skip an extra serialization step and overall should make it more seamless if SSZ is used.


rolfyone (approved):
LGTM

ralexstokes (approved):

PR #511, opened 2025-03-05
require that Electra produceAttestationData AttestationData have 0 committee_index

tersec:
Currently, it's ambiguous whether a VC which specified a non-zero `committee_index` in Electra would get an `AttestationData` in return. This is a potential source of VC/BN Electra [interoperability issues](https://github.com/ethpandaops/ethereum-package?tab=readme-ov-file#beacon-node--validator-client-compatibility), as each VC or BN might rely on the other element of its pair for that client to enforce the 0-constraint. Ensure that the BN, at least, does so.

### DIFF
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -4,6 +4,7 @@ get:
     - Validator
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
+  deprecated: true
   description: |
     Requests that the beacon node produce an AttestationData. For `slot`s in
     Electra and later, this AttestationData must have a `committee_index` of 0.
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -5,7 +5,8 @@ get:
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
   description: |
-    Requests that the beacon node produce an AttestationData.
+    Requests that the beacon node produce an AttestationData. For `slot`s in
+    Electra and later, this AttestationData must have a `committee_index` of 0.
 
     A 503 error must be returned if the block identified by the response
     `beacon_block_root` is optimistic (i.e. the attestation attests to a block
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -4,7 +4,6 @@ get:
     - Validator
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
-  deprecated: true
   description: |
     Requests that the beacon node produce an AttestationData. For `slot`s in
     Electra and later, this AttestationData must have a `committee_index` of 0.


nflaig (dismissed):
LGTM

nflaig:
> I'd be inclined to say "no" to requiring another version. We can make it optional now in the API spec with a note saying it is ignored, and then ignore it in the beacon nodes. Come fulu we can drop the parameter entirely from the spec, and if anyone continues to send it it's a relic that isn't used.

yes, I think we can do that but we should signal it now and beacon node can mark it as optional once Electra is live on mainnet

> Unless beacon nodes are actively checking for unsupported parameters and erroring on them, this should be a path to removal without needing us to bump the version

Right now, at least in Lodestar this is marked as required so we would error if the `committee_index` is omitted

tersec:
> > Unless beacon nodes are actively checking for unsupported parameters and erroring on them, this should be a path to removal without needing us to bump the version
> 
> Right now, at least in Lodestar this is marked as required so we would error if the `committee_index` is omitted

Similarly with Nimbus.

nflaig:
so do we just deprecate it as part of this PR?

tersec:
> so do we just deprecate it as part of this PR?

https://github.com/ethereum/beacon-APIs/pull/511/commits/a3a90c9175a77c6c277efe30cd2c993ba4e17f1a

Reasonable, sure.

One general caution I have a bit -- there are two my mind two relatively distinct tasks here:
- disambiguate and document a current behavior, compatible with current Electra BNs (no randomly removing parameters), which allows for interoperability across all VCs and BNs; and
- cleaning up from the beacon API this now-pointless parameter.

I'm ok with the "ignored" terminology as such, but not the "optional" part, for this particular endpoint. And in general, it seems useful to address the first bullet-point regardless of if there's agreement on the second. So if we all are okay with some particular approach to the second bullet-point, that's fine, that can be in this PR too, but I'd rather have the specs clearly say, right now, what an Electra VC and BN should do. Fusaka can wait, if it needs to.

apis/validator/attestation_data.yaml
```diff
@@ -4,8 +4,10 @@ get:
     - Validator
   operationId: "produceAttestationData"
   summary: "Produce an attestation data"
+  deprecated: true
```
nflaig, line 7: this should be on the `committee_index` query parameter

nflaig:
might be a good idea to also add a changelog entry for deprecating the `committee_index` property

apis/validator/attestation_data.yaml
tersec: Should it? This is basically @mcdee's idea, I think, to deprecate the parameter, and I already articulated considerations regarding that.

I don't actually think the endpoint should be kept with the same name but with one fewer parameter, so I'm not sure that it makes sense to deprecate a parameter as such.

I'm not strongly opposed, just so long as it's clear that it still has to be passed to the `produceAttestationData` endpoint regardless of how useless it is in Electra, but that's not how I'd interpret a per-parameter deprecation to begin with.

mcdee: Given that some beacon nodes currently require the parameter we can't remove it, but we can deprecate it so that it is clear that its value is ignored by the beacon node receiving the request.  Then over the lifespan of electra beacon nodes can stop making the parameter required, move to ignoring the parameter entirely, and then it can be removed from the spec at fulu with no ill effects.

But yes, I think the point here is that the tag appears to be in the incorrect place in the doc.

rolfyone:
Technically we'd normally avoid a breaking change like this in the API. are we sure we want to actually remove committee_index? this is going to potentially cause issues on an api marked 'validator required' so we want to be super careful here...
Basically any time we test from then on, we'd potentially be wanting the old api with this parameter prior to electra, so really what we're saying is the api no longer supports attestation production prior to electra...

We'd want to make sure
 - noone is ever running a network prior to electra again (real or test network)
 - teams are able to start at an electra genesis
before we considered removal if we went in this direction.

This may be a case where we should have created an ELECTRA interface and kept the old one indefinitely....

The cleanest way out of this is either to leave the field and state either (0) or ignored after electra, or to set this API only supporting up to deneb, and have a new API for post deneb. I don't think if we did the latter that it would free us up to delete this version of the api due to us needing to probably have networks prior to electra, and practically every team REQUIRING the api to perform duties.

apis/validator/attestation_data.yaml
nflaig: I think the approach outlined by Jim makes the most sense, if we wanted to go with a v2 api that removes the query parameters already now we should have done that as part of the attestation api v2 changes but personally I didn't feel strongly about doing that because technically it's not required, the beacon node can just ignore the `committee_index` in the query parameter which is what we / most bns do already.

I think deprecating the `committee_index` now and removing it in Fulu will probably be fine but I am still wondering if it's worth to do this clean up at all, seems like a nit to me

nflaig:
> we'd potentially be wanting the old api with this parameter prior to electra, so really what we're saying is the api no longer supports attestation production prior to electra...

this is a good point, we need to make sure the api works pre-electra... so based on that I think it would be for the best if we just deprecate the `committee_index` and have the note already part of this PR to make sure all bns properly handle it

nflaig:
should we just move ahead and mark `committee_index` as deprecated to signal that it is ignored after Electra? Together with the note added in this PR I think this would be good as is then

nflaig (approved):
LGTM - as [discussed on discord](https://discord.com/channels/595666850260713488/710831784991916072/1353666178563833940) we don't wanna deprecate the endpoint and neither the `committee_index` query parameter as the note seems sufficient, this could be reconsidered in the future.

Validator clients must keep sending the `committee_index` even post-electra even though it will be unused and always set to 0 in the `AttestationData` by the beacon node.

Issue #476, opened 2024-10-08
Validators API needs filtering


tdenisenko:
When calling `/eth/v1/beacon/states/head/validators?status=active` over 1 gigabyte of data is returned. We need to be able to get only the fields requested for various reasons. Such as only the indexes of the active validators which would reduce the size of the data returned significantly (~4 KB).

Proposal:

Allow filtering by fields, such as `/eth/v1/beacon/states/head/validators?status=active&fields=index,balance`

nflaig:
A new endpoint has been added recently to provide a minimal response `/eth/v1/beacon/states/{state_id}/validator_identities` (see [postStateValidatorIdentities](https://ethereum.github.io/beacon-APIs/?urls.primaryName=dev#/Beacon/postStateValidatorIdentities)) which gives you just the index and public key. It's not widely supported yet though, Lodestar will have it in the next release (v1.23.0)

> Allow filtering by fields

This is problematic as we want to support both JSON and SSZ responses, in case of SSZ it requires to predefine a container, a generic approach to select fields dynamically would be difficult to apply. We also try to return data structures that are defined by the CL spec.

Curious what others think.

PR #512, opened 2025-03-14
Electra state pending consolidations

rolfyone:
We had previously added pendingDeposits and pendingWithdrawals but this was missing, and I realised when I had to get pending consolidations from the debug/state, which is kind of messy.

Copied pendingDeposits, hopefully i caught any copy/paste issues!


mcdee:
lgtm

(changelog?)

nflaig (commented):
Looks good, I thought you intentionally omitted this previously because nobody was asking for it, as Jim noted, changelog entry would be good to have, we can also consider cutting a 3.1 release with some follow up changes for Electra

vasco0419:
hello
I'm a seasoned blockchain developer and game fi developer.
I wanna work with you.
what should I do?

rolfyone:
> Looks good, I thought you intentionally omitted this previously because nobody was asking for it, as Jim noted, changelog entry would be good to have, we can also consider cutting a 3.1 release with some follow up changes for Electra

i realised when i was consolidating that i completely need it!

nflaig (approved):

arkh-consensys:
This is great. Any plan on exposing the beacon state through an API in a way that the consumers can request for partial part of the state, instead of fetching the full state object?

nflaig:
> This is great. Any plan on exposing the beacon state through an API in a way that the consumers can request for partial part of the state, instead of fetching the full state object?

I am guessing you mean something similar to https://github.com/ethereum/beacon-APIs/issues/476? I would be up for supporting something like this but it must be possible to support SSZ (see https://github.com/ethereum/beacon-APIs/issues/476#issuecomment-2399501360), I feel like a PoC in a client implementation would be the first step before having a `getStateV3` proposal on the spec side. Also probably better to open an issue with some more details on what the expected UX should be.

arkh-consensys:
A similar request, but instead of going to subfields of structs, just deciding which fields of the State is requested. I will open a separate issue for this with more details

arkh-consensys:
I created this https://github.com/ethereum/beacon-APIs/issues/513

PR #516, opened 2025-03-21
Add EIP-7594 (PeerDAS) related changes

0x00101010:


types/fulu/data_column_sidecar.yaml
```diff
@@ -0,0 +1,34 @@
+Fulu:
+  DataColumnSidecar:
+    type: object
+    description: "A data column sidecar as defined in the Fulu CL spec."
+    required:
+      [
+        column_index,
```
jtraglia, line 7: This is just `index` in the consensus specs: https://github.com/ethereum/consensus-specs/blob/ff99bc03d6da29d9ef6e055bdb8500e1b2942f1e/specs/fulu/das-core.md?plain=1#L81

types/fulu/data_column_sidecar.yaml
```diff
@@ -0,0 +1,34 @@
...
+      kzg_commitments:
+        type: array
+        items:
+          $ref: "../primitive.yaml#/Bytes"
+      kzg_proofs:
+        type: array
+        items:
+          $ref: "../primitive.yaml#/Bytes"
+      signed_block_header:
+        $ref: "../deneb/block_contents.yaml#/Deneb/SignedBeaconBlockHeader"
```
jtraglia, line 30: Why is this Deneb? I imagine this should be Fulu or Phase0. It hasn't been updated since Phase0.

0x00101010: great catch, updated

types/fulu/block.yaml
```diff
@@ -0,0 +1,127 @@
...
+        eth1_data,
+        graffiti,
+        proposer_slashings,
+        attester_slashings,
+        attestations,
+        deposits,
+        voluntary_exits,
+        sync_aggregate,
+        bls_to_execution_changes,
+        blob_kzg_commitments,
```
g11tech, line 19: execution_requests is missing from pectra

types/fulu/data_column_sidecar.yaml
```diff
@@ -0,0 +1,34 @@
+Fulu:
+  DataColumnSidecar:
```
jimmygchen, line 2: I don't see this used anywhere in the beacon-APIs, are you thinking to introduce a `getDataColumns` endpoint?

types/fulu/block.yaml
```diff
@@ -0,0 +1,127 @@
...
+      message:
+        $ref: "#/Fulu/BeaconBlock"
+      signature:
+        $ref: "../primitive.yaml#/Signature"
+
+  BlindedBeaconBlockBody:
+    # TODO: update link once confirmed.
+    description: "A variant of the [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/v1.5.0-beta.2/specs/electra/beacon-chain.md#beaconblockbody) object from the CL Fulu spec, which contains a transactions root rather than a full transactions list."
+    allOf:
+      - $ref: "#/Electra/BeaconBlockBodyCommon"
```
jimmygchen, line 98:
```suggestion
      - $ref: "#/Fulu/BeaconBlockBodyCommon"
```

types/fulu/block.yaml
```diff
@@ -0,0 +1,127 @@
...
+      - type: object
+        required: [execution_payload_header, execution_requests]
+        properties:
+          execution_payload_header:
+            $ref: "../deneb/execution_payload.yaml#/Deneb/ExecutionPayloadHeader"
```
jimmygchen, line 103:
```suggestion
            $ref: "../fulu/execution_payload.yaml#/Fulu/ExecutionPayloadHeader"
```

types/fulu/execution_payload.yaml
```diff
@@ -0,0 +1,80 @@
...
+        properties:
+          transactions:
+            $ref: "../bellatrix/transactions.yaml#/Bellatrix/Transactions"
+          withdrawals:
+            $ref: "../capella/withdrawals.yaml#/Capella/Withdrawals"
+          proof_version:
+            $ref: "../primitive.yaml#/Uint8"
+
+  ExecutionPayloadHeader:
+    description: "The [`ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/v1.4.0/specs/deneb/beacon-chain.md#executionpayloadheader) object from the CL Deneb spec."
```
jimmygchen, line 71:
```suggestion
    # TODO: update link once confirmed.
    description: "The [`ExecutionPayloadHeader`](https://github.com/ethereum/consensus-specs/blob/v1.4.0/specs/deneb/beacon-chain.md#executionpayloadheader) object from the CL Fulu spec."
```

types/fulu/block_contents.yaml
```diff
@@ -0,0 +1,24 @@
+Fulu:
+  BlockContents:
+    type: object
+    description: "The required object for block production according to the Fulu CL spec."
+    required: [block, cell_proofs, blobs]
+    properties:
+      block:
+        $ref: "./block.yaml#/Fulu/BeaconBlock"
+      cell_proofs:
```
jimmygchen, line 9: How do people feel about keeping the `kzg_proofs` field name? Not clear to me how much we gain from this rename, and given the beacon APIs are fork aware, this change doesn't feel necessary.

types/fulu/data_column_sidecar.yaml
0x00101010: Cool, seems like this is strictly api related, so I'll remove this

types/fulu/block.yaml
0x00101010: No longer needed as we are not proposing a new `ExecutionPayload` with `ProofVersion` any more. `BeaconBlock` should stay the same as Electra

types/fulu/execution_payload.yaml
0x00101010: no longer needed as there is no change to `ExecutionPayload` now

types/fulu/data_column_sidecar.yaml
nflaig: > are you thinking to introduce a getDataColumns endpoint?

this seems potentially useful but I would recommend to keep this PR to minimal required changes for peerdas to make it easier to merge

types/fulu/block_contents.yaml
```diff
@@ -0,0 +1,26 @@
+Fulu:
+  BlockContents:
+    type: object
+    description: "The required object for block production according to the Fulu CL spec."
+    required: [block, kzg_proofs, blobs]
+    properties:
+      block:
+        $ref: "../electra/block.yaml#/Electra/BeaconBlock"
+      kzg_proofs:
+        description: "These should be cell proofs of the blobs."
```
dapplion, line 10:
```suggestion
        description: "Cell proofs of the blobs as defined in EIP-7594"
```

types/fulu/block_contents.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+  SignedBlockContents:
+    type: object
+    description: "The required signed components of block production according to the Fulu CL spec."
+    required: [signed_block, kzg_proofs, blobs]
+    properties:
+      signed_block:
+        $ref: "../electra/block.yaml#/Electra/SignedBeaconBlock"
+      kzg_proofs:
+        description: "These should be cell proofs of the blobs."
```
dapplion, line 23:
```suggestion
        description: "Cell proofs of the blobs as defined in EIP-7594"
```

beacon-node-oapi.yaml
```diff
@@ -157,11 +156,11 @@ paths:
     $ref: "./apis/builder/states/expected_withdrawals.yaml"
 
   /eth/v2/debug/beacon/states/{state_id}:
-    $ref: './apis/debug/state.v2.yaml'
+    $ref: "./apis/debug/state.v2.yaml"
```
nflaig, line 159: the formatting changes in this PR got a bit out of hand, we should revert those changes before merging

0x00101010: I'd suggest keeping these, lots of the formatting in these yaml files are inconsistent, there are single and double quotes everywhere, this PR just tries to move them into one consistent direction (double quote)

mcdee: If you want to tidy up the formatting perhaps break it out into a separate PR.  It would also make this PR far more readable.

nflaig: this PR initially had some formatting changes which was ok but now it has more formatting changes than anything else, we can easily merge a PR that fixes that separately and maybe even consider adding a linter to CI, if you feel strongly about this @0x00101010, just don't see why a PR that proposes PeerDAS changes should have all those changes in it

0x00101010: Cool, I think you're right, I didn't notice that oapi.yaml was changed so much, reverted all the formatting related changes.

types/fulu/execution_payload.yaml
```diff
@@ -0,0 +1,81 @@
...
+    allOf:
+      - $ref: "#/Fulu/ExecutionPayloadCommon"
+      - type: object
+        required: [transactions, withdrawals, proof_version]
+        properties:
+          transactions:
+            $ref: "../bellatrix/transactions.yaml#/Bellatrix/Transactions"
+          withdrawals:
+            $ref: "../capella/withdrawals.yaml#/Capella/Withdrawals"
+          proof_version:
```
nflaig, line 67: is this the only field that actually changes compared to Electra which requires to update `BeaconBlockBody`? I can't find that in the CL spec

types/fulu/block_contents.yaml
```diff
@@ -0,0 +1,26 @@
...
+  SignedBlockContents:
+    type: object
+    description: "The required signed components of block production according to the Fulu CL spec."
+    required: [signed_block, kzg_proofs, blobs]
+    properties:
+      signed_block:
+        $ref: "../electra/block.yaml#/Electra/SignedBeaconBlock"
+      kzg_proofs:
+        description: "Cell proofs of the blobs as defined in EIP-7594"
+        $ref: "../deneb/block_contents.yaml#/Deneb/KZGProofs"
```
nflaig, line 24: doesn't the list limit change in PeerDAS compared to Deneb?

types/fulu/execution_payload.yaml
0x00101010: Yes, this is the only field change, also cc @jtraglia to see if we want to add that to CL spec

types/fulu/block_contents.yaml
0x00101010: Great catch, updated

types/fulu/execution_payload.yaml
nflaig: >  also cc @jtraglia to see if we want to add that to CL spec

that would be good otherwise I am not sure how to merge this as beacon-api generally follows the consensus spec, but we can merge this soon, just wanna do another minor release for Electra before that in a few days

jtraglia: Ah sorry just now seeing this. I will make a PR for this.

jtraglia:
Do we need `fulu.ExecutionPayloadCommon` and `fulu.ExecutionPayloadHeader` if there are no changes since Deneb?

nflaig:
> Do we need `fulu.ExecutionPayloadCommon` and `fulu.ExecutionPayloadHeader` if there are no changes since Deneb?

No we don't need it, I just wanna focus the discussion for now on PeerDAS required changes, so I tried to keep my review minimal for now, can simplify this before merging

types/fulu/execution_payload.yaml
ralexstokes: @0x00101010 why do we have the proof version here?

the proof version byte is just to disambiguate the EL wrapper, i don't see where it leaks into the CL at all

0x00101010: I've removed Fulu BeaconBlock / SignedBeaconBlock / ExecutionPayload, since it does not change, we can reuse the Electra definition

jimmygchen (commented):
Also need `fulu` here:

https://github.com/ethereum/beacon-APIs/blob/1eda435f780bf39aa6748a1d4c11e833f6c1a75c/beacon-node-oapi.yaml#L353

beacon-node-oapi.yaml
```diff
@@ -438,6 +438,10 @@ components:
       $ref: './types/electra/consolidation.yaml#/Electra/PendingConsolidation'
     Electra.PendingPartialWithdrawal:
       $ref: './types/electra/withdrawal.yaml#/Electra/PendingPartialWithdrawal'
+    Fulu.BlockContents:
```
jimmygchen, line 441: I think we also need a `Fulu.BlobSidecars` for the `getBlobSidecars` endpoint? Could be in a separate PR though.

0x00101010: Yeah, since max blob number is not determined, prefer to do it later

jimmygchen: Blob count is not the only thing that changes - each blob sidecar has a KzgProof field that should be changed to plurals with cell proofs. I reckon we could set a high but realistic blob count for the current PeerDAS, to avoid having to update this endpoint on every BPO hard fork?

0x00101010: cool, will update it tomorrow with other changes

nflaig: > I think we also need a Fulu.BlobSidecars for the getBlobSidecars endpoint? Could be in a separate PR though.

or we could deprecate the current one and at a new `getDataColumnSidecars`? I feel like this is something consumers should give feedback on before doing something, afaik, L2 clients use those apis

nflaig:
> There are a few endpoints that lists the consensus versions, perhaps we could update them to use this enum too, to avoid having to update it everywhere?

definitely worth simplifying this but refactoring seems out of scope of this PR, we can clean up afterwards

beacon-node-oapi.yaml
jimmygchen: I think we'd  probably want `getBlobsSidecarV2` for blob consumers (L2s). `DataColumnSidecars` is unlikely useful to end users - it's a CL-specific structure used for sample distribution and requires extra effort to convert into blobs, which L2 users are unlikely interested in. There could be use cases I'm missing though.

nflaig: > I think we'd probably want getBlobsSidecarV2 for blob consumers (L2s)

how would a CL be able to reconstruct the blob if it's not a supernode? and is there any reason why we need a v2 api, the current api is fork-aware so can change the type if required

0x00101010: Let's kick off the discussion on getBlobsSidecar/V2 api, let's do it in a separate PR to speed up merging this one

nflaig (approved):
LGTM - thanks @0x00101010 for taking care of this

apis/beacon/blob_sidecars/blob_sidecars.yaml
```diff
@@ -7,6 +7,9 @@ get:
 
     If the `indices` parameter is specified, only the blob sidecars with the specified indices will be returned. There are no guarantees
     for the returned blob sidecars in terms of ordering.
+
+    After Fulu, clients may reconstruct the blob sidecars from data columns and return corresponding blobs to users. It is users' responsibility
+    to ensure that they're connected to a super node (custodies > 64 columns) to guarantee the availability of the blobs.
```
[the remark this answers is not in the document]
nflaig: Changes have been moved out of this PR to allow further discussion separately and to move ahead merging this

PR #518, opened 2025-03-29
Clarify handling of empty arrays for ids/statuses in validators api

nflaig:
Similar https://github.com/ethereum/beacon-APIs/pull/453, clarifies handling of empty arrays, ie. if the value is `[]` then no filtering should be applied and all validators should be returned. Since this [caused interop issues](https://discord.com/channels/595666850260713488/892088344438255616/1348727918951665815) just recently, I think it's worth clarifying this behavior on the spec. 
cc @james-prysm @pawanjay176 are you fine with the described behavior?


james-prysm (approved):
nice i agree

pawanjay176 (approved):
I'm good with this too. 
Just to be clear, this isn't the current behaviour in lighthouse. We return a 404 if the list is empty so this will require changes on our side.

rolfyone (approved):
LGTM

PR #522, opened 2025-04-03
Add the ability to easily query liveness of all validators

aimxhaisse:
The goal is to make monitoring tools that watch the entire network more efficient by not having to send the full list of validators indexes.

This is aligned with the current semantics of the `/eth/v1/beacon/states/{state_id}/validator_balances` call, which treats empty lists as full validator set.

### DIFF
--- a/apis/validator/liveness.yaml
+++ b/apis/validator/liveness.yaml
@@ -20,7 +20,10 @@ post:
       schema:
         $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
   requestBody:
-    description: "An array of the validator indices for which to detect liveness."
+    description: |
+      An array of the validator indices for which to detect liveness.
+
+      If the supplied list is empty (i.e. the body is `[]`) or no body is supplied then liveness will be returned for all validators.
     required: true
     content:
       application/json:


aimxhaisse:
Same mindset as https://github.com/ethereum/beacon-APIs/pull/518/

nflaig:
> Same mindset as #518

This case is a bit different in my opinion, the PR you referenced just clarified the expected behavior on the spec because there were different behaviors in each client and the params are considered "filters".

While the spec of this api states the following
> An array of the validator indices for which to detect liveness.

And the schema itself marks the body as `required: true` + the array itself has `minItems: 1`.

So I would expect that no implementation actually supports the new behavior this PR adds (we don't in Lodestar), it seems like a breaking change to the api to me which requires a version bump

The second argument against the proposed change is that this api is part of `/validators` namespace, and as noted in the spec those are intended for validator clients, not for analyzing / monitoring the chain.

That said, I am not opposed to the idea but this api should be part of `/beacon` namespace, and maybe even another sub namespace like `/beacon/monitoring` (just thinking out loud) because it's neither data retrieved from state nor blocks, in Lodestar for example we determine this based on our validator seen caches.

nflaig:
see https://github.com/ethereum/beacon-APIs/pull/131 for more context regarding the api

nflaig:
can this be closed? The PR as is cannot be merged as it changes the behavior of the existing api, feel free to open an issue if there's a specific use case you would like to see better supported by the beacon api

rolfyone:
yes, lets close this i think.

PR #524, opened 2025-04-23
[EIP-7594] Add new getBlobSidecars spec

0x00101010:


apis/beacon/blob_sidecars/blob_sidecars.yaml
```diff
@@ -7,6 +7,10 @@ get:
 
     If the `indices` parameter is specified, only the blob sidecars with the specified indices will be returned. There are no guarantees
     for the returned blob sidecars in terms of ordering.
+
+    After the Fulu fork, getBlobSidecars will still return blob sidecars with below caveats:
+    1. Only supernodes (which custodies all data columns) are required to return blob sidecars. It's up to the client if they want to implement blob reconstruction logic for non super nodes.
+    2. for the returned blob sidecars, kzg_proof should be zeroed out since we don't have need blob proofs any more.
```
jtraglia, line 13: This suggestion fixes a few little nits & expands upon a couple things.
```suggestion
    1. Only supernodes (which custody all data column sidecars) are required to return blob sidecars. It is up to the client if they want to implement blob reconstruction logic for non-super nodes.
    2. For the returned blob sidecars, `kzg_proof` should be zeroed out (`0x0000...0000`, an invalid compressed G1 point which will error if used in verification) since blob proofs are no longer computed; cell proofs are used instead. Also, this is generally considered a trusted endpoint, so recomputing blob proofs would be an unnecessary expense.
```

0x00101010: 👍 updated accordingly

types/fulu/blob_sidecar.yaml
```diff
@@ -0,0 +1,8 @@
+Fulu:
+  BlobSidecars:
+    type: array
+    items:
+      $ref: "#/Deneb/BlobSidecar"
+    minItems: 0
+    maxItems: 48 # placeholder, depending on the final value we chose and BPO schedule
```
jtraglia, line 7: Hmm I feel that we should just use a really high value here instead. Let's discuss this first.
```suggestion
    maxItems: 4096  # MAX_BLOB_COMMITMENTS_PER_BLOCK
```

0x00101010: +1 on setting it high enough that we don't ever need to change it anymore

nflaig: there was some discussion in https://github.com/ethereum/beacon-APIs/pull/488 on this previously, I am generally in favor of using sane list limits but since blob limit increase will be effectively decoupled from (named) forks with EIP-7892 I also think it's best to use 4096 now

g11tech: +1 for 4096

0x00101010: seems like we have enough consensus here, updated

wordlist.txt
```diff
@@ -37,3 +37,4 @@ KZGProofs
 KZGCommitmentInclusionProof
 LMD
 fulu
+BPO
```
nflaig, line 40: this can be removed, was previous in comment for `maxItems` but removed in last commit

nalepae:
Currently, this beacon API returns blob sidecars, including:
- the blob index,
- **the blob**,
- the blob KZG commitment,
- the signed block header, and
- the KZG commitment inclusion proof.

It seems main users of the `getBlobSidecars` endpoint (L2s) mainly need the blob, and not all the other fields available in the blob sidecar.

Proposal: What about creating a new beacon API endpoint `getBlob` (and deprecating this one), returning [only](https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#blobsidecar) the blob?

[Initial Discord thread](https://discord.com/channels/595666850260713488/1355219344011038912/1366872612395024434)

Issue #308, opened 2023-03-09
withdrawable epoch measurement API


0x8f701:
Currently we need to reimplement everything that beacon node is doing to measure the withdrawable epoch and time. It'll be good to have an API returning back informations like:

- exitEpoch
- withdrawableEpoch
- ...

If a validator is going to exit now. This can help building some applications. Validator Index will be the input.

ojuschugh1:
Hi @rolfyone @staccDOTsol @0x8f701 @dapplion  @Everyone, I have created a PR for this issue, pls review it.

https://github.com/ethereum/beacon-APIs/pull/526

nflaig:
https://ethereum.github.io/beacon-APIs/#/Beacon/getStateValidator can be used to retrieve that information per validator

PR #526, opened 2025-04-28
Add endpoint to retrieve exit and withdrawable epochs for a validator

ojuschugh1:
This PR adds a new lightweight API endpoint to retrieve a validator’s exit_epoch and withdrawable_epoch, addressing the enhancement requested in [Issue #308](https://github.com/ethereum/beacon-APIs/issues/308).

Instead of fetching the full validator set via /eth/v1/beacon/states/{state_id}/validators, this new endpoint allows querying only a specific validator by ID, simplifying application development and reducing unnecessary data transfer.

### DIFF
--- a/apis/validator/withdrawable_epochs.yaml
+++ b/apis/validator/withdrawable_epochs.yaml
@@ -1,0 +1,29 @@
+get:
+  operationId: getWithdrawableEpochs
+  summary: Get exit and withdrawable epochs of a validator
+  description: Returns the exit and withdrawable epochs for a given validator ID.
+  tags:
+    - Beacon
+  parameters:
+    - name: validator_id
+      in: path
+      required: true
+      description: ID of the validator to query
+      schema:
+        type: string
+  responses:
+    '200':
+      description: Successfully retrieved withdrawable info
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              exit_epoch:
+                type: string
+                description: Epoch when validator exits
+              withdrawable_epoch:
+                type: string
+                description: Epoch when validator becomes withdrawable
+    '404':
+      description: Validator not found
--- a/beacon-node-oapi.yaml
+++ b/beacon-node-oapi.yaml
@@ -138,6 +138,8 @@ paths:
     $ref: "./apis/beacon/light_client/optimistic_update.yaml"
   /eth/v1/beacon/pool/attestations:
     $ref: "./apis/beacon/pool/attestations.yaml"
+  /eth/v1/beacon/validators/{validator_id}/withdrawable_epochs:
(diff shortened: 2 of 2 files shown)


ojuschugh1:
@nflaig  Great point! You’re absolutely right that GET /eth/v1/beacon/states/{state_id}/validators/{validator_id} already includes exit_epoch and withdrawable_epoch. However, it also returns a large schema with many unrelated fields (e.g., pubkey, balances, status flags), which clients must filter out. By providing a dedicated /eth/v1/beacon/validators/{validator_id}/withdrawable_epochs endpoint, we can offer a more focused, bandwidth-efficient response and clearer intent for applications that only need withdrawal timing. This aligns with prior discussions in issue #308 and mirrors patterns in other client APIs that introduce specialized endpoints for common queries. What do you think about adding such an endpoint for improved ergonomics?

rolfyone:
I'd be inclined to use the state validators endpoint as @nflaig suggested, its only a handful of extra data, and im not sure we want to spin up tons of extra interfaces that are as specific as this one.

ojuschugh1:
Hi @rolfyone , thanks for your reply. Sure, I will use that as per @nflaig  suggestions for the feedback :)

nflaig:
> Hi @rolfyone , thanks for your reply. Sure, I will use that as per @nflaig suggestions for the feedback :)

can this be closed? I am having a hard time justifying adding this api to the spec as it also doesn't not use a CL spec type which means more work on the implementation and custom container if we want to support SSZ

ojuschugh1:
Sure, @nflaig—could you also ask to close this linked issue (https://github.com/ethereum/beacon-APIs/issues/308)? That way, it won’t lead to any further misunderstandings like mine. 🙂

nflaig:
> Sure, @nflaig—could you also ask to close this linked issue (#308)? That way, it won’t lead to any further misunderstandings like mine. 🙂

yeah agree this should be closed, I am not sure I understand what the author of the issue is asking for because `/eth/v1/beacon/states/{state_id}/validators/{validator_id}` already existed back then, but let's close it, can always be reopened if needed

ojuschugh1:
> > Sure, @nflaig—could you also ask to close this linked issue (#308)? That way, it won’t lead to any further misunderstandings like mine. 🙂
> 
> yeah agree this should be closed, I am not sure I understand what the author of the issue is asking for because `/eth/v1/beacon/states/{state_id}/validators/{validator_id}` already existed back then, but let's close it, can always be reopened if needed

True, thanks for the help, I was looking for a good first issue and found this one. I will find some other issue to work on now ;)

nflaig:
> True, thanks for the help, I was looking for a good first issue and found this one. I will find some other issue to work on now ;)

thanks for looking into it, you found a stale issue which is good as well to clean those up :)

PR #529, opened 2025-05-08
Add blob schedule example to config/spec api

nflaig:
To accommodate https://github.com/ethereum/consensus-specs/pull/4277 and clarify format we expect to be returned from `/eth/v1/config/spec` api.

### DIFF
--- a/apis/config/spec.yaml
+++ b/apis/config/spec.yaml
@@ -10,6 +10,7 @@ get:
     Values are returned with following format:
       - any value starting with 0x in the spec is returned as a hex string
       - numeric values are returned as a quoted integer
+      - array values are returned as a JSON array
   tags:
     - Config
     - ValidatorRequiredApi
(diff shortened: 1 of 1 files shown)


rolfyone (approved):
LGTM

barnabasbusa (approved):

jimmygchen (approved):
Looks good, thanks!

nflaig:
> In our current implementation, we sort the list while deserialize the config file, so we don't need to sort it every time the function called. I knew it won't cause much computation overhead for sorting few items, but the function will be often called while it is unnecessary redundant operation.

Currently [`get_max_blobs_per_block`](https://github.com/ethereum/consensus-specs/pull/4277/files#diff-f8f2291b3c58ef250e9d37b9193052d95307e582996f86667c57eeeba62bc750R137-R145) in the CL spec has the sorting inside the function, so if you wanna follow close to spec you would have it in the implementation as well. I am mostly wondering if this needs to be noted in the api or if it should rather return the config as it is configured in the client.

> How others client feel if the BPO entries return as descending sorted?

Does that mean if someone passes in the config, would we error if it's not sorted?

Could open a follow-up PR / issue for more discussion, the main goal of this PR was to have a common format and not return a stringified array for example which would have been alternative format we could use

hangleang:
> Does that mean if someone passes in the config, would we error if it's not sorted?

I believe clients should be able to handle regardless of what order is

PR #531, opened 2025-05-12
Deprecate pre electra attestation submissions

eserilev:
This PR aims to deprecate support for pre-electra `Attestations` in the `submitPoolAttestationsV2` endpoint. Since we are post Electra it seems that there is no reason to support the pre-Electra variant. It also allows client teams to simplify the implementation of this endpoint (e.g. Lighthouse can remove 100-ish lines of code with this deprecation). To be spec compliant clients must already be submitting `SingleAttestation`s to this endpoint, so unless there are non-spec compliant client implementations, this shouldn't cause any disruptions.

### DIFF
--- a/apis/beacon/pool/attestations.v2.yaml
+++ b/apis/beacon/pool/attestations.v2.yaml
@@ -62,7 +62,6 @@ post:
 
     If one or more attestations fail validation, the node MUST return a 400 error with details of which attestations have failed, and why.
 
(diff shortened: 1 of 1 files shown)


nflaig:
not really in favor of this on the spec side, so far all apis have been kept backward compatible, implementations can choose to stop supporting previous forks if they would like to

PR #534, opened 2025-05-24
Update example to use valid json.

Bez625:
The response for the `BLOB_SCHEDULE` should be a JSON array and the provided example does not conform to valid JSON. Updating to use double quotes as opposed to single quotes.


nflaig (approved):
Thanks @Bez625 good catch!

PR #535, opened 2025-06-03
Add data_column_sidecar event to eventstream API

Savid:
Extends the eventstream API to support data_column_sidecar events for PeerDAS (EIP-7594) data availability sampling. The event is emitted when the node receives a DataColumnSidecar that passes all gossip validations.

### DIFF
--- a/apis/eventstream/index.yaml
+++ b/apis/eventstream/index.yaml
@@ -38,6 +38,7 @@ get:
             - light_client_optimistic_update
             - payload_attributes
             - blob_sidecar
+            - data_column_sidecar
   responses:
     "200":
       description: Opened SSE stream.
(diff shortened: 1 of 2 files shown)


apis/eventstream/index.yaml
```diff
@@ -152,6 +153,11 @@ get:
               value: |
                 event: blob_sidecar
                 data: {"block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "index": "1", "slot": "1", "kzg_commitment": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505", "versioned_hash": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"}
+            data_column_sidecar:
+              description: The node has received a DataColumnSidecar (from P2P or API) that passes all gossip validations on the data_column_sidecar_{subnet_id} topic
+              value: |
+                event: data_column_sidecar
+                data: {"block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "index": "1", "slot": "1", "kzg_commitments": ["0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"]}
```
etan-status: the one for blobsidecar has `versioned_hash`. is a matching `versioned_hashes` field no longer needed here?

tersec (approved):

apis/eventstream/index.yaml
Savid: updated

nflaig (approved):

apis/eventstream/index.yaml
agnxsh: @etan-status flattening the versioned hashes to a flat byte array and 0xHex-ing it can take a while, per column, per notification, i'd suggest it would be nice if we can skip this field or something, if there's not much use case

moreover i think we'd be streaming the same set of versioned hashes for every data column no? can someone confirm? cc @nalepae

PR #537, opened 2025-06-03
Add data column sidecars debug endpoint

Savid:
Introduces support for data column sidecars as part of EIP-7594 (PeerDAS):
- Add /eth/v1/debug/beacon/data_column_sidecars/{block_id} endpoint
- Define DataColumnSidecar and related types for Fulu fork
- Add Cell primitive type for 2048-byte data columns

### DIFF
--- a/apis/debug/data_column_sidecars.yaml
+++ b/apis/debug/data_column_sidecars.yaml
@@ -1,0 +1,73 @@
+get:
+  operationId: getDebugDataColumnSidecars
+  summary: Get data column sidecars
+  description: |
+    Retrieves data column sidecars for a given block id.
+    Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.
+
+    If the `indices` parameter is specified, only the data column sidecars with the specified indices will be returned. There are no guarantees
+    for the returned data column sidecars in terms of ordering.
+  tags:
+    - Debug
+  parameters:
+    - name: block_id
+      in: path
+      required: true
+      $ref: '../../beacon-node-oapi.yaml#/components/parameters/BlockId'
+    - name: indices
+      in: query
+      description: Array of indices for data column sidecars to request for in the specified block. Returns all data column sidecars in the block if not specified.
+      required: false
+      schema:
+        type: array
+        uniqueItems: true
+        items:
+          $ref: '../../beacon-node-oapi.yaml#/components/schemas/Uint64'
+  responses:
+    "200":
+      description: "Successful response"
+      headers:
+        Eth-Consensus-Version:
+          $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+      content:
+        application/json:
+          schema:
+            title: GetDebugDataColumnSidecarsResponse
+            type: object
+            required: [version, execution_optimistic, finalized, data]
+            properties:
+              version:
+                type: string
+                enum: [fulu]
+                example: "fulu"
+              execution_optimistic:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/ExecutionOptimistic"
+              finalized:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Finalized"
+              data:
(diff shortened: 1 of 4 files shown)


types/primitive.yaml
```diff
@@ -142,3 +142,8 @@ KZGProof:
   pattern: "^0x[a-fA-F0-9]{96}$"
   description: "A G1 curve point. Used for verifying that the `KZGCommitment` for a given `Blob` is correct."
 
+Cell:
```
etan-status: Cell is a FullDAS concept, this should be a DataColumn or Column, for correct typing

types/primitive.yaml
```diff
@@ -142,3 +142,8 @@ KZGProof:
...
+  type: string
+  format: hex
+  pattern: "^0x[a-fA-F0-9]{4096}$"
+  description: "A 2048-byte hexadecimal string, prefixed with `0x`"
```
etan-status: This `description` should describe semantics, not the format (see `Blob` for how it should be done).
The format is already given by the `pattern`

tersec: inline this, don't have explicit `Cell` primitive yet because it's not in PeerDAS as such

types/fulu/data_column_sidecar.yaml
```diff
@@ -0,0 +1,44 @@
...
+  KZGCommitmentsInclusionProof:
+    type: array
+    items:
+      $ref: "../primitive.yaml#/Bytes32"
+    minItems: 4
+    maxItems: 4
+
+  DataColumnSidecar:
+    type: object
+    description: "A data column sidecar as defined in the Fulu consensus spec."
```
nflaig: would be good to add a reference to CL spec type here

similar to https://github.com/ethereum/beacon-APIs/blob/8717bf6feb9856309bee0ebf165c38e02afd0a7a/types/electra/attestation.yaml#L36-L38

beacon-node-oapi.yaml
```diff
@@ -118,6 +118,8 @@ paths:
     $ref: "./apis/beacon/blocks/attestations.v2.yaml"
   /eth/v1/beacon/blob_sidecars/{block_id}:
     $ref: "./apis/beacon/blob_sidecars/blob_sidecars.yaml"
+  /eth/v1/debug/beacon/data_column_sidecars/{block_id}:
```
dapplion, line 123: Not sure if this fits in the `debug` namespace, should it be beacon?

```
/eth/v1/beacon/data_column_sidecars/{block_id}
```

apis/debug/data_column_sidecars.yaml
```diff
@@ -0,0 +1,73 @@
...
+  tags:
+    - Debug
+  parameters:
+    - name: block_id
+      in: path
+      required: true
+      $ref: '../../beacon-node-oapi.yaml#/components/parameters/BlockId'
+    - name: indices
+      in: query
+      description: Array of indices for data column sidecars to request for in the specified block. Returns all data column sidecars in the block if not specified.
```
dapplion: Description should clarify that it will return only the columns that this node is custodying. 

Question, if this node has in custody indices [0,1] and the query parameter indices is [1,2]:
- Should the node error, or
- Should the node return [1]

types/primitive.yaml
nflaig: seems like this is the name [used in the CL spec](https://github.com/ethereum/consensus-specs/blob/d92c60c5051e2b1faab4b736757273f587219023/specs/fulu/das-core.md?plain=1#L78)

beacon-node-oapi.yaml
rolfyone: is it useful for general consumption @dapplion ?

types/primitive.yaml
Savid: moved inline

rolfyone (approved):
from discussion sounds like no blockers so merging (irl. discussions)

apis/debug/data_column_sidecars.yaml
nflaig: > Should the node return [1]

probably this, might be worth to do a follow-up PR that clarifies this

PR #538, opened 2025-06-09
Add next fork digest to `MetaData`

ensi321:
Similar to cgc, we want to add nfd to `MetaData`. 

Related to ethereum/consensus-specs#4354


rolfyone:
im not convinced we need this?

PR #539, opened 2025-06-10
Proposer lookahead

rolfyone:
This will make it a lot easier to retrieve this data rather than accessing beacon state.

I left in octet response because it's already an SSZ object so we may as well support it.

### DIFF
--- a/apis/beacon/states/proposer_lookahead.yaml
+++ b/apis/beacon/states/proposer_lookahead.yaml
@@ -42,8 +42,8 @@ get:
       $ref: '../../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
     "404":
       $ref: '../../../beacon-node-oapi.yaml#/components/responses/NotFound'
-    "415":
-      $ref: '../../../beacon-node-oapi.yaml#/components/responses/UnsupportedMediaType'
+    "406":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/NotAcceptable'
     "500":
       $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'
 


apis/beacon/states/proposer_lookahead.yaml
```diff
@@ -0,0 +1,49 @@
...
+      content:
+        application/json:
+          schema:
+            title: GetProposerLookaheadResponse
+            type: object
+            required: [version, execution_optimistic, finalized, data]
+            properties:
+              version:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion"
+                example: "electra"
```
nflaig:
```suggestion
                example: "fulu"
```

apis/beacon/states/proposer_lookahead.yaml
```diff
@@ -0,0 +1,49 @@
+get:
+  operationId: "getProposerLookahead"
+  summary: "Get State Proposer Look-ahead"
```
nflaig:
```suggestion
  summary: "Get State Proposer Lookahead"
```

apis/beacon/states/proposer_lookahead.yaml
```diff
@@ -0,0 +1,49 @@
...
+  description: |
+    Returns proposer look-ahead for state with given 'stateId'. Should return 400 if the state retrieved is prior to Fulu.
```
nflaig:
```suggestion
    Returns proposer lookahead for state with given 'stateId'. Should return 400 if the state retrieved is prior to Fulu.
```

apis/beacon/states/proposer_lookahead.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: state_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+  responses:
+    "200":
+      description: Success
+      headers:
+        Eth-Consensus-Version:
+          $ref: '../../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Version'
+          description: "The active consensus version to which the proposer look ahead data belongs."
```
nflaig:
```suggestion
          description: "The active consensus version to which the proposer lookahead data belongs."
```

apis/beacon/states/proposer_lookahead.yaml
rolfyone: i mean i was using english, but we can be incorrect i guess

apis/beacon/states/proposer_lookahead.yaml
```diff
@@ -0,0 +1,49 @@
...
+                maxItems: 64
+        application/octet-stream:
+          schema:
+            description: "SSZ serialized `List[Uint64, (MIN_SEED_LOOKAHEAD + 1) * SLOTS_PER_EPOCH]` bytes. Use Accept header to choose this response type"
+    "400":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/InvalidRequest'
+    "404":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/NotFound'
+    "415":
+      $ref: '../../../beacon-node-oapi.yaml#/components/responses/UnsupportedMediaType'
```
nflaig: just noticed this should be 406, it's incorrect in the other apis for consolidations, deposits ,etc. as well

```suggestion
    "406":
      $ref: '../../../beacon-node-oapi.yaml#/components/responses/NotAcceptable'
```

nflaig (approved):

PR #541, opened 2025-06-11
Replace 415 response with 406 for apis that only have ssz response body

nflaig:
As suggested in https://github.com/ethereum/beacon-APIs/pull/539#discussion_r2139507320


rolfyone (approved):
LGTM

PR #545, opened 2025-07-25
attestation_data api clarification for Electra

tbenr:
In electra, we are not required to call the endpoint for validators in different committees for the same slot.
The committeeIndex parameter is essentially redundant and could be always set to 0.

### DIFF
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -20,7 +20,9 @@ get:
         $ref: ../../beacon-node-oapi.yaml#/components/schemas/Uint64
     - name: committee_index
       in: query
-      description: "The committee index for which an attestation data should be created."
+      description: |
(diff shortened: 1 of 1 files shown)


apis/validator/attestation_data.yaml
```diff
@@ -20,7 +20,9 @@ get:
         $ref: ../../beacon-node-oapi.yaml#/components/schemas/Uint64
     - name: committee_index
       in: query
-      description: "The committee index for which an attestation data should be created."
+      description: |
+        The committee index for which an attestation data should be created. For `slot`s in
+        Electra and later, this paramiter MAY always be set to 0.
```
nflaig:
```suggestion
        Electra and later, this parameter MAY always be set to 0.
```

nflaig (approved):

Issue #332, opened 2023-06-27
Blob retrieval by versioned hash


terencechain:
Currently, in the construction of retrieving a blob, the request is based on the block ID, which can be either a block root or specific keywords like "finalized" or "justified." However, it would be beneficial for L2 in general to have the ability to retrieve blobs using versioned hashes. This would address a pain point for L2 mapping, as L2s primarily focus on versioned hashes and it would be more sensible to use them as the request parameter rather than beacon block roots

mcdee:
At current each potential block ID is easily identifiable.  Both block roots and versioned hashes have the same structure (32-byte array) so this would cause issues with servers in working out which should be used for retrieval.

It would be possible to send this as a query parameter, but then that breaks the existing format of the URL.  Perhaps we could add another endpoint (and do so in a different namespace) that is more focused on providing support to L2s?

terencechain:
> Perhaps we could add another endpoint (and do so in a different namespace) that is more focused on providing support to L2s?

Makes sense to me

tbenr:
Having a separated endpoint seems to make sense also from a perspective of making this optional. Since maybe not all are interested in consuming this api, we could avoid to have the versionedHash->blobSidecar index always populated in DB if the api is not used.

terencechain:
> 

+1 on optional

kasey:
For the existing endpoint, in the case where an L2 is able to compute the required slot from on-chain data, L2 server and CL client would both benefit from the ability to limit retrieval and transmission to only the necessary set of indices, which L2 knows by their versioned hashes. We could add another optional parameter `versioned_hashes` which would filter the response to a subset of blobs, just like the existing `indices` param: https://github.com/ethereum/beacon-APIs/blob/master/apis/beacon/blob_sidecars/blob_sidecars.yaml#L17-L25

jpnovais:
+1 to implement a method to fetch blobs by versioned hashes. L2 would benefit a lot from this, especially to rebuild/verify L2 state from L1 blobs.

tbenr:
@terencechain @dapplion @kasey did you endup implementing it? Did you followed [this](https://github.com/ethereum/beacon-APIs/issues/332#issuecomment-1629384505) approach?
Looking at it `versioned_hashes` could be the only input for that api, avoiding the `block_id`, or do you think that L2 wants to fix a block anyway?

PR #546, opened 2025-07-26
Add getBlobs endpoint

nflaig:
As suggested in https://github.com/ethereum/beacon-APIs/pull/524#issuecomment-2892253605, adds new `getBlobs` endpoint which only returns blobs.


apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,69 @@
+get:
+  operationId: getBlobs
+  summary: Get blobs
+  description: |
+    Retrieves blobs for a given block id.
+    Depending on `Accept` header it can be returned either as json or as bytes serialized by SSZ.
+
+    If the `indices` parameter is specified, only the blobs with the specified indices will be returned. There are no guarantees
```
jimmygchen: I'm not sure how useful is it for users to get a list of unordered blobs without indices. If this is intended for L2s, we might as well implement retrieval by version hashes?
https://github.com/ethereum/beacon-APIs/issues/332

nflaig: seems like from the previous discussion it might be better, would we still keep the `block_id` in that case as noted in https://github.com/ethereum/beacon-APIs/issues/332#issuecomment-2343790172? We index data columns sidecars by block root or slot so dropping this would require additional indexing

jimmygchen: yep you're right, I'm bringing this up because the intent of this endpoint is for L2 users - I'm not exactly sure how they use it, but i assume they have `block_id` and `version_hash`, and so they can retrieve the blob they need?

With the `getBlobSidecar` endpoint, they're able to do this by computing the `version_hash` of individual sidecars from `kzg_commitment` - but without the sidecar metadata this won't work anymore.

I see a few possible options:
1. return KZG commitments / version hash together with the blob
2. implement version hash filtering on the server side
3. returned an ordered list, so they're able to figure out which blobs they're after from the `block.kzg_commitments`

I think #3 may be the least work on the beacon node, and also keep the API clean?

nflaig: > but i assume they have block_id and version_hash, and so they can retrieve the blob they need?

Not sure how they use it either but since the blob sidecars endpoint required `block_id` already it should be fine

> I think https://github.com/ethereum/beacon-APIs/pull/3 may be the least work on the beacon node, and also keep the API clean?

(3) is the simplest for us and that's how I implemented it for now but using `version_hashes` for filtering should be easy change, if we go with (3) it makes sense to add explicit note that an ordered list must be returned

Thegaram: For Scroll, it would be more convenient to have a "get blob(s) by versioned hash(es)" endpoint (without block id).

fluentcrafter: For Linea, retrieving the blob by `blobVersionedHashes` is the best as well, without block_id or any other cl/beacon related fields. 

The reason is that when fetching L1 EL transaction details (e.g `eth_getTransactionByHash`), we only have the `blobVersionedHashes` that we can use to query the CL APIs to fetch the blobs.

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,69 @@
...
+    "200":
+      description: "Successful response"
+      content:
+        application/json:
+          schema:
+            title: GetBlobsResponse
+            type: object
+            required: [execution_optimistic, finalized, data]
+            properties:
+              execution_optimistic:
```
james-prysm, line 39: do we not need versions as a field? ( is that helpful or not?)

apis/beacon/blobs/blobs.yaml
```diff
@@ -0,0 +1,69 @@
...
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ExecutionOptimistic"
+              finalized:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Finalized"
+              data:
+                $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Blobs"
```
james-prysm, line 44: I think we have Blob but not Blobs

apis/beacon/blobs/blobs.yaml
nflaig: I left it out because the response is just a list of blobs and not a spec container that requires versioning. There no harm in adding it though if it's useful metadata to have

apis/beacon/blobs/blobs.yaml
nflaig: Blobs is just an array of Blob, reused the exisiting schema we had, or do you mean something else?

james-prysm: yes but on the components/schemas in beacon-node-oapi.yaml I don't think that definition shows up unless i just can't find it, i thought it would be under the Blob definition

apis/beacon/blobs/blobs.yaml
james-prysm: I'm not a big fan of these meta data fields in the first place ( thought they could be handled through the header) but keeping consistent with other endpoints I feel like it doesn't hurt to have it.

apis/beacon/blobs/blobs.yaml
nflaig: I added it in this PR [here](https://github.com/ethereum/beacon-APIs/pull/546/files#diff-962ea34dc6ec0e65d4e1d6c8682e97c547e6e0d2f1168925210ff7ee47ef6ce0R409-R410)

apis/beacon/blobs/blobs.yaml
nflaig: there are a lot of apis which don't have `version` but `execution_optimistic` and `finalized`

apis/beacon/blobs/blobs.yaml
james-prysm: oops thanks

james-prysm (approved):

apis/beacon/blobs/blobs.yaml
Tristan-Wilson: Right now for reading blobs in Arbitrum Nitro, we go from timestamp->slot to get beacon/blob_sidecars by slot. https://github.com/OffchainLabs/nitro/blob/48c6c56dd16d793f8259b15f655da5eadd9cc580/util/headerreader/blob_client.go#L164

Then we pull out the kzg commitment and use it to get the versioned hash, then we see if that's one of the versioned hashes we care about. Finally we pull out the blob itself and verify it against the KzgProof. 
https://github.com/OffchainLabs/nitro/blob/48c6c56dd16d793f8259b15f655da5eadd9cc580/util/headerreader/blob_client.go#L199

Changing the API to be lookup by versioned hash would definitely simplify this code, at the expense of more calls (1 vs 1 for each blob), and possibly making it harder for users to debug. We get a steady drip of support requests from people who haven't properly set up blob archive nodes (eg https://github.com/OffchainLabs/nitro/issues/3475) and currently having the API be by slot number makes it much easier to see that they're clearly trying to read earlier blobs than what their node has: https://github.com/OffchainLabs/nitro/blob/48c6c56dd16d793f8259b15f655da5eadd9cc580/util/headerreader/blob_client.go#L206
If we went to lookup by versioned hash there's no way to say why the hash is missing. So mainly for that reason I'm in favor of keeping the lookup by slot.

kasey: What if we add a new endpoint to get the fields of the data column sidecar that are common to all columns (everything but the column data and index), identified by all the standard block ids (root, slot, head, finalized etc)? This would enable the user of the API to compute versioned hashes from the commitments in the response, and then request blobs only for the versioned hashes they want. It would also enable them to use the block header (block root and parent block root) to confirm that they are consistently following the same chain. This could even be cut down to only commitments and header.

I'll also point out that enabling requests of blobs by versioned hash is a lot simpler to implement if we require a `(slot|root)` block identifier along with the versioned hash, because then the beacon api can look up the block using an existing index, compute the versioned hashes from commitments, map to a blob index, then build the corresponding blob data responses. With versioned hash only we would need to maintain another layer of indexing to support this API.

nflaig: > This would enable the user of the API to compute versioned hashes from the commitments in the response

can't you compute the versioned hashes from blob kzg commitments in the block?

> I'll also point out that enabling requests of blobs by versioned hash is a lot simpler to implement if we require a (slot|root) block identifier along with the versioned hash

This is my main concern, removing `block_id` requires the CL to do more work just to serve this api and if not all L2s are in favor of removing this, might be better to keep as is.

nflaig: @Thegaram @fluentcrafter since there isn't unanimous across L2s to drop `block_id` and it requires additional work on each CL client to add the db index for versioned hashes just to serve this api, I think it's best to keep `block_id`. Since the previous blobsidecars api required it too I would expect the code to handle this to be already in place.

Regarding filtering, could still use versioned hashes instead of indices to only return the blobs you need for given `block_id`. I don't have a opinion here, either one is quite simple to implement for CL clients.

tersec: It seems like from @Tristan-Wilson 's description slot numbers have concrete debugging and diagnostic advantages, too.

Thegaram: > I think it's best to keep block_id

In that case, from our perspective, I don't see much benefit to adding a new API.

Or is the idea here that later when we'll have much more blobs in each block, the response would grow too large? In that case, filtering based on versioned hashes would be useful.

nflaig: > In that case, from our perspective, I don't see much benefit to adding a new API.

the reason for the new api is that we no longer use and store `BlobSidecars` on the CL side once Fusaka goes live, see https://github.com/ethereum/beacon-APIs/pull/524#issuecomment-2892253605

> Or is the idea here that later when we'll have much more blobs in each block, the response would grow too large?

yeah, just to avoid returning data you don't need and with increasing blob count this matters more

nflaig:
Pushed some updates to the PR based on feedback
- use `versioned_hashes` instead of `indices` as filter (in addition to `block_id`)
- updated description that the blobs are ordered based on the order of kzg commitments in the block

rolfyone (approved):
looks like the discussion is completed here, approving

apis/beacon/blobs/blobs.yaml
Enigmatic331: Chiming in a little - I understand the OP folks are having some issues with plans to deprecate the blob_sidecars endpoint

`Fulu deprecates the /eth/v1/beacon/blob_sidecars Beacon node endpoint, replacing it with /eth/v1/beacon/blobs. However, this new endpoint only provides the raw blobs, without the new kzg commitments (for cell proofs), without which verification doesn't work. So we need to investigate how we want to support this new endpoint for untrusted beacon nodes.`

This is an endpoint we are currently using in production for data indexing as well, and would appreciate if it could be retained.

nflaig: > This is an endpoint we are currently using in production for data indexing as well, and would appreciate if it could be retained.

The reason for deprecating the `blob_sidecars` endpoint is because blob sidecars no longer exist after Fulu, we only have data column sidecars.

I'd rather update the response of the blobs endpoint to include additional data as previously suggested by @jimmygchen

PR #563, opened 2025-11-01
Add proposer duties v2 endpoint

nflaig:
Adds proposer duties v2 endpoint to account for deterministic proposer lookahead https://github.com/ethereum/consensus-specs/pull/4190 and return a different `dependent_root` based on current fork (pre/post fulu).

**NOTE:** this endpoint is not part of fulu release and is only required for gloas, it cannot be relied upon before that as clients might not have implemented it.

### DIFF
--- a/apis/validator/beacon_committee_selections.yaml
+++ b/apis/validator/beacon_committee_selections.yaml
@@ -6,9 +6,9 @@ post:
     implemented by a distributed validator middleware client. This endpoint is used to exchange partial 
     selection proofs for combined/aggregated selection proofs to allow a validator client 
     to correctly determine if any of its validators has been selected to perform an attestation aggregation duty in a slot. 
-    Validator clients running in a distributed validator cluster must query this endpoint at the start of an epoch for the current and lookahead (next) epochs for
-    all validators that have attester duties in the current and lookahead epochs. Consensus clients need not support this
-    endpoint and may return a 501.
+    Validator clients running in a distributed validator cluster must query this endpoint at the start of an epoch either for all slots of 
+    the current epoch or for all slots of the current and all slots of the next epoch in two separate requests. In both cases it should be for 
+    all validators that have attester duties in the respective epoch. Consensus clients need not support this endpoint and may return a 501.
   tags:
     - Validator
   requestBody:


apis/validator/duties/proposer.v2.yaml
```diff
@@ -0,0 +1,70 @@
...
+  operationId: "getProposerDutiesV2"
+  description: "Request beacon node to provide all validators that are scheduled to propose a block in the given epoch.
+
+    Duties should only need to be checked once per epoch,
+    however a chain reorganization could occur that results in a change of duties. For full safety,
+    you should monitor head events and confirm the dependent root in this response matches. After Fulu,
+    different checks need to be performed as the dependent root changes due to deterministic proposer lookahead.
+
+
+    Before Fulu:
```
rolfyone, line 15: should explicitly state we're required from gloas...

rolfyone (approved):
lgtm just a nit because we're tagging as required, we should put in description when it's required from to be clear.

apis/validator/duties/proposer.v2.yaml
nflaig: I don't think we ever added a note like this, the v2 endpoint will just part of `dev` release since fulu release is already out and it should be implicit that it's not part of fulu releases due to that. It's not different to how we rolled out the new attestation v2 apis, so I think this is fine (will add a note in the PR description)

rolfyone: its marked required, and not required in fulu, thats the wierdness im trying to encapsulate...

nflaig: do you mean because it's tagged as `ValidatorRequiredApi`? we can remove that tag for now and not deprecate the v1 api if that makes it clearer, and then later add the tag again+ deprecate v1

nflaig: or we add the same note I added in the PR description and later remove it again, both works for me

rolfyone: Given the version for fulu has shipped i guess its not a deal breaker either way, it just might be clearer since it's "validatorRequired" that we're fairly explicit... Maybe removing that tag for now is 'easy' - i'd be happy with that... 
I was just thinking a `Note: not required until GLOAS` would be an easy way to be explicit...

rolfyone:
now that i've actually implemented the dependent root change for fulu for teku, i do wonder if a new endpoint is correct or if we just need a better understanding of why this dependent root is different in fulu @michaelsproul... maybe the better option  is understanding that for the block proposals dependent root either 1 or 2 epochs prior are equally valid, and its just differing opinion on implementations....

I think that we'd probably need to 'hack' dependent root in teku now that it is computing proposer duties using lookahead...

apis/validator/duties/proposer.v2.yaml
```diff
@@ -0,0 +1,72 @@
...
+     - event.previous_duty_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
+
+     - event.block otherwise
+
+     - dependent_root value is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
+
+
+    The dependent_root value is the genesis block root in the case of underflow.
+    
+    Note: not required until Gloas"
```
nflaig: I believe this note is pretty unnecessary at this point and can be removed since Fulu is now already activated on mainnet

PR #564, opened 2025-11-04
Update default version example to fulu

nflaig:


apis/debug/state.v2.yaml
```diff
@@ -27,7 +27,7 @@ get:
               version:
                 type: string
                 enum: [phase0, altair, bellatrix, capella, deneb, electra, fulu]
-                example: "electra"
+                example: "fulu"
```
rolfyone, line 30: we'll definitely want the fulu state added in state v.2 in the data field

nflaig: good catch, looks like we completely missed updating the state for fulu which has the `proposer_lookahead`, we only have the new api https://github.com/ethereum/beacon-APIs/pull/539

let's do a separate PR for that

nflaig: - https://github.com/ethereum/beacon-APIs/pull/565

PR #568, opened 2025-12-22
Add getNodeVersionV2 endpoint

eth2353:
Adds a `/eth/v2/node/version` endpoint for structured version information that also includes information about the attached EL client.

Changes:
- add version information about the EL client attached to the beacon node
- explicitly structure version information mirroring how this information is exchanged on the Engine API: https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md
- deprecates the `/eth/v1/node/version` endpoint

The main motivation for this change is for validator clients to have access to information about the EL client attached to a beacon node. Right now, validator clients cannot see "past" the beacon node easily and are therefore blind to what kind of EL client is validating the execution payload. By exposing this information, multi-node validator clients can make safer decisions, e.g. they can require 3 different EL clients to declare the execution payload as valid before they attest to its validity.

### DIFF
--- a/apis/node/version.v2.yaml
+++ b/apis/node/version.v2.yaml
@@ -11,7 +11,7 @@ get:
     Version information about the execution client may not be available
     at all times and is therefore optional.
 
-    If the beacon node receives multiple values from the Engine API, the
+    If the beacon node receives multiple values from `engine_getClientVersionV1`, the
     first value should be returned on this endpoint.
   responses:
     "200":


types/node.yaml
```diff
@@ -0,0 +1,19 @@
+ClientVersionV1:
+  type: object
+  description: "A structure which uniquely identifies a client implementation and its version. Mirrors the client version specification in the [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md)."
```
nflaig: would suggest to use perma links here (either commit hash or tag if it exists)


```suggestion
  description: "A structure which uniquely identifies a client implementation and its version. Mirrors the client version specification in the [Engine API](https://github.com/ethereum/execution-apis/blob/c4988b1c427645d1b861c8c12488975f5a2f8cb9/src/engine/identification.md)."
```

and in other places

types/node.yaml
```diff
@@ -0,0 +1,19 @@
...
+  properties:
+    code:
+      $ref: './primitive.yaml#/ClientCode'
+    name:
+      type: string
+      description: "Human-readable name of the client, e.g. Lighthouse or go-ethereum"
+      example: "Lodestar"
+    version:
+      type: string
+      description: "the version string of the current implementation e.g. v4.6.0 or 1.0.0-alpha.1 or 1.0.0+20130313144700"
```
nflaig:
```suggestion
      description: "The version string of the current implementation e.g. v4.6.0 or 1.0.0-alpha.1 or 1.0.0+20130313144700"
```

wordlist.txt
```diff
@@ -37,3 +37,5 @@ KZGProofs
 KZGCommitmentInclusionProof
 LMD
 fulu
+LH
```
nflaig: do we need to add this here, the new spelling rule should ignore it

apis/node/version.v2.yaml
nflaig, line 1: need to add changelog entries for new api and deprecation note for v1 api

types/primitive.yaml
```diff
@@ -17,6 +17,12 @@ Version:
   description: "A string which uniquely identifies the client implementation and its version; similar to [HTTP User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3)."
   example: "Lighthouse/v0.1.5 (Linux x86_64)"
 
+ClientCode:
+  type: string
+  description: "A two-character client code that uniquely maps to a client name as defined in the [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md#clientcode)."
+  enum: [BU, EJ, EG, GE, GR, LH, LS, NM, NB, TE, TK, PM, RH]
```
nflaig, line 23: might be worth to specify what to return if client code is unknown, we have a non-spec value of `XX` for this

types/node.yaml
```diff
@@ -0,0 +1,19 @@
+ClientVersionV1:
+  type: object
+  description: "A structure which uniquely identifies a client implementation and its version. Mirrors the client version specification in the [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/identification.md)."
+  required: [code, name, version, commit]
+  properties:
+    code:
+      $ref: './primitive.yaml#/ClientCode'
+    name:
```
nflaig, line 8: might be worth to specify what `name` to return if information is missing, eg. if we could not retrieve the EL client info yet, I am just returning `Unknown` right now

apis/node/version.v2.yaml
```diff
@@ -0,0 +1,39 @@
...
+      content:
+        application/json:
+          schema:
+            title: GetVersionV2Response
+            type: object
+            required: [data]
+            properties:
+              data:
+                type: object
+                required: [beacon_node, execution_client]
```
mehdi-aouadi: I'd rather make the `execution_client` optional. See my comment https://github.com/ethereum/beacon-APIs/pull/568#issuecomment-3860880099

nflaig:
I am also fine with making fields optional if that's what everyone prefers to do

apis/node/version.v2.yaml
```diff
@@ -0,0 +1,51 @@
...
+            properties:
+              data:
+                type: object
+                required: [beacon_node]
+                properties:
+                  beacon_node:
+                    $ref: '../../beacon-node-oapi.yaml#/components/schemas/ClientVersionV1'
+                  execution_client:
+                    # Version information about the execution client may not be available at all times
+                    # and is therefore not required.
```
nflaig: maybe adding something similar to this as description is more useful, usually try to avoid comments like these as they are not visible when looking at the api explorer

types/primitive.yaml
```diff
@@ -17,6 +17,12 @@ Version:
   description: "A string which uniquely identifies the client implementation and its version; similar to [HTTP User-Agent](https://tools.ietf.org/html/rfc7231#section-5.5.3)."
   example: "Lighthouse/v0.1.5 (Linux x86_64)"
 
+ClientCode:
+  type: string
+  description: "A two-character client code that uniquely maps to a client name as defined in the [Engine API](https://github.com/ethereum/execution-apis/blob/dc4dbca37ef8697d782f431af19120beaf5517f5/src/engine/identification.md#clientcode)."
+  enum: [BU, EJ, EG, GE, GR, LH, LS, NM, NB, TE, TK, PM, RH, XX]
```
mehdi-aouadi: The `XX` should be removed here too

apis/node/version.v2.yaml
eth2353: Makes sense, I moved it to the endpoint description 👍

apis/node/version.v2.yaml
```diff
@@ -0,0 +1,56 @@
+get:
+  operationId: "getNodeVersionV2"
+  tags:
+    - Node
+  summary: "Get version information for the beacon node and execution client."
+  description: |
+    Retrieves structured information about the version of the beacon node
+    and its attached execution client in the same format as used on the
+    [Engine API](https://github.com/ethereum/execution-apis/blob/dc4dbca37ef8697d782f431af19120beaf5517f5/src/engine/identification.md)
+    .
```
nflaig: why is the `.` on a new line?

CHANGES.md
```diff
@@ -8,8 +8,10 @@ There are likely to be descriptions etc outside of the list below, but new query
...
-| [#563](https://github.com/ethereum/beacon-APIs/pull/563) `GET /eth/v1/validator/duties/proposer/{epoch}` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#563](https://github.com/ethereum/beacon-APIs/pull/563) `GET /eth/v2/validator/duties/proposer/{epoch}` added      |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#563](https://github.com/ethereum/beacon-APIs/pull/563) `GET /eth/v1/validator/duties/proposer/{epoch}` deprecated |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig: the diff here doesn't look right, why does it show them as deleted and re-added?

apis/node/version.v2.yaml
eth2353: Mistake on my side, fixed.

CHANGES.md
eth2353: My IDE does this kind of auto-formatting where it aligns table columns in the markdown by adding spaces, leading to that diff. I undid this change so now the diff only contains the added rows.

nflaig (dismissed):
LGTM

CHANGES.md
nflaig: note to self: changelog format is annoying, we need to simplify that, client specific columns are unused, also would be nice if we can have a way for this to not conflict between branches

james-prysm (approved):
seems good to us too, good discussion on null vs optional always good to clearly define those

eth2353:
> the BN should use deterministic heuristic if it's completely blind: the first from the list is the primary/active one for example. Decided by the BN and not the consumer.

In case of the external multiplexer, the beacon node has no say over which client is used as primary/active, it doesn't know – the external multiplexer decides which EL node(s) processes each Engine API request. The multiplexer logic may be trivial (picking the first healthy node) but can also be much more complex (e.g. require 2 out of 3 EL clients to agree on payload validity).

The above is why I consider the hybrid approach interesting but insufficient to capture the full range of possibilities (from the point of view of _behavior during consensus bugs_ which Beacon API consumers like Vero would be interested in).

___

I believe we may have reached the point where there is no single correct answer what should be returned here.

There are many possible options once you consider a beacon node is connected to multiple EL clients and they depend on how the beacon node interacts with those multiple EL clients, whether that's through a [separate multiplexer](https://github.com/TennisBowling/executionbackup) or some [internal logic](https://nimbus.guide/eth1.html#running-multiple-execution-clients). Trying to capture the full nuance of that in this Beacon API endpoint seems difficult to me.

> Since the multiplexer use case isn't standard

The overlap between people running EL multiplexers and people interested in consuming this `getNodeVersionV2` Beacon API endpoint is probably near zero. The standard is to run a _client pair_ - one beacon node and one execution client.

Therefore **I propose we refocus on that standard in this PR and simplify by simply returning a single object for the `execution_client` field.**

If there's enough appetite in the future to work out the nuances for EL multiplexers, we can iterate on a v3 endpoint then but I think there's a good chance nobody will ever need that.

apis/node/version.v2.yaml
```diff
@@ -0,0 +1,55 @@
...
+  tags:
+    - Node
+  summary: "Get version information for the beacon node and execution client."
+  description: |
+    Retrieves structured information about the version of the beacon node
+    and its attached execution client in the same format as used on the
+    [Engine API](https://github.com/ethereum/execution-apis/blob/dc4dbca37ef8697d782f431af19120beaf5517f5/src/engine/identification.md).
+
+    Version information about the execution client may not be available
+    at all times and is therefore optional.
```
nflaig, line 12: do we wanna note if multiple values are returned from engine api that only the first one should be considered?

eth2353: Do you mean it this way - if the beacon node receives multiple values over the Engine API, it should return the first of those here on the Beacon API?

If so, yeah that sounds reasonable, can add a note.

nflaig: yes, so we have consistent behavior across clients

eth2353: 👍 done

apis/node/version.v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+    If the beacon node receives multiple values from the Engine API, the
```
nflaig: maybe to be more explicit

```suggestion
    If the beacon node receives multiple values from `engine_getClientVersionV1`, the
```

nflaig (approved):

PR #569, opened 2026-01-15
EIP-8025

frisitano:
Adds the EIP-8025 beacon-node HTTP API: a `GET` for retrieving execution proofs by block id, and a `POST` for validators to submit signed execution proofs which the node verifies, stores, and gossips. Both endpoints live under `/eth/v1/beacon/execution_proofs`.

Aligned with the [`eth-act/lighthouse`](https://github.com/eth-act/lighthouse) reference implementation. `MAX_PROOF_SIZE` is 400 KiB to match the consensus-specs bump in ethereum/consensus-specs#5162.

Related CL specs: ethereum/consensus-specs#5055 (refactor, merged), ethereum/consensus-specs#5162 (size bump, merged). Original CL PR: ethereum/consensus-specs#4828 (merged).


apis/beacon/proofs/execution_proofs.yaml
```diff
@@ -0,0 +1,39 @@
+get:
+  operationId: getExecutionProofs
+  summary: Get execution proofs
+  description: |
+    Retrieves execution proofs known by the node.
+  parameters:
+    - name: new_payload_request_root
```
kevaundray: Why new_payload_request_root versus beacon root?

apis/prover/execution_proofs.yaml
```diff
@@ -0,0 +1,30 @@
...
+  tags:
+    - Prover
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+            $ref: '../../beacon-node-oapi.yaml#/components/schemas/EIP8025.ExecutionProof'
```
jsign: Just leaving a comment here if we need to add `ProofGenId` for proper connection from the request? (mentioned [here](https://github.com/ethereum/consensus-specs/pull/4828#discussion_r2699140636))

apis/prover/execution_proofs.yaml
```diff
@@ -0,0 +1,30 @@
+post:
+  operationId: submitExecutionProofs
```
jsign: Maybe basic question from my side: should [this](https://github.com/ethereum/consensus-specs/pull/4828/files#diff-ba43a2e25d530a702d5b293279b105851e9bce11a6afe7a3964643388d7e0519R137) have the same name?

apis/beacon/proofs/execution_proofs.yaml
frisitano: Good question. I had used `new_payload_request_root` as that had been the theme of this batch of PR's. Now I take a moment to think, I see value in using the beacon root here and also for p2p req/resp in the CL specs. What is your opinion?

kevaundray: I lean more towards beacon root since this is used in a lot of other places like DataColumnsByRoot

frisitano: I agree, I will update it. Does it also make sense to migrate to beacon root for p2p req/resp?

apis/prover/execution_proofs.yaml
frisitano: replied in CL [PR](https://github.com/ethereum/consensus-specs/pull/4828#discussion_r2699397772).

apis/beacon/proofs/execution_proofs.yaml
kevaundray: Checking the flow, I think having it in p2p requires/resp also makes sense in terms of simplicity -- my thinking was that the CL will need to have a map from beacon root to new_payload_request_root or whatever the public input is for the proof anyways

apis/prover/execution_proofs.yaml
frisitano: They are in different contexts, so it's fine to have different names imo. Above, we have the HTTP endpoint that the prover receives requested proofs from the proof engine. Linked is the p2p req/res api. I think it's reasonable to have different naming.

apis/beacon/proofs/execution_proofs.yaml
frisitano: Agreed. I think the beacon root is generally more available. For example, in the context of performing sync we are more likely to have the beacon root available than the new_payload_request_root.

apis/beacon/execution_proofs/execution_proofs.yaml
```diff
@@ -0,0 +1,47 @@
...
+  operationId: submitExecutionProofs
+  summary: Submit signed execution proofs
+  description: |
+    Receives validator-signed execution proofs (typically from a validator client).
+    The beacon node verifies each proof's BLS signature, requests the proof engine
+    to validate the proof against the associated `NewPayloadRequest`, stores accepted
+    proofs, and broadcasts them on the `execution_proof` gossip topic.
+
+    Proofs whose proof engine verification status is `VALID` or `ACCEPTED` are gossiped.
+    Proofs reported as `INVALID` cause a `400` response. `SYNCING` and `NOT_SUPPORTED`
```
jsign: Nit:
```suggestion
    Proofs reported as `INVALID` cause a `400 Bad Request` response. `SYNCING` and `NOT_SUPPORTED`
```
Maybe since last sentence below has this style to reference HTTP status codes?

apis/beacon/execution_proofs/execution_proofs_by_block_id.yaml
```diff
@@ -0,0 +1,64 @@
+get:
+  operationId: getExecutionProofs
+  summary: Get execution proofs
+  description: |
+    Retrieves signed execution proofs for a given block id.
+    Depending on `Accept` header it can be returned either as JSON or as bytes serialized by SSZ.
+
+    Proofs are returned as an ordered list of `SignedExecutionProof` objects signed by validators.
```
jsign: Nit: worth saying ordered by which field?

apis/beacon/execution_proofs/execution_proofs_by_block_id.yaml
```diff
@@ -0,0 +1,64 @@
...
+      description: "Block not found"
+      content:
+        application/json:
+          schema:
+            $ref: "../../../beacon-node-oapi.yaml#/components/schemas/ErrorMessage"
+          example:
+            code: 404
+            message: "Block not found"
+    "406":
+      $ref: "../../../beacon-node-oapi.yaml#/components/responses/NotAcceptable"
```
jsign, line 60: Worth explaining more here or in description when this happens?
I suspect it is related to the `Accept` field in the request?

jsign (approved):
LGTM, just left some nits for your consideration.

apis/beacon/execution_proofs/execution_proofs_by_block_id.yaml
frisitano: Good catch. I decided to remove the `ordered` term, as it surfaces an implementation detail from the lighthouse proof engine, but the proof engine is specifically specified as implementation-dependent in the consensus specs. I have now added terminology to explicitly state that clients MUST NOT depend on the order.

apis/beacon/execution_proofs/execution_proofs_by_block_id.yaml
frisitano: You are correct its related to the `Accept` field. The convention in these specs across all methods is to just have a bare `$ref`, so I think we should maintain that standard here.

PR #577, opened 2026-02-06
Remove blob sidecars endpoint and event

nflaig:
This removes the previously deprecated blob sidecars endpoint and event.
- `GET /eth/v1/beacon/blob_sidecars/{block_id}` removed 
- `blob_sidecar EVENT` removed (this event is no longer emitted since Fulu)

The new endpoint to retrieve blobs is `GET /eth/v1/beacon/blobs/{block_id}` (see [getBlobs](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlobs) for details) which is already supported by all CL clients.

**Note:** CL clients should keep supporting these until the next hard fork (gloas) to avoid breaking any integrations.


james-prysm (approved):
I think this makes sense, please leave it open for some time to get more feedback . we should keep the endpoint around until gloas I think

PR #552, opened 2025-09-03
Standardized Epbs Beacon Api

shane-moore:
Since the epbs [consensus specs](https://ethereum.github.io/consensus-specs/specs/gloas/beacon-chain/) are in good shape, we can formalize the beacon api endpoints.

### Updated:
- `POST /eth/v2/beacon/blocks`
  -  Publishes a signed gloas `BeaconBlock` to the network

### New:
**ExecutionPayloadBid**
-  `GET /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}`
   -  Retrieves the builder’s execution payload bid so it can be signed
- `POST /eth/v1/beacon/execution_payload_bid`
  -  Publishes a signed execution payload bid to the network, and the bid must pass gossip validation for success

**PTC**
- `POST /eth/v1/validator/duties/ptc/{epoch}`
  -  Retrieves Payload Timeliness Committee (PTC) duties for a given epoch and validator indices
- `GET /eth/v1/validator/payload_attestation_data/{slot}`
  - Produces the payload attestation data for a given slot, to be signed by the validator if in the PTC that slot
- `POST /eth/v1/beacon/pool/payload_attestations`
  -  Submits a signed payload attestation message for gossip and pool inclusion
- `GET /eth/v1/beacon/pool/payload_attestations`
  -  Retrieves payload attestations currently in the beacon node’s operation pool
  
  ## TBD
  - may need a new endpoint that can be used to signal creation and gossiping of `DataColumnSidecar` by whoever's bid is included in the consensus block or perhaps this can be triggered after the call to release a `SignedExecutionPayloadEnvelope`
  - ~~may need to later separate the `GET /eth/v1/validator/payload_attestation_data/{slot}` and `POST /eth/v1/beacon/pool/payload_attestations` into two endpoints if we end up doing dual-ptc deadlines, which would result in separate payload attestation objects, one for `payload_present` and another for `blob_data_available`~~

PR #580, opened 2026-02-12
Produce block v4 with payload

shane-moore:
Introduces the block production and execution payload envelope endpoints for post-Gloas (ePBS) forks. These were separated out from #552 to be discussed and reviewed independently. Prior discussion in [PR 552 thread](https://github.com/ethereum/beacon-APIs/pull/552#discussion_r2642624992) and [shane-moore/beacon-APIs#2](https://github.com/shane-moore/beacon-APIs/pull/2) as well.

## New endpoints:
 - **`GET /eth/v4/validator/blocks/{slot}`** — Post-Gloas block production endpoint (`produceBlockV4`). Supports an `include_payload` query parameter that controls whether the execution payload envelope and blobs are returned inline (stateless multi-BN operation) or cached by the beacon node for separate retrieval (stateful single-BN operation).
 - **`GET /eth/v1/validator/execution_payload_envelopes/{slot}/{beacon_block_root}`** — Fetch the cached execution payload envelope for the current proposer's self-built payload, returned as a `BlindedExecutionPayloadEnvelope` (the minimal object the validator client signs over). `beacon_block_root` is a required path param so the lookup is re-org resistant. Only available for the current slot; returns 404 for past slots or when no envelope has been cached. Used in stateful mode.
 - **`POST /eth/v1/beacon/execution_payload_envelopes`** — Publish a signed execution payload envelope. The request body is selected by the `Eth-Execution-Payload-Blinded` header: `false` carries `SignedExecutionPayloadEnvelopeContents` (envelope + blobs + KZG proofs, stateless flow); `true` carries a `SignedBlindedExecutionPayloadEnvelope` that the beacon node reconstructs from its cache before broadcasting (stateful flow).

## Blinded execution payload envelope
Folds in shane-moore/beacon-APIs#10, resolving the earlier open question on whether a blinded envelope should be used in the stateful flow. `BlindedExecutionPayloadEnvelope` replaces the `payload` field with `payload_root` (the hash tree root of the full execution payload). Its hash tree root matches the full `ExecutionPayloadEnvelope`, so a validator signature over the blinded form is valid against the full envelope. This keeps the object the VC signs minimal in the stateful flow, where the beacon node already has the full payload and blobs cached from block production. Discussion: [r3310000141](https://github.com/ethereum/beacon-APIs/pull/580#discussion_r3310000141), [r3277279908](https://github.com/ethereum/beacon-APIs/pull/580#discussion_r3277279908).


apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,67 @@
...
+  summary: Get execution payload envelope
+  description: |
+    Retrieves execution payload envelope for a given slot and builder, which can then be signed by a validator. The envelope contains the full
+    execution payload along with associated metadata. Depending on `Accept` header, it can be returned
+    either as json or as bytes serialized by SSZ.
+  tags:
+    - Validator
+    - ValidatorRequiredApi
+  parameters:
+    - name: slot
```
nflaig, line 21: while implementing this, I noticed the slot alone is not definitive enough (eg. in case of forks), we should consider also passing the `beacon_block_root` here

apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,67 @@
...
+      in: path
+      required: true
+      description: "Slot for which the execution payload envelope is requested."
+      schema:
+        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+    - name: builder_index
```
james-prysm: Will this endpoint be used for all builders outside of self building? I think the pipeline might endup being very different which makes me question if we would ever be caching more than just the current slot's payload envelope. terence asked this internally and I didn't have a good answer.

nflaig: this endpoint is only meant for self-building if you are the proposer or the "act as a builder" mode where you build the payload for another proposer

builders that produce payloads outside of the beacon node would get their own endpoint which would be under the `/builder` namespace (at least that's how I am thinking about it right now)

it should be sufficient to only cache the current slot as you consume the cache within the same slot, maybe if that's the case my comment here https://github.com/ethereum/beacon-APIs/pull/580#discussion_r2804137298 is also not needed but I feel like the block root is useful to specify, not just the slot

nflaig: After thinking about it more, maybe we keep the current apis under `/validator` to be only used by proposers, which means we can remove `builder_index` here as it's only used for self-building

apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,67 @@
+get:
+  operationId: getExecutionPayloadEnvelope
+  summary: Get execution payload envelope
+  description: |
+    Retrieves execution payload envelope for a given slot and builder, which can then be signed by a validator. The envelope contains the full
```
james-prysm: can we specify this is not used for general payload envelope returns and they should expect a 404 if it's not for the current proposer? i.e. we aren't going to go get from execution for older slots

apis/validator/execution_payload_envelope.yaml
shane-moore: hm, what do you think about making it optional to pass `beacon_block_root` like in https://github.com/ethereum/beacon-APIs/pull/580/changes/0c458594da02947f60bf6c7f4cae876ed2cdb1f8? pretty sure the VC already computes the block's tree hash root during signing, so passing it here should be straightforward

apis/validator/execution_payload_envelope.yaml
shane-moore: makes sense, removed `builder_index` in https://github.com/ethereum/beacon-APIs/pull/580/changes/0c458594da02947f60bf6c7f4cae876ed2cdb1f8

apis/validator/execution_payload_envelope.yaml
shane-moore: agreed, https://github.com/ethereum/beacon-APIs/pull/580/changes/0c458594da02947f60bf6c7f4cae876ed2cdb1f8

apis/validator/execution_payload_envelope.yaml
nflaig: Yes, we do already compute the block root on the vc side. I am not a big fan of optional parameters, so I'd rather have it as required or remove it if it's not needed, so far it makes our implementation simpler, maybe get more feedback from others who already implemented the apis

apis/validator/block.v4.yaml
```diff
@@ -0,0 +1,154 @@
...
+      description: |
+        Percentage multiplier to apply to the builder's payload value when choosing between a
+        builder payload header and payload from the paired execution node. This parameter is only
+        relevant if the beacon node is connected to a builder, deems it safe to produce a builder
+        payload, and receives valid responses from both the builder endpoint _and_ the paired
+        execution node. When these preconditions are met, the server MUST act as follows:
+
+        * if `exec_node_payload_value >= builder_boost_factor * (builder_payload_value // 100)`,
+          then return a full (unblinded) block containing the execution node payload.
+        * otherwise, return a blinded block containing the builder payload header.
```
bharath-123: I don't think this applies with gloas anymore. Regardless of whether we self-build or fetch a bid from a builder. We will get back a `SignedExecutionPayloadBid` in the beacon block. In the case of the builder, the envelope won't be available with the validator.

apis/validator/block.v4.yaml
```diff
@@ -0,0 +1,154 @@
...
+        * `builder_boost_factor=0`: prefer the local execution node payload unless an error makes it
+          unviable.
+        * `builder_boost_factor=100`: profit maximization mode; choose whichever
+          payload pays more.
+        * `builder_boost_factor=2**64 - 1`: prefer the external builder payload unless an error or
+          beacon node health check makes it unviable.
+
+        Servers should use saturating arithmetic or another technique to ensure that large values of
+        the `builder_boost_factor` do not trigger overflows or errors. If this parameter is
+        provided and the beacon node is not configured with a builder then the beacon node MUST
```
bharath-123: I guess with ePBS, the beacon node is always configured with a builder as it listens to the builder bids on the gossip topic anyways.

apis/validator/block.v4.yaml
nflaig: this needs to be rephrased for gloas, I think I brought that up in the other PR as well

apis/validator/block.v4.yaml
nflaig: not necessarily, we currently have an option to only produce local blocks which I intend to keep, but the default behavior will likely change with gloas to either use bid from p2p or build local, there are still valid reasons to force the node to build a local block, eg. builders censoring, which especially until FOCIL is shipped requires local blocks or a really sophisticated logic in clients to filter out bids from censoring builders

apis/beacon/execution_payload/envelope_post.yaml
```diff
@@ -0,0 +1,53 @@
+post:
+  operationId: publishExecutionPayloadEnvelope
+  summary: Publish signed execution payload envelope
+  description: |
+    Instructs the beacon node to broadcast a signed execution payload envelope to the network,
+    to be gossiped for payload validation. A success response (20x) indicates
+    that the envelope passed gossip validation and was successfully broadcast onto the network.
```
ensi321, line 7: In [`publishBlockV2`](https://github.com/ethereum/beacon-APIs/blob/d8c98590a4380720252f64c1042a178f7c1d3940/apis/beacon/blocks/blocks.v2.yaml#L8) we return 202 if block passed gossip validation but failed integration. 

Are we doing something similar for `publishExecutionPayloadEnvelope`?

ensi321: I guess from another angle to think about this is we broadcast the payload regardless the result of the payload integration, or we integrate first, if it goes well, we then gossip?

michaelsproul: We probably want to propagate the envelope as quickly as possible, so ideally:

1. Apply gossip validation
2. Broadcast
3. Run `process_execution_payload` and import to DB

If (1) fails, we 400. If (3) fails we should probably 202 just like for blocks.

---

I was hoping we could avoid the nastiness of `broadcast_validation` for envelopes, but I think there might actually be an unbundling attack ([background](https://blog.sigmaprime.io/mev-unbundling-rpc.html)) for envelopes now:

I think there is an unbundling attack something like:

1. Malicious proposer publishes block `A` at slot `N` committing to `bidA` and `payloadA`, where `bidA` comes from an external builder (could be P2P or otherwise).
2. Malicious proposer publishes block `B` (slashable) at slot `N` committing to `bidB` and payload `payloadB`. They will be slashed at the next slot.
3. Builder for `bidA` starts to publish `payloadA`. Even on beacon nodes that have seen the equivocating block, the payload will PASS gossip validation as long as block A was imported.
4. Proposer publishes `payloadB` for block `B`. Envelope gossip validation passes for `payloadB` because it is signed by a different builder from `payloadA`.
5. Result: either A or B could become head, depending on timing. If B wins, the builder of A has revealed potentially valuable information, which could be worth the proposer slashing themselves for (they could use it for a self-build in the next slot).

Proposed fixes:

1. We could put the burden on the builder to check for block equivocations prior to revealing their payload (no API/spec changes).
2. We could add a gossip condition to reject envelopes for slashable blocks.
3. We could add a `broadcast_validation=equivocation` flag (like we have for pre-Gloas blocks) to check that the envelope's block is not an equivocation. This flag would likely be used by all builders when interacting with their BNs.
4. We could make the block equivocation check a default part of the API (same as `broadcast_validation=equivocation` from fix 3, but no flag required and no ability to opt-out).

apis/validator/execution_payload_envelope.yaml
james-prysm: @shane-moore coul you update the pr description too to reflect

shane-moore: good call, updated description!

apis/validator/block.v4.yaml
shane-moore: made minimal changes to the description to remove the blinded/unblinded references in https://github.com/ethereum/beacon-APIs/pull/580/changes/8bf908038b5aa3204dd99745bfb0e33f47823bc6

agreed `builder_boost_factor` needs to be rethought for gloas. we discussed briefly in the old PR [here](https://github.com/ethereum/beacon-APIs/pull/552#discussion_r2652167439) and briefly in the [discord](https://discord.com/channels/595666850260713488/874767108809031740/1455332856602689628)

I also found this builder-specs [PR](https://github.com/ethereum/builder-specs/pull/144) defining how BN can set preferences between external bids/p2p bis/VC bids

The question is whether we want to be able to override these BN preferences via the VC. I'm sure the answer is yes, and the question becomes "how"?

Would we want to modify the validator registrations [endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/registerValidator) to signal a VC's per builder bid boost preferences similar to as proposed in the builder spec PR? Pretty sure this POST fires every epoch or when one of the values change

This could replace `builder_boost_factor` for example

Open to other ideas as well

apis/validator/execution_payload_envelope.yaml
shane-moore: I feel you, perhaps @eserilev @tersec @StefanBratanov or others from your teams have some thoughts here

apis/beacon/execution_payload/envelope_post.yaml
shane-moore: based off the discord [convo](https://discord.com/channels/595666850260713488/692078615269212180/1479010145663979611), seems like a `broadcast_validation` flag is warranted. I built it out in https://github.com/ethereum/beacon-APIs/pull/580/changes/b9b6fc62949362a478e0e571e19d506789727d29 to use same `BroadcastValidation` schema as `produceBlocksV2` since all the possible values seem to make sense for envelopes as well, but lmk if have some other thoughts on it

wheelercameron193-hub (approved):

apis/validator/execution_payload_envelope.yaml
nflaig: so looks like we might wanna use this for builders too if the validator "acts as a builder"

~~worth revisiting if we wanna re-add the `builder_index` here~~

**Edit:**

we might be fine, I think it's worth to wait until someone actually implement the "act as a builder" flow e2e, I rather tend towards not mixing up apis required by validators and builders too much

see [discord](https://discord.com/channels/595666850260713488/874767108809031740/1488539668218904596) for conversation about this

apis/validator/execution_payload_envelope.yaml
shane-moore: @nflaig, thought more about this. made `beacon_block_root` a required field in https://github.com/ethereum/beacon-APIs/pull/580/changes/5413e57627938cfe842e37d9c7f687309a31917e since just passing the slot alone isn't sufficient under potential fork equivocations. open to feedback if folks have a different stance on this

tersec: Yes, prefer required to optional fields.

kvqddbvyk6-cyber: So what exactly are you guys doing

kvqddbvyk6-cyber: Trying to get my deposit from my beaconcli

rolfyone:
is this sorted now?

shane-moore:
> is this sorted now?

this pr could be merged as is imo. not sure if others think otherwise.  my lingering question in https://github.com/ethereum/beacon-APIs/pull/580#discussion_r2897481552 is about what to do with `builder_boost_factor` going forward since it doesn't make sense as is, but it probably doesn't make sense to spec it out in beacon api spec yet until builder api spec for how boosting certain builder's bid will work, i.e. https://github.com/ethereum/builder-specs/pull/144.  I could make a follow-up gh issue for this if preferable

wheelercameron193-hub:
well Idk What's going on no one is actually telling me what going on and my keystore is now gone

rolfyone:
sounds like the outstanding issues from discord are
 - builder_boost_factor
 - stateful endpoint
at least from @tersec - not sure if these are shared concerns but flagging here.  Given this is required for gloas I'd be interested in seeing progress on this api to unblock gloas but we'll need to reconcile these issues in some way.

apis/beacon/execution_payload/envelope_post.yaml
jimmygchen: I'm just catching up on the discussions here, but I think the `broadcast_validation` flag and its values make sense to me.

If im understanding correctly, the `broadcast_validation =consensus_and_equivocation` value on the **publish_block** endpoint will no longer be useful for unbundling protection? Probably no harm to keep it though.

jimmygchen:
> * stateful endpoint
>   at least from @tersec - not sure if these are shared concerns but flagging here.  Given this is required for gloas I'd be interested in seeing progress on this api to unblock gloas but we'll need to reconcile these issues in some way.

@rolfyone could you elaborate the stateful endpoint issue? AFAICT the functionality is covered in this PR, or am I missing something?

**EDIT**: never mind, i see the R&D discussion now.

bharath-123:
> > is this sorted now?
> 
> this pr could be merged as is imo. not sure if others think otherwise. my lingering question in [#580 (comment)](https://github.com/ethereum/beacon-APIs/pull/580#discussion_r2897481552) is about what to do with `builder_boost_factor` going forward since it doesn't make sense as is, but it probably doesn't make sense to spec it out in beacon api spec yet until builder api spec for how boosting certain builder's bid will work, i.e. [ethereum/builder-specs#144](https://github.com/ethereum/builder-specs/pull/144). I could make a follow-up gh issue for this if preferable

`builder_boost_factor` probably needs to be rethought with ePBS. We probably need a `builder_boost` for p2p and builder-api bids or sth like that? 

I think builder-specific flags like builder_boost, min_bid etc can be specified in the builder-specs. The community did want some standardization of these flags. I can update https://github.com/ethereum/builder-specs/pull/144 to include all the necessary flags and I think the builder-specs is probably the right place to keep these builder based configurations.

apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,78 @@
...
+        404 if there is no match. Requiring the block root makes the lookup re-org resistant:
+        if a re-org occurs between block production and envelope retrieval, the validator client
+        will receive a 404 rather than an envelope committed to a stale block.
+      schema:
+        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Root"
+  responses:
+    "200":
+      description: "Successful response"
+      headers:
+        Eth-Consensus-Version:
```
james-prysm, line 42: we don't need this header on a get request right?

apis/beacon/execution_payload/envelope_post.yaml
```diff
@@ -0,0 +1,83 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          anyOf:
+            - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedExecutionPayloadEnvelopeContents"
+            - $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedExecutionPayloadEnvelope"
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `SignedExecutionPayloadEnvelopeContents` or `SignedExecutionPayloadEnvelope` bytes. Use Content-Type header to specify this format"
```
james-prysm: how are you checking which one was passed the contents of the payload envelope , i'm checking ssz offsets right now, but wondering if it makes sense to just add something in the header to separate the two

ensi321: > how are you checking which one was passed the contents of the payload envelope , i'm checking ssz offsets right now, but wondering if it makes sense to just add something in the header to separate the two

Yea Lodestar ran into the same problem. I think we should have `Eth-Blob-Data-Included` or something to distinguish the two.

apis/validator/execution_payload_envelope.yaml
nflaig: the header is in the response, and from what I can see, it's needed

kvqddbvyk6-cyber (approved):

kvqddbvyk6-cyber (approved):

apis/beacon/execution_payload/envelope_post.yaml
nflaig: > If im understanding correctly, the broadcast_validation =consensus_and_equivocation value on the publish_block endpoint will no longer be useful for unbundling protection?

no, it's not needed for this anymore, we could discuss if we wanna deprecate `broadcast_validation` on `publishBlockV2`, but might still be useful to have basic checks before publishing the block

apis/beacon/execution_payload/envelope_post.yaml
nflaig: We can either have a separate endpoint or add a header to differentiate both, ideally for the local stateful flow we just wanna pass a blinded execution payload envelope around to get the most out of it as passing the `transactions` around is not useful. If we go with `BlindedExecutionPayloadEnvelope` we could also nicely reuse the existing `Eth-Execution-Payload-Blinded` header here

apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,78 @@
...
+            title: GetExecutionPayloadEnvelopeResponse
+            type: object
+            required: [version, data]
+            properties:
+              version:
+                type: string
+                enum: [gloas]
+                example: "gloas"
+              data:
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.ExecutionPayloadEnvelope"
```
nflaig: I would be in favor of just returning a `BlindedExecutionPayloadEnvelope` here, that's the minimal object we need to pass to the validator client to sign over

apis/validator/block.v4.yaml
```diff
@@ -0,0 +1,155 @@
...
+        bandwidth but requires the validator client to publish via the same beacon node that
+        produced the block (stateful operation).
+
+        This parameter only affects self-building scenarios. When using an external builder's bid,
+        only the beacon block is returned regardless of this parameter (the beacon node does not
+        have access to the builder's execution payload).
+      schema:
+        type: boolean
+        default: true
+    - name: builder_boost_factor
```
nflaig, line 66: @bharath-123 if a client doesn't have builders configured via api, we don't wanna submit builder preferences at all, so moving boost factor and other parameters there seem wrong to me, we need a way for the validator to signal this to the beacon node. And I don't think the beacon api should be concerned about off protocol builders or per builder configs

apis/validator/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,78 @@
...
+  parameters:
+    - name: slot
+      in: path
+      required: true
+      description: "Slot for which the execution payload envelope is requested."
+      schema:
+        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+    - name: beacon_block_root
+      in: query
+      required: true
```
nflaig, line 29: I think we should move this to `path` instead

apis/validator/block.v4.yaml
nflaig: > Would we want to modify the validator registrations [endpoint](https://ethereum.github.io/beacon-APIs/#/Validator/registerValidator) to signal a VC's per builder bid boost preferences similar to as proposed in the builder spec PR? Pretty sure this POST fires every epoch or when one of the values change

We want to get rid of this endpoint, by default, there won't be any registrations or builder preferences submitted to the beacon node, we only submit proposer preferences.

I left a comment related to this here https://github.com/ethereum/beacon-APIs/pull/580#discussion_r3310068662. In my opinion, the beacon-api spec should have no notion of per builder configuration. The `builder_boost_factor` is still useful for the validator client to signal to the beacon node how it should decide on whether to self-build or pick a bid from the p2p pool, and the bid from the p2p pool is simply the one with the highest `value`.

shane-moore: agree on keeping `builder_boost_factor` and not introducing per-builder config in the beacon-api. but i'm not sure the pre-Gloas formula carries over cleanly

pre-Gloas, the BN got `exec_node_payload_value` from its paired EL and compared it against `builder_payload_value`

in Gloas the bid's `value` field is the builder→proposer payment, and the spec mandates that self-builds set `bid.value` to zero. so if we keep the formula `local_bid_value >= builder_boost_factor * (external_bid_value // 100)` using bid `value` fields:

- `factor=0` → returns local (0 ≥ 0)
- `factor=100` → returns external (0 ≥ external_value is false)
- `factor=2**64-1` → returns external

`100` and `2**64-1` seem to collapse to the same meaning

do we still expect the BN to somehow make `factor=100` (profit max) meaningful? if yes, the formula stays but we should say so explicitly. if no, maybe we switch `builder_boost_factor` to a boolean like `prefer_local=true|false`

shane-moore:
just pushed [`b613dde`](https://github.com/ethereum/beacon-APIs/pull/580/commits/b613dde): the envelope's `payload` was still referencing `Deneb.ExecutionPayload`, missing `block_access_list` (EIP-7928) and `slot_number` (EIP-7843). new `Gloas.ExecutionPayload` adds them.

apis/beacon/execution_payload/envelope_post.yaml
shane-moore: an idea of how we could do the `BlindedExecutionPayloadEnvelope` and reuse `Eth-Execution-Payload-Blinded` header to differentiate https://github.com/shane-moore/beacon-APIs/pull/10

apis/validator/execution_payload_envelope.yaml
shane-moore: done in [`7c1c6b1`](https://github.com/ethereum/beacon-APIs/pull/580/commits/7c1c6b1)

apis/validator/execution_payload_envelope.yaml
shane-moore: agreed, done in https://github.com/shane-moore/beacon-APIs/pull/10

nflaig:
> just pushed [`b613dde`](https://github.com/ethereum/beacon-APIs/pull/580/commits/b613dde): the envelope's `payload` was still referencing `Deneb.ExecutionPayload`, missing `block_access_list` (EIP-7928) and `slot_number` (EIP-7843). new `Gloas.ExecutionPayload` adds them.

we've just merged https://github.com/ethereum/beacon-APIs/pull/607 which was already reviewed and had more updates for alpha.8, you will need to resolve conflicts here .. sorry

wheelercameron193-hub (approved):

apis/validator/block.v4.yaml
nflaig: just to clarify this, we wanna compare the local payload value vs. the builder bid (the highest bid we have, could even include bids from api), the changes to `builder_boost_factor` description should be minimal, it isn't really different from what we do today, let me know @shane-moore if that clears things up

apis/validator/block.v4.yaml
```diff
@@ -0,0 +1,155 @@
...
+      required: false
+      description: |
+        Percentage multiplier to apply to the builder's bid value when choosing between a
+        builder bid and payload from the paired execution node. This parameter is only
+        relevant if the beacon node is connected to a builder, deems it safe to produce a builder
+        payload, and receives valid responses from both the builder endpoint _and_ the paired
+        execution node. When these preconditions are met, the server MUST act as follows:
+
+        * if `exec_node_payload_value >= builder_boost_factor * (builder_payload_value // 100)`,
+          then return a block committing to the local bid (including execution node payload if `include_payload` parameter is set to true)
```
nflaig: I think the confusing part here is the term "local bid" and framing builder bids as "external". The better way to phrase this is local payload (or " local execution node payload" as in v3) vs. builder bid

shane-moore: Done in 411435069c1e0962de5534fff7e2ea7ffd51c970. Description now uses "local execution node payload" and "builder bid" throughout

apis/validator/block.v4.yaml
shane-moore: thanks, that clears it up. Done in 411435069c1e0962de5534fff7e2ea7ffd51c970. Preconditions now read "has at least one viable builder ExecutionPayloadBid available" as to cover p2p and builder api derived bids. Open to any other tweaks!

beacon-node-oapi.yaml
```diff
@@ -220,6 +224,8 @@ paths:
     $ref: "./apis/validator/liveness.yaml"
   /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}:
     $ref: "./apis/validator/execution_payload_bid.yaml"
+  /eth/v1/validator/execution_payload_envelope/{slot}/{beacon_block_root}:
```
nflaig: just realized something, shouldn't this be plural to match other apis like `/eth/v4/validator/blocks/{slot}`?

```suggestion
  /eth/v1/validator/execution_payload_envelopes/{slot}/{beacon_block_root}:
```

I noticed same goes for `/eth/v1/beacon/execution_payload_envelope/{block_id}` which we already merged

beacon-node-oapi.yaml
```diff
@@ -106,6 +106,8 @@ paths:
     $ref: "./apis/beacon/blocks/blocks.v2.yaml"
   /eth/v1/beacon/execution_payload_bid:
     $ref: "./apis/beacon/execution_payload/bid.yaml"
+  /eth/v1/beacon/execution_payload_envelope:
```
nflaig: same here, this should be
```suggestion
  /eth/v1/beacon/execution_payload_envelopes:
```

to match `/eth/v2/beacon/blocks`

and also `/eth/v1/beacon/execution_payload_bid` seems like it should be plural (already merged that)

shane-moore: Done, pluralized both new envelope paths. Left the already-merged `execution_payload_bid` and `execution_payload_envelope/{block_id}` GET as-is since they are out of scope here, but since you noted bid should be plural too, how do you want to handle those: pluralize them in this PR, do a separate follow-up PR, or open a tracking issue? Happy either way, and easy to revert the plural here if anyone prefers singular.

beacon-node-oapi.yaml
shane-moore: Pluralized here too.

beacon-node-oapi.yaml
nflaig: I opened https://github.com/ethereum/beacon-APIs/pull/613 for existing endpoints not part of this PR, let's get more feedback from others there

wheelercameron193-hub (approved):

beacon-node-oapi.yaml
nflaig: @shane-moore we've agreed on #613 and I have just merged it, I think we can go ahead and apply the same update of the paths here

shane-moore: Nice! Added in 947f91d

apis/beacon/execution_payload/envelope_post.yaml
shane-moore: https://github.com/shane-moore/beacon-APIs/pull/10 is merged into this branch now

apis/validator/execution_payload_envelope.yaml
shane-moore: https://github.com/shane-moore/beacon-APIs/pull/10 is merged to this branch

shane-moore:
folded in shane-moore#10, which adopts a `BlindedExecutionPayloadEnvelope` (replaces payload with `payload_root`) for the stateful publish flow. The `GET` now returns the blinded envelope and the `POST` selects the body via the `Eth-Execution-Payload-Blinded` header

rolfyone:
@nflaig intentional to close this? there isnt a v4/blocks in the dev instance...

james-prysm:
I don't think this was intentional lol i will reopen

james-prysm (approved):
I'm voting ship it, i think it now covers everything, please don't merge until at least another person reviews

nflaig:
> @nflaig intentional to close this? there isnt a v4/blocks in the dev instance...

ok this is crazy, yesterday I rebased my [prysm fork](https://github.com/nflaig/prysm) and pushed the new commits which include [James' commit](https://github.com/nflaig/prysm/commit/9069afc6d0275ed2877e1750a8350e028ab9bb8b) that has the `Fixes https://github.com/ethereum/beacon-APIs/pull/580` and this caused github to close the PR, @github what are you doing, sort your stuff out, there is no way this should happen

potuz:
I left some comments about `builder_boost_factor` and how I think it should be treated here https://github.com/ethereum/beacon-APIs/issues/620

shane-moore:
> I left some comments about `builder_boost_factor` and how I think it should be treated here https://github.com/ethereum/beacon-APIs/issues/620
> 
> 

Hey folks, I'm OoO til July 4th getting married 🤵👰

Glanced at #620, and I had similar concerns about builder_boost_factor. I leave it to y'all to decide whether to merge this PR as is and handle #620 as a follow-up or reach consensus on #620 approach first and then someone pushes an impl to this PR in my absence. Have no strong preference either way

potuz:
I'd rather merge this quickly (and also try to reach consensus on 620 quickly). Congratulations on the wedding!

rolfyone:
Ok let's just merge this and address `builder_boost_factor` in #620 as suggested.

nflaig (commented):
LGTM, great work @shane-moore and congrats on the wedding!

wheelercameron193-hub (approved):

types/gloas/execution_payload_envelope.yaml
```diff
@@ -28,3 +28,60 @@ Gloas:
...
+    required: [signed_execution_payload_envelope, kzg_proofs, blobs]
+    properties:
+      signed_execution_payload_envelope:
+        $ref: "#/Gloas/SignedExecutionPayloadEnvelope"
+      kzg_proofs:
+        $ref: "../fulu/block_contents.yaml#/Fulu/KZGProofs"
+      blobs:
+        $ref: "../deneb/block_contents.yaml#/Deneb/Blobs"
+
+  BlindedExecutionPayloadEnvelope:
```
nflaig, line 47: - will be removed in https://github.com/ethereum/beacon-APIs/pull/624 in case anyone is looking at the diff here to implement this

PR #582, opened 2026-02-16
Update getDebugDataColumnSidecars for Gloas

### LINKED ISSUE #574: Update data column sidecar for gloas
as per title, but let's wait a bit until the spec stabilizes, there have just been changes to `DataColumnSidecar` in [v1.7.0-alpha.2](https://github.com/ethereum/consensus-specs/releases/tag/v1.7.0-alpha.2)

mehdi-aouadi:
Update `getDebugDataColumnSidecars` for Gloas

Closes https://github.com/ethereum/beacon-APIs/issues/574


types/gloas/data_column_sidecar.yaml
```diff
@@ -0,0 +1,33 @@
...
+  DataColumnSidecars:
+    type: array
+    items:
+      $ref: '#/Gloas/DataColumnSidecar'
+    minItems: 0
+    maxItems: 128
+
+  DataColumnSidecar:
+    type: object
+    description: "A [`DataColumnSidecar`](https://github.com/ethereum/consensus-specs/blob/e8b77415ffce13ee2b14a5c22417a1f77f1dca34/specs/gloas/p2p-interface.md?plain=1#L57) as defined in the Gloas consensus spec."
```
nflaig: would be better to use tag and not refer to plain, I don't think we do that usually

apis/debug/data_column_sidecars.yaml
```diff
@@ -45,7 +45,9 @@ get:
               finalized:
                 $ref: "../../beacon-node-oapi.yaml#/components/schemas/Finalized"
               data:
-                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Fulu.DataColumnSidecars"
+                anyOf:
+                  - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Fulu.DataColumnSidecars"
```
nflaig: latest fork should be listed first

nflaig: version enum above needs to be updated too

types/gloas/data_column_sidecar.yaml
mehdi-aouadi: pointed to v1.7.0-alpha.2

apis/debug/data_column_sidecars.yaml
mehdi-aouadi: done

nflaig (dismissed):
LGTM

nflaig:
@mehdi-aouadi did you also look at the event?

https://github.com/ethereum/beacon-APIs/blob/5b46a66d9ee8c932876b2c1a3a68ebac97c307e1/apis/eventstream/index.yaml#L153-L157

worth thinking about if we wanna update that for gloas and remove `kzg_commitments` there too, although the event type is different from spec type

I don't think it's difficult for clients to emit the KZG commitments there in gloas as they should have them available from the bid

mehdi-aouadi:
> @mehdi-aouadi did you also look at the event?
> 
> https://github.com/ethereum/beacon-APIs/blob/5b46a66d9ee8c932876b2c1a3a68ebac97c307e1/apis/eventstream/index.yaml#L153-L157
> 
> worth thinking about if we wanna update that for gloas and remove `kzg_commitments` there too, although the event type is different from spec type
> 
> I don't think it's difficult for clients to emit the KZG commitments there in gloas as they should have them available from the bid

imo we'd better remove them from the event too since they live elsewhere in the bid

nflaig:
> imo we'd better remove them from the event too since they leave elsewhere in the bid

worth asking ethpandops maybe, the problem with just removing them is that it breaks the format which is not ideal but they are considered debug, so might be fine

maybe break this out in a separate PR?

mehdi-aouadi:
> > imo we'd better remove them from the event too since they leave elsewhere in the bid
> 
> worth asking ethpandops maybe, the problem with just removing them is that it breaks the format which is not ideal but they are considered debug, so might be fine
> 
> maybe break this out in a separate PR?

Makes sense. Let's keep the scope clean. Will raise a separate PR

nflaig (approved):
looks good, going to merge this as changes are straightforward and aligned with latest CL spec

PR #583, opened 2026-02-24
remove kzg commitment from data column sidecar event

mehdi-aouadi:
Remove the `kzg_commitments` from the `data_column_sidecar` event since they have been removed from [DataColumnSidecar](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#modified-datacolumnsidecar). It might be expensive to get them from the bid.


nflaig:
>  It might be expensive to get them from the bid.

the spec requires to run [this check](https://github.com/nflaig/consensus-specs/blob/a21e27dd861e770368df3a16e393efe1ec2e0efb/specs/gloas/p2p-interface.md?plain=1#L410-L411)
> _[REJECT]_ The sidecar is valid as verified by `verify_data_column_sidecar(sidecar, bid.blob_kzg_commitments)`.

since the event is emitted after gossip validation, the kzg commitments should be available

that said, I am not against removing them, although there is no great way to version events right now, but we did change the format of the `attestation` event in the past across forks so considering limited usage of this event it might be fine here too

but I would expect we'd just remove them for slots after gloas

mehdi-aouadi:
> the spec requires to run [this check](https://github.com/nflaig/consensus-specs/blob/a21e27dd861e770368df3a16e393efe1ec2e0efb/specs/gloas/p2p-interface.md?plain=1#L410-L411)
> 
> > _[REJECT]_ The sidecar is valid as verified by `verify_data_column_sidecar(sidecar, bid.blob_kzg_commitments)`.
> 
> since the event is emitted after gossip validation, the kzg commitments should be available

Theoretically, from a pure spec perspective, yes but from an implementation perspective that might add complexity since it requires adding them to the container or transmitting them alongside for this sole and only reason. That gossip check already added complexity and imo we should keep the event bus as cheap as possible

nflaig (approved):
in favor of removing them for slots after gloas as it aligns with the CL spec type

PR #584, opened 2026-03-06
Builder Api to construct payload envelope

### LINKED ISSUE #575: Builder api to construct envelope from their execution payload
I think this API assumes that the execution payload and blobs are built locally. This might not be the case for a builder who builds the block using custom block building software. The builder will have the execution payload, blobs bundle and execution requests ready with them, they will require a beacon-api which will construct the ExecutionPayloadEnvelope for them to sign.

It would be desirable to also be able to pass in the bid, beacon block, execution payload, blobs bundle, execution requests to get the ExecutionPayloadEnvelope corresponding to the bid sent by the builder.

_Originally posted by @bharath-123 in https://github.com/ethereum/beacon-APIs/pull/552#discussion_r2769231258_

bharath-123:
This PR attempts to resolve https://github.com/ethereum/beacon-APIs/issues/575

We introduce a builder prefixed api called `ConstructExecutionPayloadEnvelope` which takes in the beacon block root of the given slot, execution payload and execution requests. It returns an `ExecutionPayloadEnvelope` which will have to be signed by the builder and published to the PTC.


apis/builder/get_execution_payload_envelope.yaml
```diff
@@ -0,0 +1,65 @@
...
+  operationId: getExecutionPayloadEnvelope
+  summary: Get execution payload envelope for signing
+  description: |
+    Builder-facing endpoint that accepts the components needed to construct an
+    `ExecutionPayloadEnvelope`. The beacon node computes the `beacon_block_root`
+    from the supplied `SignedBeaconBlock` and derives the post-execution
+    `state_root`, returning a complete `ExecutionPayloadEnvelope` ready for the
+    builder to sign and broadcast.
+  tags:
+    - Beacon
```
nflaig: I think this tag needs to be removed

apis/builder/get_execution_payload_envelope.yaml
```diff
@@ -0,0 +1,65 @@
+post:
+  operationId: getExecutionPayloadEnvelope
```
nflaig: need a different id here, this one is already taken, maybe `constructExecutionPayloadEnvelope`

beacon-node-oapi.yaml
```diff
@@ -149,6 +151,9 @@ paths:
   /eth/v1/beacon/pool/bls_to_execution_changes:
     $ref: "./apis/beacon/pool/bls_to_execution_changes.yaml"
 
+  /eth/v1/builder/get_execution_payload_envelope:
```
nflaig:
```suggestion
  /eth/v1/builder/execution_payload_envelope:
```

apis/builder/get_execution_payload_envelope.yaml
bharath-123: `constructExecutionPayloadEnvelope` is great

potuz:
Why does this endpoint need the full block? if it's just for verification it should need just the bid, payload and requests,  isn't it?

apis/builder/execution_payload_envelope.yaml
```diff
@@ -0,0 +1,64 @@
...
+          title: ConstructExecutionPayloadEnvelopeRequest
+          type: object
+          required: [signed_beacon_block, execution_payload, execution_requests]
+          properties:
+            signed_beacon_block:
+              $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedBeaconBlock"
+            execution_payload:
+              $ref: "../../beacon-node-oapi.yaml#/components/schemas/Deneb.ExecutionPayload"
+            execution_requests:
+              $ref: "../../beacon-node-oapi.yaml#/components/schemas/Electra.ExecutionRequests"
```
nflaig: this needs to be wrapped into `data` container, we also wanna `version` the data in request body

bharath-123:
> Why does this endpoint need the full block? if it's just for verification it should need just the bid, payload and requests, isn't it?

Not sure what you mean by verification here.

This endpoint is for the builders to construct the full `ExecutionPayloadEnvelope` with the post beacon state root if their bid gets included in the block. I think the beacon block root would be required to construct the correct envelope corresponding to the beacon block.

The builder can sign the envelope externally.

I think passing the beacon block root, payload and execution requests should be good?

nflaig:
this is no longer required due to https://github.com/ethereum/consensus-specs/pull/5094, builders should be able to easily construct the `ExecutionPayloadEnvelope` themselves as it is no longer required to do a state transition to compute the `state_root`

bharath-123:
Yes I will close this PR for now!

PR #585, opened 2026-03-06
Add execution block hashes in chain reorg SSE

bharath-123:
With gloas, there can be a payload reorg where the beacon block stays the same but the payload associated with the block changes. 

It might make sense to add the old and new execution block hashes of the head blocks in the chain_reorg event


michaelsproul:
This approach is similar to option 3 from:

- https://github.com/ethereum/beacon-APIs/issues/589

nflaig:
> This approach is similar to option 3 from:
> 
> * [Update `head` event in backwards compatible way #589](https://github.com/ethereum/beacon-APIs/issues/589)

I am not in favor of option 3. there but I think adding new fields to this event is fine as they are distinct from existing fields.

nflaig:
> but the payload associated with the block changes.

what exactly do you mean here? the payload associated with a beacon block will always be the same based on `bid.block_hash` but a payload reorg can happen where it changes from FULL to EMPTY (or vice versa)

or do you mean we should emit `chain_reorg` also if a payload reorg happens? in that case I think the meaning of `depth` becomes overloaded

Maybe it's best if we add a separate event for payload reorgs which is emitted when the beacon block is unchanged and only the payload status / EL head changes.

michaelsproul:
I think using the same reorg event for block/payload reorgs is OK, so long as we specify that the depth relates only to the beacon block depth.

In a sense the `block_hash` for an empty slot is the most recently applied payload, the hash of which is easy to obtain from `block.signed_execution_payload_bid.message.parent_block_hash`.

Is that what you had in mind @bharath-123?

nflaig:
I am a bit skeptical about this still, is there anyone that wants this or has implemented it? I can takeover @bharath-123's PR and get it merged, but I haven't looked at the usefulness or implementation of doing this change at all.

also we are not using `chain_reorg` in our validator client, so I don't really know what are all the use cases for this. but for any use cases from a validat client pov, I don't see any usefulness to emit this in a case of a payload reorg as it doesn't change anything, also you can only reorg a payload via the next block proposal, so the block root changes anyways? maybe I am misunderstanding something about how/when this would be emitted

PR #586, opened 2026-03-11
Remove next epoch lookahead from PTC duties

nflaig:
The PTC assignments are not stable for the next epoch so we should not be required to serve these duties via the api as we will need to re-compute them for the current epoch anyways. This PR effectively removes the requirement for clients to serve next epoch PTC duties.

Related https://github.com/ethereum/consensus-specs/pull/4992


rolfyone (approved):
LGTM

satushh (approved):

james-prysm (approved):

PR #587, opened 2026-03-11
Add version to gloas events

nflaig:
All events that emit data based on spec containers should be wrapped by `{"version": "fork_name", "data": { ... }}` so we can easily change them across forks and make it easier for consumers as they can just look at the `version` field and know the format/schema based on the spec container for that fork. This change overall makes the events more forward compatible.


rolfyone (approved):
LGTM

michaelsproul (approved):
Looks good.

I've been meaning to open an issue about updating the `head` event in a backwards-compatible way. The dependent root fields are wrong since Fulu

PR #588, opened 2026-03-11
Add `execution_payload` and `execution_payload_gossip` event

nflaig:
This adds `execution_payload` and `execution_payload_gossip` to the eventstream.

The events mirror the existing `block` and `block_gossip` events.

 - `execution_payload_gossip` is emitted when an execution payload passes gossip validation
 - `execution_payload` is emitted when an execution payload is successfully imported
 - `execution_payload_available` is emitted when an execution payload and all data are available for payload attestation without verifying the execution payload itself

These events enable measuring execution payload timings via sidecar software such as [contributoor](https://github.com/ethpandaops/contributoor).


rolfyone:
we should probably add versions to this too...

apis/eventstream/index.yaml
```diff
@@ -155,6 +157,16 @@ get:
               value: |
                 event: data_column_sidecar
                 data: {"block_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "index": "1", "slot": "1"}
+            execution_payload:
+              description: The node has received a `SignedExecutionPayloadEnvelope` (from P2P or API) that is successfully imported on the fork-choice `on_execution_payload` handler
```
ensi321: May I know the difference between this and `execution_payload_available`?
This one emits when the payload is fully imported, `execution_payload_available` emits when payload is available for PTC vote. 
Is there a gap/delay between the occurrence of these two events?

nflaig:
> we should probably add versions to this too...

was thinking about that, I am not that convinced that versioning here makes sense as the schema is not based on a spec container

apis/eventstream/index.yaml
nflaig: ~~I guess you are right, there shouldn't be much of a difference in the timing and would at most be minimal based on how the client specific import pipeline works.~~

~~We can probably remove it and only stick to `execution_payload_gossip` and `execution_payload_available`~~

`execution_payload_available` is emitted when the execution payload + all data (based on custody) is available, it does not require the execution payload to be valid, while `execution_payload` does since it is emitted after successful import into fork choice via `on_execution_payload` handler, so it includes the time it takes for the EL to execute and verify the execution payload

apis/eventstream/index.yaml
```diff
@@ -38,6 +38,8 @@ get:
             - light_client_optimistic_update
             - payload_attributes
             - data_column_sidecar
+            - execution_payload
```
nflaig, line 41:
```suggestion
```

rolfyone:
> was thinking about that, I am not that convinced that versioning here makes sense as the schema is not based on a spec container

still, we might be better off being consistent?

nflaig:
> still, we might be better off being consistent?

but that means not having the version/data container here? eg. `block_gossip` doesn't have that either because the event doesn't emit data based on a spec container while evens in https://github.com/ethereum/beacon-APIs/pull/587 do, so having the version there makes sense

I am open to adding a version here too if that's really what we want

rolfyone:
@michaelsproul thoughts?

dknopik:
I'm not Michael, but implementing this in Lighthouse. I agree with @nflaig, let's version only the directly embedded consensus objects. Versioning the others by fork name as well would be inconsistent as noted above, and would be strange to use in cases where we e.g. want to add some field to an event outside of the fork lifecycle.

james-prysm:
haven't started implementing this yet but looks good to me

nflaig:
applied changes due to https://github.com/ethereum/consensus-specs/pull/5094, the events no longer emit the `state_root`

rolfyone (approved):

Issue #589, opened 2026-03-11
Update `head` event in backwards compatible way


michaelsproul:
Since the dependent root changes in Fulu, the `head` event's dependent roots are misleadingly named.

The current pre-Fulu definitions are:

- `previous_duty_dependent_root = get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
- `current_duty_dependent_root = get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`

These are the correct _proposer_ dependent roots **prior** to Fulu, i.e. the dependent root for epoch `N`, is the last block of epoch `N - 1`.

With Fulu, these dependent roots are completely wrong. The proposer dependent root for epoch `N` is now the block at the end of epoch `N - 2`. The spec's extant dependent roots are also unrelated to the attestation dependent roots, which follow the same cadence as the new proposer dependent roots (i.e. epoch `N` is determined at the end of `N - 2`).

The correct post-Fulu definitions are:

- `previous_epoch_dependent_root = get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 2) - 1)`
- `current_epoch_dependent_root = get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
- `next_epoch_dependent_root = get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`

Note that we can now offer lookahead for the next epoch for both proposer and attester duties.

## Proposed Fixes

There are several ways to fix this without breaking the current events endpoint.

### Option 1: New `/eth/v2/events` endpoint

Perhaps the most drastic: we add a new `/eth/v2/events` endpoint with a new `head` event that is _versioned_ (similar to https://github.com/ethereum/beacon-APIs/pull/587), and we take the opportunity to internally version all other events for future proofing. If we introduce this endpoint before Gloas, we can remove `/eth/v1/events` at the Heze fork boundary.

### Option 2: New event type under v1

If we don't want to add a whole new `/eth/v2/events` endpoint, we could add a new event type under the existing event stream, with the Fulu changes, and a future-proof versioned structure. There are several things we could name this new event without conflict with existing events. Some ideas:

- `new_head`: a double meaning on "new" – new in the sense of being newly made head, _and_ new in the sense of replacing the old `head` event
- `head_updated`: kind of fine
- `head_block`
- `head_v2`: ugly IMO
- etc

[...]

Issue #572, opened 2026-01-27
Update state v2 api for gloas


nflaig:
I think we need some clarification on the /eth/v2/debug/beacon/states/{state_id} api too.
With ePBS we gonna have 2 state updates per slot, first when the block is processed and second when the payload is processed (if revealed).
It is currently not specified which state is returned by the endpoint, so I guess the response changes once the payload is revealed?

_Originally posted by @pk910 in https://github.com/ethereum/beacon-APIs/issues/552#issuecomment-3794809811_

rolfyone:
so pre-state and post payload state would need to be retrieved using the hash, and the 'post state' is what would be a number OR hash...
We have this limitation today, I've not looked at getting a pre-state for a while because i generally just wind it forward from previous slot, but in theory its do-able...

maybe what we need is at least a way to determine all the states in a slot or the pre-state hash of a given slot and payload-applied hash of a given slot or something... I'm not sure how complicated that would be to implement :) but at least then you'd have the hash to lookup...

I do agree we'd want to define whats returned at a given label, I think it's never been consistent for 'finalized', but i could be wrong...

ensi321:
> It is currently not specified which state is returned by the endpoint, so I guess the response changes once the payload is revealed?

I guess we also need to gather PTC votes and attestation votes to see if they support the fact that payload is available.

I think a more correct approach would be to consult our fork choice, see if the current chain that we are tracking has payload at `slot`, if so then we return the state after payload processing.

potuz:
I'll leave here my preferences on this endpoint so that we do not have to hack this behavior locally. And I believe these are the right options anyway: 

- For identifiers like blockroot I think the consensus post-state is the right option.  The reason being there may be an actual CL block for that blockroot and no payload.  
- For identifiers like slot I also think that the consensus post-state is the right option. The reason being that there may be a CL block at that slot and no payload for that slot. This situation is minimally more ambiguous than the previous one when during that slot the block is missed and the last state was full, in this case one could argue that we could return the head state before that slot advanced to the slot. But this complexity is rather useless, it does not save the caller *any* time to understand what they are getting. So sticking to the simple algorithm: "find out what's the latest block root before  or equal than that slot, call the previous helper for that blockroot, is the simplest and most elegant option". 
- For identifiers like finalized and justified: I think that the consensus post-state is the right option. Again the reason is that at the given blockroot/slot/start of the epoch there may be a block without a payload. Same reasoning applies for missed slots as above. 
[...]

PR #590, opened 2026-03-19
Add `head_v2` event and deprecate `head` event

chong-he:
This PR adds a `head_v2` event to the event stream as per discussion in #589


apis/eventstream/index.yaml
```diff
@@ -170,6 +171,11 @@ get:
...
+              description: The node has finished processing, resulting in a new head. previous_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 2) - 1)`, current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow. This `new_head` event is for post-Fulu slots. For pre-Fulu slots, use the `head` event.
+              value: |
+                event: new_head
+                data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "epoch_transition":false, "previous_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "current_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "next_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "execution_optimistic": false}
```
rolfyone: we should probably add version tag in this event @nflaig thoughts?

michaelsproul: I think Nico was pushing back on the `version` idea [here](https://github.com/ethereum/beacon-APIs/issues/589#issuecomment-4045362102):

> I am not convinced to use {"version": "fork_name", "data": { ... }} everywhere, that is, if they don't emit data based on spec containers, or not fork specific, it feels a bit unnecessary. Although since we change the head event now, we never know. It's just not consistent with format of existing events. When we go with v2 at some point, then yes, maybe it makes sense to wrap all events in the version container.

Although I agree with you, we may as well version it for future proofing.

michaelsproul (approved):

apis/eventstream/index.yaml
nflaig: not against adding a `version` here but in this case it's not trivial for a consumer to interpret it, while in the spec container case it's simple, you just pick the fork container based on `version` and use it to deserialize the data

while here, if we have to change the event in the future I'd expect we likely change the semantics of the event, not just the data format in which case a new event likely makes sense anyways

and clients should not break if new fields are added to the event, ie. https://github.com/ethereum/beacon-APIs/pull/585 should be totally fine, if a client can't handle that, it needs to be fixed on that end imo

apis/eventstream/index.yaml
```diff
@@ -170,6 +171,11 @@ get:
               value: |
                 event: payload_attestation_message
                 data: {"version":"gloas", "data":{"validator_index": "123", "data": {"beacon_block_root": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "slot": "10", "payload_present": true, "blob_data_available": true}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
+            new_head:
+              description: The node has finished processing, resulting in a new head. previous_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 2) - 1)`, current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow. This `new_head` event is for post-Fulu slots. For pre-Fulu slots, use the `head` event.
```
nflaig: do we wanna update the duty apis as part of this PR? currently their description still refers to the old `head` event

https://github.com/ethereum/beacon-APIs/blob/b4abce4f97266d75bc7922be80398f5c1a0939df/apis/validator/duties/attester.yaml#L15-L17

apis/eventstream/index.yaml
chong-he: Added the `version` in response

apis/eventstream/index.yaml
chong-he: Thanks, I have made the changes similar to `proposer.v2.yaml`. But I am not sure if I am doing it right?

apis/validator/duties/attester.yaml
```diff
@@ -10,17 +10,29 @@ post:
...
+     - event.block otherwise
 
-    The dependent_root value is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)`
-    or the genesis block root in the case of underflow."
+     - dependent_root value is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`
+
+
+    After Fulu:
+
+     - event.previous_epoch_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
```
chong-he: I use the term `previous_epoch_dependent_root` here (`instead of previous_duty_dependent_root`) as I thought after Fulu it is "epoch", not "duty". Is this right? If so, maybe I can also update this part in proposer.v2.yaml:

https://github.com/ethereum/beacon-APIs/blob/b4abce4f97266d75bc7922be80398f5c1a0939df/apis/validator/duties/proposer.v2.yaml#L26

Edit: Should it actually be `current_epoch_dependent_root` instead of `previous_epoch_dependent_root` for post-Fulu case?

apis/validator/duties/attester.yaml
```diff
@@ -10,17 +10,29 @@ post:
...
+
+     - event.block otherwise
+
```
chong-he: ~~If it is `previous_epoch_dependent_root`, should this be `epoch -2` instead of `epoch -1`?~~

Edit: this is obsolete, refer to the comment above instead

chong-he (commented):
Added some questions that I have

apis/eventstream/index.yaml
```diff
@@ -51,7 +52,7 @@ get:
             description: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#Event_stream_format
           examples:
             head:
-              description: The node has finished processing, resulting in a new head. previous_duty_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and current_duty_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. Both dependent roots use the genesis block root in the case of underflow.
+              description: Deprecated, use head_v2 instead.
```
nflaig:
```suggestion
              description: Deprecated, use `head_v2` instead.
```

apis/eventstream/index.yaml
```diff
@@ -41,6 +41,7 @@ get:
             - execution_payload_available
             - execution_payload_bid
             - payload_attestation_message
+            - head_v2
```
nflaig: can we add that next to `head` at the top, same for the definition below, I think this makes it easier if those events are next to each other

chong-he: moved in [61ffaec](https://github.com/ethereum/beacon-APIs/pull/590/commits/61ffaecc544df89196542a6fc291b0f9c6603118)

apis/eventstream/index.yaml
```diff
@@ -169,7 +175,7 @@ get:
               description: The node has received a `PayloadAttestationMessage` that passes validation rules of the `payload_attestation_message` topic
               value: |
                 event: payload_attestation_message
-                data: {"version":"gloas", "data":{"validator_index": "123", "data": {"beacon_block_root": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "slot": "10", "payload_present": true, "blob_data_available": true}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
+                data: {"version":"gloas", "data":{"validator_index": "123", "data": {"beacon_block_root": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "slot": "10", "payload_present": true, "blob_data_available": true}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
```
michaelsproul: Should we also take the opportunity to disambiguate the `slot` field? As far as I know, clients do different things with this field because it was never precisely specified whether it should be:

1. The slot of the head block matching `beacon_block_root`, or
2. The current wall-clock slot, or fork choice store slot

This makes it somewhat less useful for validator clients like Vouch/Vero, which might want to keep track of head block slots.

In `head_v2` we could fix this with two fields:

- `head_block_slot: Slot`
- `wall_clock_slot: Slot`

What do we think?

chong-he: To be clear, this is meant for the `head_v2` event response: https://github.com/ethereum/beacon-APIs/pull/590/changes#diff-3cc6eaa800c1a4bd6adcd78f1722b466f7a24048a66e320c8f92872fbcd9eefbR63

apis/validator/duties/attester.yaml
chong-he: After discussed with @michaelsproul , we revise this part to make it clearer:

- Remove before Fulu / after Fulu as now the chain is at Fulu
- rename to use `current_epoch...` instead of `current_duty...`
- remove `event.block otherwise` bullet point as it serves no purpose (and could be confusing)
- changes made to: `attester.yaml`, `proposer.v2.yaml` and `ptc.yaml`

apis/eventstream/index.yaml
eth2353: I was personally not aware of this ambiguity and assumed every CL client returned option 1 - "The slot of the head block matching `beacon_block_root`".

I don't see a reason to ever return a different slot there but I very well may be missing something. Quite the opposite, it seems like a bad idea to me to return dozens of head events with the same `slot` while syncing / catching up.

> As far as I know, clients do different things

Can we have CL client teams confirm what they do? This may not even be an issue if every client already does option 1 right now, and we just need to clarify that in the spec.

apis/eventstream/index.yaml
nflaig: the more I think about the `version` here, the more I dislike it, it's not really clear what it is "versioning" in that case because the `data` isn't really fork specific

apis/eventstream/index.yaml
nflaig: I would read that as `head.slot` so slot should definitely not be the wall clock slot, as for lodestar, we emit the slot of the head block

apis/eventstream/index.yaml
chong-he: I am ok to remove the `version`. The response in `data` is not fork-dependent as the fields `x_epoch_dependent_root` is fixed since Fulu now. What do others think?

rolfyone: i think version is probably more future proof, but im not going to argue if noone wants it... i do reserve the right to an 'i told you so' moment, this has been a recurring theme and it'll be back imo.

apis/eventstream/index.yaml
```diff
@@ -51,10 +52,15 @@ get:
...
+            head_v2:
+              description: The node has finished processing, resulting in a new head. previous_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 2) - 1)`, current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow.
+              value: |
+                event: head_v2
+                data: {"version": "gloas", "data":{"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "epoch_transition":false, "previous_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "current_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "next_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "execution_optimistic": false}}
```
michaelsproul: Just thought of another field we might want to add to this event, possibly _only_ after Gloas: the head's payload status.

With Gloas fork choice it's likely we'll get two head events every slot:

1. `block_root: X, payload_status: pending`
2. `block_root: X, payload_status: full`

If we do have `version`ing on `head_v2`, we could add the `payload_status` attribute just for Gloas. Or otherwise we could just default them to `pending` prior to Gloas (although there are arguments for pending, empty or full for pre-Gloas blocks 😅 ).

apis/eventstream/index.yaml
michaelsproul: Happy to just keep one `slot` field then and explicitly state that it must be the `slot` of the head block.

It's good there's no current ambiguity, I thought based on the discrepancy in `/eth/v1/node/syncing` (documented here: https://github.com/sigp/lighthouse/issues/3421), that there could be a discrepancy here too.

chong-he: `slot` definition added in [cdebb44](https://github.com/ethereum/beacon-APIs/pull/590/commits/cdebb44f3454d82fba47b400eca9b630aa9d45dc)

apis/eventstream/index.yaml
nflaig: I don't think "noone wants it", seems mostly me that is even arguing against it, but I don't feel strongly about it (even though I dislike it), so I would just go with keep the PR as is which includes the `version`, so whatever @chong-he prefers I am fine with.

>  i do reserve the right to an 'i told you so' moment

seems like the biggest risk to not including it, more so then any technical reason :grin:

apis/eventstream/index.yaml
nflaig: need to think more about it, but head seems unrelated to payload, what's the purpose or use case you had in mind?

regarding `pending`, I don't think we "leak" that anywhere currently outside of fork choice, so not sure that should be used anywhere else

if we would emit the event twice per slot, I would rather just do `full` / `empty` (if not received by ptc deadline), (I guess you would need `pending` for first time the event is emitted)

but maybe it should rather have `payload_status` as in status of the previous payload, maybe `previous_payload_status`


>  (although there are arguments for pending, empty or full for pre-Gloas blocks 😅 ).

I haven't seen a pending or empty block since the merge happened, using anything other than `full` for pre-gloas seems confusing to me, at least in lodestar we use full but it's a fork choice implementation detail anyways

michaelsproul: > need to think more about it, but head seems unrelated to payload, what's the purpose or use case you had in mind?

I don't think it's unrelated any more, the output of fork choice `get_head` now includes a payload status. I can imagine that some EL tools could be interested in knowing when certain payloads become enshrined as head. They can use the new `*payload` events, but I think they'll also want to know that the payload became canonical (not just that it exists).

> I haven't seen a pending or empty block since the merge happened, using anything other than full for pre-gloas seems confusing to me, at least in lodestar we use full but it's a fork choice implementation detail anyways

In Lighthouse we treat pre-Gloas blocks as pending in some places (no payload applied), and empty (no payload applied) :grin: But yeah I agree it's an impl detail, conceptually `full` (or no status) makes the most sense.

michaelsproul: IMO pending inevitably leaks into the state-based APIs, which have to now make the differentiation between block-states (pending) and payload-states (full).

nflaig: this kinda ties into discussion here https://github.com/ethereum/beacon-APIs/issues/572 which doesn't seem fully resolved, I would trust @twoeths opinion on this which from his last comment seems like we wanna be more explicit about payload status after gloas

so we probably wanna align what `state_id=head` returns from the state-based apis, and how the head event is emitted

in that case it might make sense to emit two head events per slot

chong-he: I updated the description to have the possibility of having two head events per slot after Gloas, given the `payload_status` could be received after.

From my understanding, it will be something like:

First head emitted event: 

`data: {"version": "gloas", "data":{"slot":"10", "block":"x", "state":"y", payload_status: empty, "epoch_transition": ...}}`
 
 Second head emitted event for the same slot  with full payload which changes the state:
 
`data: {"version": "gloas", "data":{"slot":"10", "block":"x", "state":"z", payload_status: full, "epoch_transition": ...}}`
 
If we do have the `payload_status` in the `head_v2` event, then I think we keep the `version` field.

apis/eventstream/index.yaml
chong-he: I retain the `version` field as per the discussion below: https://github.com/ethereum/beacon-APIs/pull/590#discussion_r3083629297

rolfyone:
happy to run with this if everyone else is now? @nflaig @michaelsproul ?

apis/eventstream/index.yaml
```diff
@@ -51,10 +52,15 @@ get:
...
+              value: |
+                event: head_v2
+                data: {"version": "gloas", "data":{"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "payload_status": "empty", "epoch_transition":false, "previous_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "current_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "next_epoch_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "execution_optimistic": false}}
```
nflaig: what's the purpose of having the `previous_epoch_dependent_root` in the head event?

chong-he: This is the equivalent field to the current head event field `previous_duty_dependent_root`, so I keep it in the response. Is it not required now?

chong-he: oh I think I got what you mean. The `previous_duty_dependent_root` in head is equivalent to the now `current_epoch_dependent_root` in `head_v2`, so we don't need to have `previous_epoch_dependent_root` in the event. Is that what you mean?

nflaig: yes, I don't see it being used or useful for anything, might be missing something, but generally speaking, we shouldn't need to worry about past epochs here as those duties are already passed

chong-he: makes sense, I removed in [aa984ea](https://github.com/ethereum/beacon-APIs/pull/590/commits/aa984ea51a6536980b45de6b9d394ae5ad1ab2bd)

apis/eventstream/index.yaml
nflaig: > First head emitted event:

what about `payload_status: pending`?

> Second head emitted event for the same slot with full payload which changes the state:

so we would not emit this if there was no payload for the slot? or would you just emit `empty` after some time in the slot, ie. if not observed after ptc deadline or maybe just if ptc voted for not available?

chong-he: Actually I am confused and not sure. Should the first emitted event (after receiving the beacon/consensus part of the block), should the `payload_status` be `pending`? This is from here: https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/fork-choice.md#modified-get_head

And if it starts with `pending` for the first head event, then we update to either `empty` or `full` in the second emission?

michaelsproul: I am now anti-pending, the output of `get_head` is always empty or full, because pending nodes always have at least one child (their empty node), so can never be returned by `get_head`.

michaelsproul:
Yep. LGTM.

apis/eventstream/index.yaml
```diff
@@ -54,10 +55,15 @@ get:
...
+              description: Deprecated, use `head_v2` instead.
               value: |
                 event: head
                 data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "epoch_transition":false, "previous_duty_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "current_duty_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "execution_optimistic": false}
+            head_v2:
+              description: The node has finished processing the beacon block, resulting in a new head with an empty `payload_status`. The node may emit a second head event for the same slot if there is an update in the `payload_status` from empty to full. `slot` is the slot of the head block. Current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow.
```
michaelsproul: The first sentence here reads a little awkwardly, although I do get the intent.

```suggestion
              description: The node's fork choice has selected a new head consisting of a beacon block and its `payload_status`. The node may emit a second head event for the same beacon block and slot if there is an update in the `payload_status` from empty to full. `slot` is the slot of the head block. Current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow.
```

Technically we could emit more than 2 events for the same beacon block and slot if there are weight-based reorgs between empty and full, but I think it's widely understood that this spec text is not 100% binding for implementations (I want to avoid a case where some implementation takes this too literally, or a consumer gets upset that the implementations do not follow this spec exactly, but I think that's kind of unlikely).

rolfyone (dismissed):
LGTM

apis/validator/duties/attester.yaml
```diff
@@ -9,14 +9,12 @@ post:
...
-    resulting in a change of duties. For full safety, you should monitor head events and confirm the
+    resulting in a change of duties. For full safety, you should monitor head_v2 events and confirm the
     dependent root in this response matches:
 
-      - event.previous_duty_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
+      - event.current_epoch_dependent_root when `compute_epoch_at_slot(event.slot) == epoch`
 
-      - event.current_duty_dependent_root when `compute_epoch_at_slot(event.slot) + 1 == epoch`
-
-      - event.block otherwise
```
nflaig, line 19: it's intended we removed this? I doubt any client uses `event.block`  and duties > 1 epoch in the future are not stable, so seems fine to me to remove but worth pointing out

michaelsproul: yeah I agree, fine to remove

chong-he: Yes we intend to remove this, some comments about it can be found here: https://github.com/ethereum/beacon-APIs/pull/590#discussion_r2993073913

nflaig: ah great, thanks for linking that comment I missed that

> remove event.block otherwise bullet point as it serves no purpose (and could be confusing)

yes, it is confusing indeed, +1 for removing it

apis/eventstream/index.yaml
```diff
@@ -54,10 +55,15 @@ get:
...
+              description: Deprecated, use `head_v2` instead.
               value: |
                 event: head
                 data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "state":"0x600e852a08c1200654ddf11025f1ceacb3c2e74bdd5c630cde0838b2591b69f9", "epoch_transition":false, "previous_duty_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "current_duty_dependent_root":"0x5e0043f107cb57913498fbf2f99ff55e730bf1e151f02f221e977c91a90a0e91", "execution_optimistic": false}
+            head_v2:
+              description: The node's fork choice has selected a new head consisting of a beacon block and its `payload_status`. The node may emit a second head event for the same beacon block and slot if there is an update in the `payload_status` from empty to full. `slot` is the slot of the head block. Current_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and next_epoch_dependent_root is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow.
```
nflaig:
```suggestion
              description: The node's fork choice has selected a new head consisting of a beacon block and its `payload_status`. The node may emit a second head event for the same beacon block and slot if there is an update in the `payload_status` from empty to full. `slot` is the slot of the head block. `current_epoch_dependent_root` is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch - 1) - 1)` and `next_epoch_dependent_root` is `get_block_root_at_slot(state, compute_start_slot_at_epoch(epoch) - 1)`. All dependent roots use the genesis block root in the case of underflow.
```

nflaig: noticed there is some ambiguity here what `epoch` is, this is already the case in the previous `head` event but might be worth to define that `epoch` is `compute_epoch_at_slot(slot)`

chong-he: Added epoch definition in [ccb3f0b](https://github.com/ethereum/beacon-APIs/pull/590/commits/ccb3f0b21f08be4448c8fc577b2c3ca3a908e586), thanks!

nflaig (approved):
LGTM

PR #591, opened 2026-03-19
clarification on index for produceAttestationData

rolfyone:
related to discussion in discord, I think this is probably a cleaner representation of the index in attestation data creation.

### DIFF
--- a/apis/validator/attestation_data.yaml
+++ b/apis/validator/attestation_data.yaml
@@ -27,8 +27,9 @@ get:
       deprecated: true
       description: |
         The committee index for which an attestation data should be created. For `slot`s in
-        Electra and later, this parameter MAY always be set to 0. For `slot`s in Gloas and
-        later, this parameter MAY be omitted.
+        Electra and later, this parameter MAY always be set to 0.
+        For `slot`s in Gloas and later, this parameter SHOULD be omitted, as the Beacon node
+        SHOULD be setting the index based on network rules.
       required: false
       schema:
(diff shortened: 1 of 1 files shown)


apis/validator/attestation_data.yaml
```diff
@@ -27,8 +27,9 @@ get:
       deprecated: true
       description: |
         The committee index for which an attestation data should be created. For `slot`s in
-        Electra and later, this parameter MAY always be set to 0. For `slot`s in Gloas and
-        later, this parameter MAY be omitted.
+        Electra and later, this parameter MAY always be set to 0.
+        For `slot`s in Gloas and later, this parameter SHOULD be omitted, as the Beacon node
+        SHOULD be setting the index based on network rules.
```
nflaig, line 32: do we really need the 2nd part here about the beacon node?

it's already in the description
https://github.com/ethereum/beacon-APIs/blob/b4abce4f97266d75bc7922be80398f5c1a0939df/apis/validator/attestation_data.yaml#L12-L13


the stronger wording from MAY to SHOULD seems fine to me

rolfyone: well, it seems like what we actually wanted was that the user shouldnt pass in index post gloas, ever. and if it *is* being passed in that its likely ignored. this was very unclear to me previously, so i think it should probably be made clear?

nflaig: but this is kinda the case since electra since the beacon node has to set `index=0`, either way this seems fine even if a bit repetitive (but doesn't hurt)

nflaig (approved):

PR #598, opened 2026-04-20
Add fast confirmation event

mkalinin:
Introduces `fast_confirmation` event with the following fields:
* `block`: A beacon block root output by the Fast confirmation rule.
* `slot`: A slot of the confirmed beacon block.

This event is supposed to be triggered after each execution of the Fast confirmation rule which should happen once per slot.

Fast confirmation rule is an algorithm that runs on top of the beacon chain consensus protocol and outputs a beacon block that won’t be reorged if network synchrony and adversarial threshold assumptions hold. Introduced by https://github.com/ethereum/consensus-specs/pull/4747

### Rationale
Fast confirmed block is a property of the beacon chain consensus protocol similar to finalized block and head which might be useful for consumers of the beacon chain data e.g. L2s. For JSON-RPC API consumers there is a “safe” block tag that denotes block hash of the most recent confirmed payload.


apis/eventstream/index.yaml
```diff
@@ -41,6 +41,7 @@ get:
             - execution_payload_available
             - execution_payload_bid
             - payload_attestation_message
+            - fast_confirmation
```
nflaig, line 46: we could add that next to the `finalized_checkpoint` event, but this is also fine

apis/eventstream/index.yaml
```diff
@@ -170,6 +171,11 @@ get:
               value: |
                 event: payload_attestation_message
                 data: {"version":"gloas", "data":{"validator_index": "123", "data": {"beacon_block_root": "0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "slot": "10", "payload_present": true, "blob_data_available": true}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
+            fast_confirmation:
+              description: The node has executed fast confirmation algorithm
+              value: |
+                event: fast_confirmation
+                data: {"confirmed_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "confirmed_slot": "1"}
```
nflaig: should this be just `slot` and `block_root` (or `block`) instead?

mkalinin: `confirmed_root` reflects `confirmed_root` in the `FastConfirmationStore` and `confirmed_slot` is simply in line with the root. I am looking at other `data` definitions in the event stream and thinking that `block`, `slot` would work better. Even though, `block_root` is more precise it’s better to be in line with the event stream notation.

mkalinin: switched to `block`, `slot`

apis/eventstream/index.yaml
rolfyone: this list is getting pretty big, we should sort them alphabetically maybe (can be another pr, just q for @nflaig )

nflaig: sure that works too, no strong opinion, appending at the end works too, I guess the order doesn't really matter

nflaig (approved):
LGTM

PR #601, opened 2026-04-28
Add SSZ response support for validators and validator_balances endpoints

### LINKED ISSUE #333: Support SSZ for validators endpoint

Zyra-V21:
## Motivation

The `/eth/v1/beacon/states/{state_id}/validators` endpoint currently returns JSON only. On mainnet today this is roughly 1 GB per state for the full validator set, while the equivalent SSZ payload is around an order of magnitude smaller and trivially cheaper to serialize and parse on both sides. The same applies to `validator_balances` (around 85 MB JSON per state).

For analytics and indexing tooling that snapshots validators frequently — for example [Migalabs](https://migalabs.io)' [goteth](https://github.com/migalabs/goteth) and similar consensus-layer indexers — JSON parsing and beacon-node CPU time during state snapshots are a real bottleneck. Exposing SSZ on these endpoints reduces both ends of the cost.

This proposal does not introduce a new design: it implements the agreement reached in the discussion of #333 (which has been open and unactioned since 2023).

## Scope

Adds `application/octet-stream` (SSZ) response support to:

- `GET /eth/v1/beacon/states/{state_id}/validators`
- `POST /eth/v1/beacon/states/{state_id}/validators`
- `GET /eth/v1/beacon/states/{state_id}/validators/{validator_id}`
- `GET /eth/v1/beacon/states/{state_id}/validator_balances`
- `POST /eth/v1/beacon/states/{state_id}/validator_balances`

Each affected endpoint also gains a `406 Not Acceptable` response, matching the pattern used by other SSZ-enabled endpoints in the spec (`debug/state.v2`, `pending_partial_withdrawals`, `validator_identities`, etc.).

## Validator status encoding

`ValidatorResponse` is wire-encoded as the SSZ container:

```
{ index: uint64, balance: uint64, status: byte, validator: Validator }
```

`Validator` is the consensus-spec object. `status` is a single `byte`. The byte mapping is documented in the `ValidatorStatus` schema and is the one proposed by @mcdee in #333:

| Value  | Status                |
| ------ | --------------------- |
| `0x00` | unknown               |
| `0x01` | pending_initialized   |
| `0x02` | pending_queued        |
| `0x03` | active_ongoing        |
| `0x04` | active_exiting        |
| `0x05` | active_slashed        |
| `0x06` | exited_unslashed      |
| `0x07` | exited_slashed        |
| `0x08` | withdrawal_possible   |
| `0x09` | withdrawal_done       |

`byte` is used rather than `uint8` per the discussion in #333 (cc @arnetheduck, @michaelsproul) to avoid implying the status is a numeric value with arithmetic semantics.

`ValidatorBalanceResponse` SSZ encoding is the simpler `{ index: uint64, balance: uint64 }` container; no new type definitions are required.

## What this PR does not change

- Existing JSON responses are untouched. Clients that do not advertise `application/octet-stream` continue to receive the same JSON they get today.
- No changes to request schemas, query parameters, or POST request bodies.
- No new endpoints, no deprecations.
- The existing `ValidatorStatus` JSON enum (`active_ongoing`, etc.) remains the canonical string form. The byte mapping is only relevant on the SSZ wire.

## Lint

`redocly lint beacon-node-oapi.yaml` reports the same baseline of pre-existing warnings/errors before and after this patch (84 errors, 15 warnings, all unrelated to the touched files). No new lint findings are introduced.

## Background and credits

The endpoint design, the choice of `byte` over `uint8`, and the status mapping all come from the discussion in #333. This PR is the implementation of that agreement; the credit for the design belongs to:

- @mcdee for opening #333, proposing the SSZ container shape, and the status byte mapping.
- @arnetheduck and @michaelsproul for the `byte` vs `uint8` clarification.
- @rolfyone for the consistency review around the embedded `Validator` object.
- @dapplion for the framing on debug/non-debug load semantics.

`validator_balances` was added because it shares the same indexer use case and avoids leaving an obviously asymmetric gap.

Related prior art that did not result in a merged spec change:
- ChainSafe/lodestar#6059 and ChainSafe/lodestar#6060 (closed; were waiting on the spec change this PR addresses).
- #367 added POST variants for these endpoints but kept JSON-only responses; the SSZ side was left open.

Closes #333.


types/api.yaml
```diff
@@ -53,6 +53,20 @@ ValidatorStatus:
...
+    | `0x01` | pending_initialized   |
+    | `0x02` | pending_queued        |
+    | `0x03` | active_ongoing        |
+    | `0x04` | active_exiting        |
+    | `0x05` | active_slashed        |
+    | `0x06` | exited_unslashed      |
+    | `0x07` | exited_slashed        |
+    | `0x08` | withdrawal_possible   |
+    | `0x09` | withdrawal_done       |
+
```
etan-status, line 69: mmh, would be cleaner to defer until SSZ receives proper enum support.

until then, would align validators/builders endpoints to match the same form (and keep JSON-only for now).

SSZ support doesn't need to be coordinated around a hardfork, client can indicate preferences via Accept header and server can just start shipping SSZ response once we decided on how to tackle enums.

Zyra-V21: Hey @etan-status, thanks for taking a look 🙏

I think we might actually be on the same page already: the PR doesn't use an SSZ enum type for status. It's encoded as a plain byte with the mapping just documented in the schema, exactly because @arnetheduck / @michaelsproul pushed back on `uint8`  in #333 to avoid implying any numeric/arithmetic semantics. So in that sense it's already "not depending on proper SSZ enum support" -> it's an opaque byte we agreed on out-of-band, not something the SSZ spec has to model.

So I want to make sure I understand what you'd prefer:

1. keep the opaque byte as-is (no SSZ enum machinery involved), or
2. drop status from the SSZ payload entirely for now and revisit once there's a proper enum story, or
3. hold the whole validators SSZ response as JSON-only until then?

I'd push back a bit on (3) : the main driver here is that the full validator set is ~1 GB of JSON per state on mainnet, which is a real bottleneck for indexers like goteth, so getting SSZ out is kind of the whole point. The byte was meant to be the minimal, non-controversial way to carry status without waiting on an enum decision.

One thing that might help unblock us regardless: validator_balances is just { index: uint64, balance: uint64 } : no enum, no status, nothing to defer. Happy to split that out so it can land independently while we settle the status question on the validators side.

And yeah, fully agree the validators/builders endpoints should end up consistent ;  I'll follow #610/#614 so whatever we land on for enums applies the same way across both. Lmk which direction you'd like and I'll adjust.

nflaig: I would prefer (3) because (1) would mean we add the format as proposed here into the spec, if the enum ssz type has a different byte mapping, it would be a breaking change and we need to do a v2, there is also a cost to implementing this is in a non-standard way, fun fact, I did this almost 3 years ago here https://github.com/ChainSafe/lodestar/pull/6059 but we never shipped it. But there was a lodestar branch that was able to return validators response in ssz format here https://github.com/ChainSafe/lodestar/pull/6060 and it worked. But even after almost 3 years, I don't think the solution as proposed in this PR is correct as much as I want the validator response to be ssz encoded. And lastly, regarding (2), this is obviously a breaking change, and not in favor of this, we discussed removing `status` before (many times actually) and there was never a rough consensus on it.

nflaig: Also @Zyra-V21 as a side note, can you please stop using claude to comment for you, it's fine to use AI to generate code, but responding to comments and forcing others to read through the AI slop I find personally a bit direspectul if I didn't know you this PR would be closed already.

Zyra-V21: quick note on provenance: the 0x00 unknown … 0x09 withdrawal_done mapping here isn't new, its @mcdee's from #333, and its the same byte encoding your own chainsafe/lodestar#6059 used, with an explicit "update to canonical once the spec defines it" todo. that's basically the posture im proposing too: ship the agreed byte form now, align if/when a canonical enum lands.

and that "if/when" is the whole point. the canonical-json work this waits on (@arnetheduck's consensus-specs#2983) was open from 2022 and closed unmerged in jan 2025. so (3) isn't "defer briefly", its json-only indefinitely, for a ~1 gb/state payload. the ssz form is opt-in via accept and purely additive, so if a canonical encoding ever supersedes this byte it's a versioning question for an opt-in content type, not a break to anything shipping today.

couple of concrete things: can validator_balances land now? it's just { index, balance }, no status, no enum, nothing your concern touches. and for validators, tell me the exact encoding you'd accept and i'll match it and align it with the builders endpoints (#610/#614) so we don't fragment this.

p.s: slop means low-quality, mass-produced content. this is a single sourced comment, and it's accurate, show me one wrong line if it isn't. saying a technically-sound pr would already be closed over how it's written rather than what it argues is shooting the messenger. let's stay on the encoding.

nflaig: I will let @lodekeeper handle the situation from here, so if you can convince him that either (1) or (2) are good options we can reconsider it

lodekeeper: Thanks @nflaig — happy to take the call.

**`validator_balances`: let's land it now.** It's `{index: uint64, balance: uint64}` — no status, no enum, nothing in this thread touches it. ~85 MB of JSON → a fraction of that in SSZ is a pure win for indexers. @Zyra-V21 please split it into its own PR; I don't see a credible objection to it landing independently.

**`validators`: I think we should ship SSZ, and resolve the `status` question by treating it as what it actually is — a derived field — and leaving it out of the SSZ body.**

`status` isn't independent data: it's a pure function of the `Validator` container (`activation_eligibility_epoch`, `activation_epoch`, `exit_epoch`, `withdrawable_epoch`, `slashed`), the queried state's epoch, and `balance` — all of which are already in an SSZ `{index, balance, validator}` response. So omitting it loses nothing; any consumer can derive the exact status the JSON endpoint returns. And the ~1 GB is dominated by the `Validator` containers, not the 1-byte status, so we still get essentially the full efficiency win.

This addresses @nflaig's core concern head-on: we put **no** bespoke/non-canonical enum encoding into the spec, so there's no future "canonical enum byte mapping differs → breaking v2" trap. It's also *not* option 2 as originally framed ("drop for now, add back when enum support lands") — that would be exactly the breaking-change trap. This is the opposite: `status` stays a JSON-layer convenience over the canonical fields, permanently, because it's recoverable. Nothing to add back.

On option 3 (JSON-only until canonical SSZ enums exist): I don't think we can hang a ~1 GB/state endpoint on that. consensus-specs#2983 was open from 2022 and closed unmerged in Jan 2025 — "defer until proper enum support" is realistically "never" right now. SSZ is opt-in via `Accept`, so JSON consumers are untouched regardless; there's no reason to withhold the SSZ form from the analytics/indexer use case that needs it.

If the room strongly prefers JSON/SSZ response symmetry (status present in both), the acceptable fallback is option 1 but with the spec **declaring** the #333 `uint8` mapping (stable + agreed since 2023, already implemented in Lodestar #6059) as the canonical encoding for this field — not provisional — which removes the "different future mapping" risk by definition. But my preference is the derived-field route; it keeps the spec enum-free and sidesteps the question entirely.

Concrete next steps:
1. Split `validator_balances` into its own PR and land it.
2. For `validators`, define the SSZ schema as `{index: uint64, balance: uint64, validator: Validator}` (no `status`) + a short spec note that `status` is derivable from `validator` + epoch + `balance` for SSZ consumers.

@etan-status — does the derived-field framing work for you? It keeps the spec enum-free while still unblocking SSZ for the endpoints that actually need it.

etan-status: I have a strong preference for consistency, as in, having validators and builders use the same strategy. And also for SSZ and JSON to represent the same information.

Regarding the SSZ format, I understand the desire given the size of the JSON response. It sounds a bit silly to even have a prolonged discussion here, given that the `uint8` mapping solves this in a pragmatic way, but given that this was acceptable by the community for years, I don't see why this is becoming some sort of emergency _now_. There are fewer builders than validators, there shouldn't be any urgency.

Where I disagree is that "defer until proper enum support" is realistically "never". This discussion about enums was surfaced only relatively recently and low-profile, and there is precedent of new SSZ types making it into the specs, with EIP-7916 and EIP-7495 currently being slated for Glamsterdam devnet 7. Deferring doesn't mean never. Specifically for enums, we would have to agree on:

- whether uint8 is the sole flavor (uint32 would allow representing ForkDigest contexts in SSZ, too)
- Holey enums / not starting at 0 / deprecation of individual cases later on
- JSON capitalization (web3signer uses uppercase, while others use Pascalcase, or lowercase, we mix it all!)
- interactions with CompatibleUnion SSZ type (can the different cases "carve out" different subsets of an enum / matching valid values have to share the same name, while extra values are allowed as long as they don't contradict across cases)
- support for Flags (C# style etc)
- other stuff that doesn't come to mind immediately...

SSZ primarily is a Merkleization format, not just a serialization format. So, adding a type on the SSZ level means that smart contracts who want to consume the data have to be considered. Even though in this specific usage, no Merkleization takes place, on the SSZ level, that's still necessary to do properly.

The discussion is too involved to do on some random comment thread for some micro-optimization, especially with Gloas still taking quite a bit of developer focus away. I'd suggest deferring SSZ for now on these endpoints, but not "forever" but just follow the proper process of discussing the questions with involved community members, making a proper EIP, implementing it in a couple SSZ libraries, do proper thorough testing etc, and only then revive the SSZ discussion for these endpoints.

Other aspects would be specific community members actually describing that they struggle with the JSON size rather than just assuming that they do (HTTP compression would probably help a ton already), or if there would be demand for enums from the core protocol itself, so that a niche RPC endpoint isn't the sole use of an entire SSZ data type...

Alternatively, we could also deprecate v1 JSON, and propose a v2 that replaces status with current_slot (and maybe finalized_slot if that's needed), and compute it on the client side. That could be done quickly, without a lengthy EIP process. At some UX cost for users who prefer `curl | jq` pipelines (but these users don't benefit from SSZ anyway).

lodekeeper: That all lands for me, and I think your v2 alternative is the cleanest path — let me converge on it.

**Conceding the "never":** fair correction. With EIP-7495/7916 slated for Glamsterdam devnet-7 the SSZ type system is clearly still moving, and an enum type has enough open design questions (uint8-only? holey / non-zero-start / per-case deprecation, JSON casing, CompatibleUnion + Flags interplay, and the Merkleization / smart-contract surface you noted) that it shouldn't be settled in this thread. I'll drop the "defer == never" framing.

**Consistency + your v2 idea:** agreed on both counts — SSZ and JSON should carry the same information, and validators/builders should use the same strategy. Your v2 (replace the derived `status` enum with the inputs needed to compute it, same shape in JSON and SSZ) does exactly that and sidesteps the enum question entirely instead of baking a provisional encoding into the spec. It's strictly better than my earlier "omit status from SSZ only" — it keeps the two formats symmetric, which was the wart in my version.

On the inputs: `current_slot` alone is sufficient to derive all nine statuses — they're a pure function of the `Validator` epoch fields + `slashed` + `balance` + current epoch, and `current_epoch = current_slot // SLOTS_PER_EPOCH`. `finalized_slot` isn't needed for status (none of the standard statuses depend on finalization), so v2 can be just `{index, balance, validator, current_slot}` unless there's a separate reason to carry finalized.

Concretely:
1. **`validator_balances`** has no `status` at all, so it doesn't need the v2 treatment — SSZ can be added to v1 directly. @Zyra-V21's offer to split it out is the easy immediate win; nothing in this discussion touches `{index, balance}`. Let's land that first.
2. **`validators` (and `builders` for consistency)**: pursue your v2 — drop the `status` enum, add `current_slot`, offer both JSON and SSZ. No enum in the spec, formats symmetric, and quick as you said (no EIP process).

That gets the ~1 GB endpoint into SSZ without rushing the enum design. Does that match what you had in mind?

etan-status: Yes, my take would be, either:

1. just let it be JSON-only, mirror validators design to builders endpoint for consistency --> no urgency
2. v2 that replaces per-validator status with top-level current_slot. lower QoL for curl | jq flow, but immediately unlocks SSZ for efficiency --> pragmatic compromise
3. proper enum design, following the EIP process --> slow, but the rolls-royce solution

roughly in this order of preference. note that v2 doesn't need hard fork coordination, one can phase it in gradually, dropping the need for a solution by G* release.

etan-status: for (2) it would be the requested state's slot instead of the wall clock slot, btw.

lodekeeper: Thanks @etan-status — that ordering works for me and gives a clear path forward.

Net:
- **`validator_balances`** is unblocked regardless of which option wins for `validators` — it's `{index, balance}`, no status, no enum. @Zyra-V21, splitting it into its own PR is the immediate SSZ win; nothing here holds it back.
- **`validators`**: the design question is settled — if/when we add SSZ here it's the v2 route (top-level `current_slot`, status derived client-side), no enum baked into the spec, and as you note it phases in gradually without hardfork coordination. Given the ~1 GB JSON I'd lean toward doing the v2 rather than parking it, but with the three options clearly laid out that's @Zyra-V21's call on appetite/timing.

So nothing's blocked on the enum discussion anymore. Thanks both for working it through.

nflaig: for reference, there was already an attempt to introduce a v2 without the `status` enum and there was no support for it, aside from what's on github in the PR here https://github.com/ethereum/beacon-APIs/pull/449, there were also concerns from different consumers, I would have to look up the discussions on discord etc., but the tldr is that there was no rough consensus to move forward with it, and imo if there is no consensus to do a specific api change, the default should always be to stay with the status quo as doing api changes introduces a lot of effort in clients, tooling etc.

lodekeeper: Thanks — that's the missing piece, and it changes my read. I looked at #449 (mcdee's "Add validators v2.", same ~600 MB motivation): open ~20 months, closed unmerged Jan 2026. So the v2-without-`status` shape isn't untried — it's been attempted and didn't reach rough consensus, with consumer concerns on top.

Given that, I agree the default should be status quo for `validators`. Re-introducing essentially the #449 design in this PR without new consensus cuts against exactly the principle you're describing — API changes carry real cost in clients/tooling, and "no consensus → keep status quo" is the right call. So I'll withdraw the lean toward doing the v2 here; if SSZ for `validators` is to happen it needs the consensus #449 lacked, which realistically means the proper enum process, not a re-run of v2 on this thread.

The one piece I'd still tease apart is `validator_balances` — it wasn't part of #449 and has no `status`/enum at all (`{index, balance}`), so the consumer concerns that sank the validators v2 don't apply. It's a purely additive, opt-in SSZ format for an ~85 MB payload. That feels like the part with a real shot at consensus, so worth @Zyra-V21 splitting into its own focused PR to test that — separate from the `validators` question, which stays status quo here.

nflaig: adding ssz to the `validator_balances` is a no brainer, @lodekeeper can you check lodestar, we should support this since ~2 years already, it makes sense to document that in the spec, it's also backward compatible to add ssz to an existing endpoint (due to content-type negotiation) so this seems overdue

lodekeeper: Confirmed — Lodestar already supports SSZ for `validator_balances`.

`GET /eth/v1/beacon/states/{state_id}/validator_balances` has no `onlySupport: json` restriction in its route definition, so it does content-type negotiation and returns SSZ on `Accept: application/octet-stream`. The encoding is `List[Container{index: uint64, balance: uint64}]` (`ValidatorBalanceListType`) — which matches the shape proposed here exactly, so documenting it just standardizes what we already serve.

It's been in since [#6749](https://github.com/ChainSafe/lodestar/pull/6749) ("provide first-class ssz support on api", merged 2024-06-10), shipped in **v1.20.0** — so ~2 years, as you remembered.

For contrast, `validators` (`getStateValidators`/`postStateValidators`) is explicitly `onlySupport: WireFormat.json` on our side — JSON-only — which lines up with leaving that endpoint at status quo per the earlier discussion.

So +1 to documenting the `validator_balances` SSZ format in the spec — it's backward-compatible (Accept-header opt-in) and already implemented.

Zyra-V21: I'll take a read at everything carefully and will be back by Monday! Thanks everyone for your collaboration, have a good weekend

Zyra-V21: thanks all, just went through all the comments above and im happy where this landed

i'll pull validator_balances out into its own pr. since @lodekeeper confirmed lodestar has been serving ssz on it since #6749 (v1.20.0) with the same {index, balance} shape, this is basically just writing down what's already shipping. opt in via the accept header so nothing changes for json clients.

fwiw on @etan-status's point about whether anyone actually struggles with the json size vs just assuming it: that's literally us (@migalabs) . this is the exact spot goteth hits it. we snapshot balances every epoch and backfill big historical ranges, and at full set size it's the json encode/parse on both ends that hurts, not really the wire bytes (gzip helps there, but not with the cpu cost). ssz makes wide backfills a lot cheaper.

and yeah, fine to leave validators at status quo for now. i didn't realise #449 had already tried the v2-without-status shape and couldn't get consensus, so no point re-running that here. if ssz for validators ever happens it should go through the proper enum process like @etan-status described, and stay consistent with builders. happy to help on that when it's a thing.

thanks @nflaig @lodekeeper @etan-status for working through it

lodekeeper: Sounds good 👍 The concrete goteth/migalabs angle (encode/parse CPU on epoch-by-epoch balance backfills, not just wire bytes) is a useful data point to have on record. And since Lodestar already serves the `{index, balance}` SSZ shape, the split PR should be pure spec documentation — nothing new to implement on our side. I'll keep an eye out for it and happy to review. Thanks all.

nflaig:
@Zyra-V21 can we close this one? I think there is not enough support to move forward with it

Zyra-V21:
> @Zyra-V21 can we close this one? I think there is not enough support to move forward with it

Sure! Thanks for your time @nflaig

PR #604, opened 2026-05-09
Add parent_beacon_block_root to ExecutionPayloadEnvelope

zilm13:
Spec adds new field `parent_beacon_block_root` to `ExecutionPayloadEnvelope` container, see https://github.com/ethereum/consensus-specs/pull/5152
Do we need it here too?

### DIFF
--- a/types/gloas/execution_payload_envelope.yaml
+++ b/types/gloas/execution_payload_envelope.yaml
@@ -1,8 +1,8 @@
 Gloas:
   ExecutionPayloadEnvelope:
     type: object
-    description: "The [`ExecutionPayloadEnvelope`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#executionpayloadenvelope) object from the CL Gloas spec."
-    required: [payload, execution_requests, builder_index, beacon_block_root, parent_beacon_block_root, slot, state_root]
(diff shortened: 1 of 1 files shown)


types/gloas/execution_payload_envelope.yaml
```diff
@@ -2,7 +2,7 @@ Gloas:
   ExecutionPayloadEnvelope:
     type: object
     description: "The [`ExecutionPayloadEnvelope`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#executionpayloadenvelope) object from the CL Gloas spec."
-    required: [payload, execution_requests, builder_index, beacon_block_root, slot, state_root]
+    required: [payload, execution_requests, builder_index, beacon_block_root, parent_beacon_block_root, slot, state_root]
```
nflaig: we need to update all containers for alpha.7 changes, the last update was alpha.5, adding parent root like this is strange while keeping the `state_root` field which doesn't exist anymore, for this PR, let's update the whole container to match alpha.7

zilm13: Sorry, I thought it was some extra only added in API, now i see it were the fields from alpha.2 spec. Updated to match alpha.7 spec

nflaig (commented):
lgtm, thanks @zilm13

nflaig (approved):

Issue #595, opened 2026-03-31
Move endpoint to get unsigned bid to builder namespace


nflaig:
The endpoint `GET /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}` is already part of the spec but instead of being part of `/validator` it would be better to move it to `/builder` as this endpoint will only be used by staked builders, not validators.

nflaig:
flagged by Potuz, we might also wanna consider allowing to pass `execution_payment` if builders wanna use this

PR #605, opened 2026-05-11
Add POST /eth/v1/beacon/execution_payload_envelope

pk910:
## Motivation

PR #552 standardized the bid POST and envelope-by-block GET but left the envelope publish path unspecified. Issue #575 raised an envelope **construction** endpoint for builders to sign against; it was closed without merging, and the **publish** path it implied was never spec'd.

Result: every Gloas CL that wants to support external builders today either ships its own path under its own conventions (Prysm exposes `POST /eth/v1/beacon/execution_payload_envelope` with a wrapped body; Lighthouse exposes the same path but takes only a bare envelope and ignores externally-supplied blobs; Lodestar accepts the same path but only for the self-build cached case) or does not expose any REST publish path at all (Grandine, Nimbus, Teku). External builder tooling - e.g. [`ethpandaops/buildoor`](https://github.com/ethpandaops/buildoor) - already targets the Prysm-style endpoint.

This PR specifies that endpoint so the surface converges before more clients ship divergent variants.

### New endpoint

`POST /eth/v1/beacon/execution_payload_envelope`

- Header: `Eth-Consensus-Version: gloas` (required).
- Body: `application/json` (`Gloas.SignedExecutionPayloadEnvelopeContents`)
  or `application/octet-stream` (SSZ).
- Responses: 200 / 202 / 400 / 415 / 500, matching the existing block
  publish and bid publish conventions.

On 20x the beacon node has gossip-validated the envelope and broadcast it on the `execution_payload` topic; it is also expected to integrate the envelope via `on_execution_payload_envelope` though it MAY broadcast before integration completes.

### New type

`Gloas.SignedExecutionPayloadEnvelopeContents` wraps:

- `signed_execution_payload_envelope: SignedExecutionPayloadEnvelope`
- `blobs: List[Blob, MAX_BLOB_COMMITMENTS_PER_BLOCK]`
- `kzg_proofs: List[KZGProof, MAX_BLOB_COMMITMENTS_PER_BLOCK]`

The wrapper name and field names mirror Prysm's existing `SignedExecutionPayloadEnvelopeContents`.

### Blob array semantics

The envelope itself, gossiped on the `execution_payload` topic, does not contain blobs - blobs ride on `DataColumnSidecar` gossip, which the publishing beacon node derives from these arrays. So this endpoint is also the only path the beacon node has to obtain blobs from an external builder.

- **External builder.** MUST supply the full `blobs` and `kzg_proofs` arrays. This is the primary use case.
- **Self-build over REST.** When the same beacon node built the payload, the caller MAY send empty `blobs` and `kzg_proofs` arrays; the beacon node MUST reconstruct from its local self-build cache and MUST reject with 400 if it cannot.

When non-empty, `blobs.length` and `kzg_proofs.length` MUST equal `ExecutionPayloadBid.blob_kzg_commitments.length`, the entries MUST be in the same order, and the beacon node MUST verify them against the committed commitments before broadcasting. `kzg_proofs` are cell proofs, matching `compute_cells_and_kzg_proofs` (the same shape used by `DataColumnSidecar.kzg_proofs`).

## Related

- #552 — standardized bid POST and envelope GET; this PR completes the publish side.
- #575 — closed; raised a builder envelope-construction endpoint. This PR does not re-introduce construction (the current envelope shape is fully derivable client-side from Engine API output plus the standard beacon-block lookup), but addresses the publish gap that issue left open.
- #595 — open; proposes moving the unsigned-bid GET to `/builder`. If that lands, this PR's endpoint can move alongside it as a follow-up.


bharath-123:
I believe its already been specified here: https://github.com/ethereum/beacon-APIs/pull/580/changes#diff-962ea34dc6ec0e65d4e1d6c8682e97c547e6e0d2f1168925210ff7ee47ef6ce0R109

PR #606, opened 2026-05-15
Extend /eth/v1/node/peers with peer scoring and disconnect reasons

barnabasbusa:
## Summary

Extends the existing `/eth/v1/node/peers` and `/eth/v1/node/peers/{peer_id}` responses with four optional fields exposing per-peer scoring and disconnect information:

- **`agent_version`** — the peer's libp2p identify agent string (e.g. `Lighthouse/v8.1.3-...`)
- **`score`** — the client-native peer score
- **`disconnect_reason`** — why the client last disconnected from the peer (controlled vocabulary)
- **`downscore_reasons`** — the distinct set of reasons the peer was downscored this session (controlled vocabulary)

Two new string enums, `PeerScoreReason` and `PeerDisconnectReason`, define the controlled vocabularies.

## Motivation

Peer scoring is one of the least observable parts of a running consensus node, yet it drives most peer churn. Every client already computes rich scoring internally but exposes it only through divergent, non-standard endpoints:

- Lighthouse — `GET /lighthouse/peers`
- Lodestar — `GET /eth/v1/lodestar/lodestar_peer_score_stats`
- Teku — `GET /teku/v1/nodes/peer_scores`
- Prysm — internal `ScoreInfo` proto (WIP REST endpoint)

This fragmentation means operators, monitoring tools, and testing frameworks (e.g. Kurtosis / interop debugging) can't ask "why is this node dropping peers?" in a client-agnostic way. Standardizing these fields lets tooling diagnose connectivity, fork/network mismatches, and rate-limiting uniformly across the network.

## Design notes

- **Strictly additive.** All four fields are optional, so existing consumers of `/eth/v1/node/peers` are unaffected.
- **Extends the existing schema rather than adding a new endpoint.** An earlier revision proposed a dedicated `/eth/v1/debug/node/peer_scores` endpoint; feedback moved it onto the existing `Peer` object to avoid a parallel peer-listing API.
- **Native score is intentionally not normalized.** Client score ranges differ widely ([-100,+100] for Lighthouse/Lodestar/Grandine, [-10,+20] for Teku, [0,1000] for Nimbus, [-100,+1] for Prysm), so `score` is only meaningful relative to other peers on the *same* client.
- **Controlled vocabularies with an `unknown` fallback.** The reason enums capture the realistic cross-client union rather than a full taxonomy; clients map finer-grained internal tags onto the closest listed value, and consumers are told to tolerate unknown values for forward compatibility.


types/p2p.yaml
```diff
@@ -57,6 +57,40 @@ Peer:
...
+      type: array
+      description: |
+        Score-affecting events observed for this peer within the
+        client's recent-history window (implementation-defined,
+        typically the last few minutes). OPTIONAL. The most recent
+        event SHOULD appear first. Clients MAY omit if no events
+        within the window or if event tracking is not implemented.
+      items:
+        $ref: "./p2p.yaml#/PeerScoreReason"
+      example: ["rpc_bad_blocks_by_range", "rpc_rate_limited"]
```
nflaig, line 86: I like the suggestion from @rolfyone to add a `extra_data` field similar to fork choice v1 api

types/p2p.yaml
```diff
@@ -57,6 +57,40 @@ Peer:
       $ref: "./p2p.yaml#/PeerConnectionState"
     direction:
       $ref: "./p2p.yaml#/PeerConnectionDirection"
+    agent_version:
```
nflaig, line 60: we might wanna do a v2 instead of adding new fields, but curious what others think

rolfyone: im happy with new fields as long as they're not mandatory.

types/p2p.yaml
rolfyone: did we want to add this before merging? or add if needed?

barnabasbusa: so you want to add also extra data or replace a field with extra data?

types/p2p.yaml
barnabasbusa: yeah none of this should be mandatory. 
if we ever do a v2 we should have these mandatory.

types/p2p.yaml
rolfyone: I'd potentially skip for now, but will let @nflaig express their preference, i dont mind either way really.

rolfyone:
I'm not sure how common it is, but we only appear to have 'large_penalty' and 'small_penalty' for downscore reasons, so an initial implementation from teku may not have that - are the other fields still useful if we can't easily provider downscore reason or disconnect reason @barnabasbusa ?
sample:
```
    {
      "peer_id": "...",
      "last_seen_p2p_address": "/ip4/0.0.0.0/udp/4477/quic-v1",
      "state": "connected",
      "direction": "inbound",
      "agent_version": "Prysm/v7.1.2/7950a249266a692551e5a910adb9a82a02c92040",
      "score": -483.74838746020384
    }
```

CHANGES.md
```diff
@@ -8,7 +8,8 @@ There are likely to be descriptions etc outside of the list below, but new query
...
+| `GET /eth/v1/node/peers` and `GET /eth/v1/node/peers/{peer_id}` extended with optional `agent_version`, `score`, `disconnect_reason`, `downscore_reasons` fields         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#552](https://github.com/ethereum/beacon-APIs/pull/552) `GET /eth/v1/validator/execution_payload_bid/{slot}/{builder_index}` added         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#552](https://github.com/ethereum/beacon-APIs/pull/552) `GET /eth/v1/validator/payload_attestation_data/{slot}` added         |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig, line 14: changelog seems to be messed up, I also don't think we need a changelog entry for the changes in this PR

types/p2p.yaml
```diff
@@ -80,3 +114,46 @@ Multiaddr:
...
+    - rpc_bad_blocks_by_range
+    - rpc_bad_blocks_by_root
+    - gossip_invalid_block
+    - gossip_invalid_attestation
+    - gossip_invalid_blob_sidecar
+    - gossip_invalid_data_column_sidecar
+    - sync_bad_batch
+    - status_unviable_fork
+    - behaviour_penalty
+    - unknown
```
nflaig: any reason why these are `enum` instead of just `example`? it seems to me that we wanna keep this list open to extend and not restrict to these values, also not all clients might return all the examples from here

also, why is there no `gossip_invalid_payload_envelope` for example? and many others are missing, so if it's not an exhaustive list might as well limit it to just a few examples as guidance for implementers

types/p2p.yaml
```diff
@@ -80,3 +114,46 @@ Multiaddr:
...
+    forward compatibility.
+  enum:
+    - bad_score
+    - irrelevant_network
+    - unviable_fork
+    - too_many_peers
+    - rate_limited
+    - io_error
+    - client_shutdown
+    - unknown
```
nflaig, line 149: same as above, consider using `example`

types/p2p.yaml
nflaig: for this specific api it's fine to add new optional fields, it's also more of a debug api, but mostly it's fine because we don't or rather can't support ssz here because of the format/return values, I just wanna make it clear that this is not a general pattern we wanna have, or encourage

types/p2p.yaml
nflaig: since we keep v1 and add new optional fields, there is no point in adding `extra_data`, but for a v2 with a more strict schema for top-level fields (ie. only required fields) this could be useful as you can add experimental (relevant for new forks) or client specific fields into `extra_data`. But for now we can skip it

types/p2p.yaml
```diff
@@ -57,6 +57,40 @@ Peer:
...
+      example: "Lighthouse/v8.1.3-66919c2/aarch64-linux"
+    score:
+      type: number
+      format: double
+      description: |
+        Client-native peer score. OPTIONAL. The scale and meaning is
+        implementation-defined - consumers SHOULD treat it as a relative
+        signal within a single client, not directly comparable across
+        clients. Lower values indicate worse standing. Clients that do
+        not maintain a per-peer score MAY omit this field.
```
nflaig: are we fine with having all these vibe coded descriptions? I definitely don't have time to polish this but if someone can it would be great

nflaig:
image

@barnabasbusa can you add some rationale to the PR body/description why we want/need this change for posterity this would be useful to have/persist here

types/p2p.yaml
rolfyone: they wanted a limited list i think so its semi-standard across all clients...

nflaig: we can make it a limited list sure, but then someone has to take time and make it useful and relevant

there are 0 gloas related gossip errors besides the data column one, nothing related to bid, payload envelopes or ptc attestations

also why is there `gossip_invalid_blob_sidecar`? that doesn't make any sense, if we want a limited list, then it's needs to be well-defined and useful

types/p2p.yaml
```diff
@@ -57,6 +57,32 @@ Peer:
...
+      example: -17.4828
+    disconnect_reason:
+      description: |
+        Reason the client last disconnected from this peer. OPTIONAL.
+        This field MUST only be set if the connection status is disconnected or disconnecting.
+      $ref: "./p2p.yaml#/PeerDisconnectReason"
+    downscore_reasons:
+      type: array
+      description: |
+        Reasons that the client has been down scored in their current session. OPTIONAL.
```
dknopik, line 82: Is this supposed to be added per occurrence, or once if it occurred at least once?

E.g., if `gossip_invalid_block` happened twice, should we return
```json
["gossip_invalid_block", "gossip_invalid_block"]
```
or 
```json
["gossip_invalid_block"]
```

I assume the latter, but it might be worth clarifying this in the description.

rolfyone: i would expect the latter. The description was a mess so i reduced it substantially because it was ai word slop... can potentially describe it as a distinct set...

PR #607, opened 2026-05-18
Reflect recent update to gloas specs

ensi321:
Reflect the latest consensus-specs `v1.7.0-alpha.8`.

- Add `execution_requests_root` to `Gloas.ExecutionPayloadBid`.
- Add `parent_execution_requests` to `Gloas.BeaconBlockBody`.
- Add `BlockAccessList` (EIP-7928) and `Gloas.ExecutionPayload` extending Deneb with `block_access_list` and `slot_number` (EIP-7843).
- Point `ExecutionPayloadEnvelope.payload` to `Gloas.ExecutionPayload`.
- Register `Gloas.ExecutionPayload` in `beacon-node-oapi.yaml`.
- Fix `ExecutionPayloadBid` field widths: `parent_block_hash` and `block_hash` to `Hash32`, `prev_randao` to `Bytes32`. Add `Hash32` to `types/primitive.yaml`.


types/primitive.yaml
```diff
@@ -66,6 +66,12 @@ Root:
   example: "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
   pattern: "^0x[a-fA-F0-9]{64}$"
 
+Hash32:
```
ensi321, line 69: I am not sure if we want this, or we can just use `Byte32`. I think it is nice to have a separate `Hash32` type for EL related fields in CL.

But I also see `block_hash` in `ExecutionPayloadCommon` being `Root`.

rolfyone: its differentiated in the spec, so it might make it cleaner to have hash32 defined here too. fair call imo.

rolfyone (dismissed):
LGTM

apis/validator/execution_payload_bid.yaml
```diff
@@ -36,10 +36,10 @@ get:
                 enum: [gloas]
                 example: "gloas"
               data:
-                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.ExecutionPayloadBid"
+                $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedExecutionPayloadBid"
```
nflaig: I don't understand this change, this is not a pool api, it is meant to be used by builders to retrieve a bid to sign over

https://github.com/ethereum/beacon-APIs/blob/6934ec6b63c1e89286f99446689796b9302594f2/validator-flow.md?plain=1#L79-L80

side note: we should also move this to builder namespace https://github.com/ethereum/beacon-APIs/issues/595

types/gloas/block.yaml
```diff
@@ -1,8 +1,8 @@
 Gloas:
   BeaconBlockBody:
     type: object
-    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#beaconblockbody) object from the CL Gloas spec."
-    required: [randao_reveal, eth1_data, graffiti, proposer_slashings, attester_slashings, attestations, deposits, voluntary_exits, sync_aggregate, bls_to_execution_changes, signed_execution_payload_bid, payload_attestations]
+    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.9/specs/gloas/beacon-chain.md#beaconblockbody) object from the CL Gloas spec."
```
nflaig: alpha.9 does not exist

types/gloas/block_access_list.yaml
```diff
@@ -0,0 +1,6 @@
+Gloas:
+  BlockAccessList:
```
nflaig: this is just a new field in the execution payload, I'd prefer not to have this as a type like this, rather inline it in `ExecutionPayload`

beacon-node-oapi.yaml
```diff
@@ -449,6 +449,10 @@ components:
       $ref: "./types/gloas/execution_payload_bid.yaml#/Gloas/ExecutionPayloadBid"
     Gloas.SignedExecutionPayloadBid:
       $ref: "./types/gloas/execution_payload_bid.yaml#/Gloas/SignedExecutionPayloadBid"
+    Gloas.ExecutionPayload:
+      $ref: "./types/gloas/execution_payload.yaml#/Gloas/ExecutionPayload"
+    Gloas.BlockAccessList:
+      $ref: "./types/gloas/block_access_list.yaml#/Gloas/BlockAccessList"
```
nflaig: definitely against exposing `BlockAccessList` here

types/primitive.yaml
nflaig: I am fine with adding `Hash32` but should make sure it's used consistently across the board, and not just gloas types, a separate PR for this might be better

apis/validator/execution_payload_bid.yaml
ensi321: > side note: we should also move this to builder namespace https://github.com/ethereum/beacon-APIs/issues/595

Probably a separate PR for it

types/gloas/execution_payload.yaml
```diff
@@ -0,0 +1,15 @@
...
+    description: "The [`ExecutionPayload`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.8/specs/gloas/beacon-chain.md#executionpayload) object from the CL Gloas spec."
+    allOf:
+      - $ref: "../deneb/execution_payload.yaml#/Deneb/ExecutionPayload"
+      - type: object
+        required: [block_access_list, slot_number]
+        properties:
+          block_access_list:
+            type: string
+            format: hex
+            description: "RLP encoded `BlockAccessList` as defined in [EIP-7928]"
```
nflaig: is this supposed to have a link to the EIP?
```suggestion
            description: "RLP encoded `BlockAccessList` as defined in [EIP-7928](https://eips.ethereum.org/EIPS/eip-7928)"
```

nflaig (commented):
the PR description needs to be adjusted

types/gloas/block.yaml
```diff
@@ -1,8 +1,8 @@
 Gloas:
   BeaconBlockBody:
     type: object
-    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.2/specs/gloas/beacon-chain.md#beaconblockbody) object from the CL Gloas spec."
-    required: [randao_reveal, eth1_data, graffiti, proposer_slashings, attester_slashings, attestations, deposits, voluntary_exits, sync_aggregate, bls_to_execution_changes, signed_execution_payload_bid, payload_attestations]
+    description: "The [`BeaconBlockBody`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.8/specs/gloas/beacon-chain.md#beaconblockbody) object from the CL Gloas spec."
```
nflaig, line 4: I think it would be good if we can bump all spec references to `v1.7.0-alpha.8` that currently reference an older alpha.x release

types/gloas/execution_payload.yaml
```diff
@@ -0,0 +1,15 @@
...
+            pattern: "^0x[a-fA-F0-9]{0,2147483648}$"
```
nflaig: might be good to also add an example above `pattern`

eg. `example: "0xe5e494abcf8e0d4e9587369b2301d0790347320302cc09c0c0cac90187b1a2bc2ec50000c0c0"`

types/primitive.yaml
nflaig: it's not too many changes so just applied it here in https://github.com/ethereum/beacon-APIs/pull/607/changes/c5947801957436c68ecde502f453ef956013c55f

types/gloas/block_access_list.yaml
nflaig: I realized it's more consistent with the current spec to have this in `primitives.yaml` instead so I moved it there in https://github.com/ethereum/beacon-APIs/pull/607/changes/7f342436e0907af12c1a05fded14fbdbc1cd3c4d

nflaig (approved):
LGTM

PR #593, opened 2026-03-24
add proposer preferences related api specifications

bharath-123:
This PR partially resolves https://github.com/ethereum/beacon-APIs/issues/570

This PR adds the following:
1. Adds the proposer preferences type in the beacon api specs
2. Adds the proposer preferences SSE event api spec
3. Adds the proposer preferences pool api to get pooled preferences and to allow users to submit preferences.

PR #608, opened 2026-05-19
gloas proposer preference apis

### LINKED ISSUE #570: Add spec for proposer preferences
This is currently low prio since https://github.com/ethereum/consensus-specs/pull/4777 won't be part of devnet-0.

We probably want an endpoint to retrieve proposer preferences from op pool and a new event `proposer_preferences` which can be subscribed to by builders.

We also need a new api for validator clients to submit `SignedProposerPreferences` to the beacon node to be broadcast.

james-prysm:
Supersedes #593 with https://github.com/ethereum/consensus-specs/pull/5196, https://github.com/ethereum/consensus-specs/pull/5236 added

Adds the proposer preferences API for Gloas, deprecates the legacy
  fee-recipient/gas-limit endpoints, and tightens `beacon_committee_subscriptions` for
  CGC bookkeeping.

  ## Added

  - `POST /eth/v1/validator/proposer_preferences` — VC submission of signed
    preferences and publishes them on the `proposer_preferences` gossipsub topic.
  - `proposer_preferences` SSE event for messages passing gossip validation.
  - `Gloas.ProposerPreferences` / `Gloas.SignedProposerPreferences` types, including
    `dependent_root` (used for gossip dedup and validation).

  ## Deprecated

  - `POST /eth/v1/validator/prepare_beacon_proposer` — superseded by signed proposer
    preferences. Pre-Gloas support is REQUIRED; post-Gloas the endpoint MAY be a no-op
    or removed.
  - `POST /eth/v1/validator/register_validator` — fee recipient and gas limit move to
    signed proposer preferences. Same pre/post-Gloas support rules apply.

  ## Updated

  - `POST /eth/v1/validator/beacon_committee_subscriptions` — from Gloas onwards, BNs
    use the subscription set to identify which validators are using the node and to
    size the node's custody group count (CGC). VCs SHOULD submit one entry per
    attached active validator per attestation duty, including when multiple
    validators share the same `(slot, committee_index)`, so the BN tracks every
    attached validator rather than only one per subnet.
    
    

Fixes #570


apis/beacon/pool/proposer_preferences.yaml
```diff
@@ -0,0 +1,47 @@
...
+      content:
+        application/json:
+          schema:
+            title: GetPoolProposerPreferencesResponse
+            type: object
+            required: [version, data]
+            properties:
+              version:
+                type: string
+                enum: [gloas]
```
james-prysm: maybe fulu should be available too?

bharath-123:
Awesome! I ll close my PR!

apis/beacon/pool/proposer_preferences.yaml
rolfyone: is there any reason not to include fulu?

apis/validator/beacon_committee_subscriptions.yaml
```diff
@@ -8,6 +8,13 @@ post:
     If validator `is_aggregator`, beacon node must:
     - announce subnet topic subscription on gossipsub
     - aggregate attestations received on that subnet
+
+    From the Gloas fork onwards, beacon nodes use the set of validators that have submitted
+    subscriptions via this endpoint to identify which validators are using the node for duties,
+    and adjust the node's custody group count (CGC) accordingly. Validator clients SHOULD
```
rolfyone: historically CGC was up only, this implies it might be able to be reduced now?

james-prysm: No I can try to reword it. I'm just implying that we should use this as a replacement to prepare beacon proposer endpoint for cgc purposes.

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,56 @@
...
+      limit information supplied via `POST /eth/v1/validator/prepare_beacon_proposer` and
+      `POST /eth/v1/validator/register_validator`. From the Gloas fork onwards, signed proposer
+      preferences supersede both endpoints as the source of proposer fee recipient and gas
+      limit preferences.
+    - For validators served by the beacon node, if no signed proposer preferences are available
+      for a given proposal slot, the beacon node MAY fall back to its configured default fee
+      recipient and target gas limit when constructing or accepting bids for that validator.
+      Bids whose fee recipient or gas limit do not match a validator's preferences will be
+      rejected at gossip validation, so applying sensible local defaults reduces missed
+      proposals during the transition.
```
rolfyone: should (vs may) fall back? what would be the alternative?

wordlist.txt
```diff
@@ -45,3 +45,5 @@ ptc
 PtcDuty
 produceBlockV
 stateful
+CGC
+dedup
```
rolfyone: dedup is a horrible horrible word, can we just spell de-duplicate out instead?

apis/beacon/pool/proposer_preferences.yaml
```diff
@@ -0,0 +1,47 @@
+get:
+  operationId: getPoolProposerPreferences
```
michaelsproul: A bit pedantic of me, but I find it strange that there is a pool API for a message type that is not included in blocks. In my mind the `/pool` APIs are for messages that can be packed into blocks like attestations, payload attestations, sync committee messages, etc.

apis/beacon/pool/proposer_preferences.yaml
michaelsproul: the proposer preferences concept doesn't exist or make sense for Fulu, does it?

apis/beacon/pool/proposer_preferences.yaml
michaelsproul: Separately, why do beacon nodes have to store these messages at all? My understanding is that they exist mainly for the sake of builders, so there is limited value in a beacon node keeping track of them after validation (aside from the minimal amount of data that is required to be tracked to detect duplicates).

michaelsproul: I guess builders mostly use the SSE event, but also load from this pool to catch up on any events that happened in the recent past prior to their subscription starting?

bharath-123: Yeah the SSE event is the most important one. I guess its not so important to catch up on any events cause the preferences are gossiped every epoch. So in the worse case a builder would have to wait 2 epochs before building their blocks which I think is fine?

james-prysm: should i remove it from the pool namespace? i can do that. i just didn't think it should be under validator, the original proposal from bharath was under pool.

waiting for some more opinions here I can remove the endpoint in general if not very useful

apis/validator/proposer_preferences.yaml
james-prysm: zeroed out i guess, in a sense that is also a local default so i'll change it to should

apis/beacon/pool/proposer_preferences.yaml
nflaig: There is no fulu type for `SignedProposerPreferences`, so I'd just keep gloas here

apis/beacon/pool/proposer_preferences.yaml
```diff
@@ -0,0 +1,47 @@
...
+                type: string
+                enum: [gloas]
+                example: "gloas"
+              data:
+                type: array
+                items:
+                  $ref: "../../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedProposerPreferences"
+        application/octet-stream:
+          schema:
+            description: "SSZ serialized `List[SignedProposerPreferences]` bytes. Use Accept header to choose this response type"
```
nflaig: `List` requires a limit, need to think more about what the right limit here is, technically you can have proposer preferences from different branches, so `2 * SLOTS_PER_EPOCH` might not be sufficient

apis/beacon/pool/proposer_preferences.yaml
nflaig: > Separately, why do beacon nodes have to store these messages at all?

we do need to store them to be able to validate bids against target gas limit and fee recipient of the proposer

> A bit pedantic of me, but I find it strange that there is a pool API for a message type that is not included in blocks.

in lodestar we currently call this `ProposerPreferencesPool`, but we can think about renaming this if we wanna keep that notion to only use pool for messages that are included in blocks

apis/validator/prepare_beacon_proposer.yaml
```diff
@@ -1,7 +1,13 @@
 post:
   operationId: "prepareBeaconProposer"
   summary: Provide beacon node with proposals for the given validators.
+  deprecated: true
   description: |
+    Deprecated from Gloas onwards in favor of
+    `POST /eth/v1/validator/proposer_preferences`, which supersedes both fee recipient and
+    gas limit preferences. Beacon node support for this endpoint pre-Gloas is REQUIRED;
+    post-Gloas support is OPTIONAL and the endpoint MAY be removed.
```
nflaig, line 9: I don't think we should keep supporting this endpoint going forward, ideally, we wanna get rid of it in hegota

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,56 @@
+post:
+  operationId: submitSignedProposerPreferences
```
nflaig: do we want the "signed" part in the operation id?

nflaig (commented):
just some initial comments, haven't done an in-depth review yet

apis/validator/prepare_beacon_proposer.yaml
james-prysm: should i reword this or is this wording fine

apis/validator/proposer_preferences.yaml
james-prysm: good question will double check other ids

james-prysm: they don't good catch i will remove it

apis/beacon/pool/proposer_preferences.yaml
james-prysm: yeah was just thinking from perspective of having to submit proposer preferences while still in fulu and having the ability to get what you submitted?

nflaig: while the clock slot is still fulu, the `proposal_slot` would be in gloas, we would use the message slot anyways to determine the fork? I guess the same question exists in gossip, which uses the gloas fork digest even if messages are broadcast before the fork (at least in lodestar)

the same question exists if we wanna extend the proposer preferences in the future but I think we should always use the message slot

apis/beacon/pool/proposer_preferences.yaml
james-prysm: for now i removed it from the pools namespace as suggested

tersec: I'm not sure I see the utility of it either.

james-prysm:
updated wording to fall back to parent gas limit

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
...
+      when constructing or accepting bids for that validator. Bids whose fee recipient or gas
+      limit do not match a validator's preferences will be rejected at gossip validation, so
+      applying sensible defaults reduces missed proposals during the transition.
+  parameters:
+    - in: header
+      schema:
+        $ref: "../../beacon-node-oapi.yaml#/components/schemas/ConsensusVersion"
+      required: true
+      name: Eth-Consensus-Version
+      description: "The active consensus version to which the signed proposer preferences being submitted belongs."
```
syjn99, line 30: Do we need to enforce the consensus version here?

nflaig: that's required to make the endpoint forward compatible, otherwise if `SignedProposerPreferences` format changes in the future (eg. in heze) you cannot signal to the receiver what's the data format you are sending (required by receiver to decode the data), the receiver could always compute it from the current clock slot in the case here, but since proposer preferences should be send 1 epoch before the fork (due to epoch lookahead), that's not a good option either.

apis/beacon/pool/proposer_preferences.yaml
james-prysm: I removed it

apis/validator/beacon_committee_subscriptions.yaml
```diff
@@ -8,6 +8,13 @@ post:
     If validator `is_aggregator`, beacon node must:
     - announce subnet topic subscription on gossipsub
     - aggregate attestations received on that subnet
+
+    From the Gloas fork onwards, beacon nodes use the set of validators that have submitted
```
nflaig: we might wanna omit this, generally, this is not gloas specific, we should make sure even pre-gloas to transition to use `prepareBeaconCommitteeSubnet` to determine the custody requirments

```suggestion
    Beacon nodes use the set of validators that have submitted
```

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - Validator clients SHOULD begin submitting signed proposer preferences at least one epoch
+      before the Gloas fork activates, so the gossip caches are warm at the fork boundary.
+    - Until the Gloas fork activates, beacon nodes MUST continue to honor fee recipient and gas
+      limit information supplied via `POST /eth/v1/validator/prepare_beacon_proposer` and
+      `POST /eth/v1/validator/register_validator`. From the Gloas fork onwards, signed proposer
+      preferences supersede both endpoints as the source of proposer fee recipient and gas
+      limit preferences.
+    - For validators served by the beacon node, if no signed proposer preferences are available
+      for a given proposal slot, the beacon node should fall back to the parent block's gas limit
+      when constructing or accepting bids for that validator. Bids whose fee recipient or gas
```
nflaig: I find the "constructing bids" term here a bit confusing, you mean when preparing the local payload?

nflaig: worth pointing out if there are no proposer preferences for a given slot then the bids for this slot would be ignored, at least from p2p. so does this mean it should still accept bids via api though?

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
+post:
+  operationId: submitProposerPreferences
+  summary: Submit signed proposer preferences
+  description: |
+    Verifies given signed proposer preferences and publishes them on the `proposer_preferences`
+    gossipsub topic.
+
+    Notes:
+    - Validator clients SHOULD begin submitting signed proposer preferences at least one epoch
```
nflaig: does this render correctly in the api explorer? my concern is that the bullet points are not actually rendered below each other, I believe there needs to be a new line between each, but would need to render locally to confirm

CHANGES.md
```diff
@@ -29,6 +29,11 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#590](https://github.com/ethereum/beacon-APIs/pull/590) `head EVENT` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#598](https://github.com/ethereum/beacon-APIs/pull/598) `fast_confirmation EVENT` added                            |                                                  |                                                   |                                                    |                                                 |                                           |
+| [#593](https://github.com/ethereum/beacon-APIs/pull/593) `POST /eth/v1/validator/proposer_preferences` added         |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig: wrong PR number

types/gloas/proposer_preferences.yaml
```diff
@@ -0,0 +1,27 @@
+Gloas:
+  ProposerPreferences:
+    type: object
+    description: "The [`ProposerPreferences`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#proposerpreferences) object from the CL Gloas spec."
+    required: [dependent_root, proposal_slot, validator_index, fee_recipient, target_gas_limit]
+    properties:
+      dependent_root:
+        description: "Block root the message is anchored to. Used by gossip validation to de-duplicate and to anchor signature verification to the checkpoint state at `epoch(proposal_slot) - 1`."
```
nflaig: `epoch(proposal_slot) - 1` what is this referencing to? we might wanna align this more with how it's done on the CL spec side

types/gloas/proposer_preferences.yaml
```diff
@@ -0,0 +1,27 @@
...
+      validator_index:
+        $ref: "../primitive.yaml#/Uint64"
+      fee_recipient:
+        $ref: "../primitive.yaml#/ExecutionAddress"
+      target_gas_limit:
+        $ref: "../primitive.yaml#/Uint64"
+
+  SignedProposerPreferences:
+    type: object
+    description: "The [`SignedProposerPreferences`](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#signedproposerpreferences) object from the CL Gloas spec."
```
nflaig: can you please double check the links here, this one doesn't work (I haven't tested others)

but here it should be `https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/gloas/p2p-interface.md#new-signedproposerpreferences`


(also note: we should use a tag or commit hash, not `master`, my preference is to use tags)

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+            $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.SignedProposerPreferences"
+      application/octet-stream:
+        schema:
+          description: "SSZ serialized `List[SignedProposerPreferences]` bytes. Use Content-Type header to indicate that SSZ data is contained in the request body."
```
nflaig: I think I mentioned this before, we need a list limit here. But that raises the question, how many do we allow submitting at once here? this should just match the proposer lookahead I would think, so at max you can submit current + next epoch proposer preferences, so my suggestion would be to use `(MIN_SEED_LOOKAHEAD + 1) * SLOTS_PER_EPOCH` as limit.

CHANGES.md
```diff
@@ -29,6 +29,11 @@ There are likely to be descriptions etc outside of the list below, but new query
...
+| [#593](https://github.com/ethereum/beacon-APIs/pull/593) `POST /eth/v1/validator/prepare_beacon_proposer` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#593](https://github.com/ethereum/beacon-APIs/pull/593) `POST /eth/v1/validator/register_validator` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#593](https://github.com/ethereum/beacon-APIs/pull/593) `POST /eth/v1/validator/beacon_committee_subscriptions` updated         |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig: I don't think this should be listed as updated

CHANGES.md
james-prysm: Crap will do this tomorrow and get through the items before acdt

nflaig: I personally don't think we need this merged before ACDT so no rush from my side lol, it strange to me that people want this merged, nobody besides us (lodestar) is running api spec tests anyways, it doesn't hurt to keep PRs open to allow for more people to chime in/give feedback, once a PR is merged it's much less likely and also harder to give feedback... but what do I know, seems like ACD wants this merged by monday :man_shrugging:

CHANGES.md
nflaig: hmm maybe we can keep, might be worth pointing out

apis/validator/prepare_beacon_proposer.yaml
nflaig: might be fine, no strong opinion tbh

apis/validator/register_validator.yaml
```diff
@@ -1,7 +1,14 @@
 post:
   operationId: "registerValidator"
   summary: Provide beacon node with registrations for the given validators to the external builder network.
+  deprecated: true
   description: |
+    Deprecated from Gloas onwards. Under Gloas, the external builder relay flow is replaced
+    by the in-protocol builder market; proposer fee recipient and gas limit preferences are
```
nflaig: > the external builder relay flow is replaced by the in-protocol builder market

unfortunately nothing will change here, we will also still need another endpoint to submit builder preferences to the beacon node similar to `registerValidator` and forward them to the builder(s) connected via api, although there is no need to submit for all validators each epoch, but rather just the ones that have an upcoming proposal

apis/eventstream/index.yaml
```diff
@@ -194,6 +195,11 @@ get:
               value: |
                 event: fast_confirmation
                 data: {"block": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "slot": "1", "current_slot": "2"}
+            proposer_preferences:
+              description: The node has received a `SignedProposerPreferences` (from P2P or API) that passes gossip validation on the `proposer_preferences` topic
+              value: |
+                event: proposer_preferences
+                data: {"version":"gloas", "data":{"message": {"dependent_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "proposal_slot": "32", "validator_index": "123", "fee_recipient": "0x0000000000000000000000000000000000000000", "target_gas_limit": "30000000"}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
```
nflaig: nit

```suggestion
                data: {"version":"gloas", "data":{"message": {"dependent_root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2", "proposal_slot": "32", "validator_index": "123", "fee_recipient": "0x0000000000000000000000000000000000000000", "target_gas_limit": "60000000"}, "signature": "0x1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505cc411d61252fb6cb3fa0017b679f8bb2305b26a285fa2737f175668d0dff91cc1b66ac1fb663c9bc59509846d6ec05345bd908eda73e670af888da41af171505"}}
```

apis/validator/proposer_preferences.yaml
nflaig: > at least one epoch

the "at least" might be confusing, it's not possible with the current `MIN_SEED_LOOKAHEAD = 1` to submit them earlier than one epoch

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
...
+    Notes:
+    - Validator clients SHOULD begin submitting signed proposer preferences at least one epoch
+      before the Gloas fork activates, so the gossip caches are warm at the fork boundary.
+    - Until the Gloas fork activates, beacon nodes MUST continue to honor fee recipient and gas
+      limit information supplied via `POST /eth/v1/validator/prepare_beacon_proposer` and
+      `POST /eth/v1/validator/register_validator`. From the Gloas fork onwards, signed proposer
+      preferences supersede both endpoints as the source of proposer fee recipient and gas
+      limit preferences.
+    - For validators served by the beacon node, if no signed proposer preferences are available
+      for a given proposal slot, the beacon node should fall back to the parent block's gas limit
```
nflaig:
```suggestion
      for a given proposal slot, the beacon node SHOULD fall back to the parent block's gas limit
```

types/gloas/proposer_preferences.yaml
nflaig: maybe we can just keep it simple here, details can be found in the CL spec [here](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/validator.md#broadcasting-signedproposerpreferences)

apis/validator/beacon_committee_subscriptions.yaml
nflaig: addressed by https://github.com/ethereum/beacon-APIs/pull/608/changes/add85aeb577fbd9427228c3078e3c67c8d7f2bce

CHANGES.md
nflaig: updated in cb6d57fc0efb9d6260c23022ec81c44d21b7c652

types/gloas/proposer_preferences.yaml
nflaig: fixed in 217c65c52cff4e736a22cd37c017b9c25971486d

types/gloas/proposer_preferences.yaml
nflaig: updated in https://github.com/ethereum/beacon-APIs/pull/608/changes/8000110ae1c8a7ff5a442bd47c12ddfdb6c1d339

apis/validator/proposer_preferences.yaml
nflaig: updated in 20108408495ceb1efd829a3af148a9227e34e8c3

CHANGES.md
nflaig: I removed this entry in f267c92a0c5a981157d5b55a787edf439cf0193c

apis/validator/proposer_preferences.yaml
nflaig: done in a30e2462582d60e448b4f740e96f89bd82fcc3fc

apis/validator/proposer_preferences.yaml
```diff
@@ -0,0 +1,55 @@
+post:
+  operationId: submitProposerPreferences
+  summary: Submit signed proposer preferences
+  description: |
+    Verifies given signed proposer preferences and publishes them on the `proposer_preferences`
+    gossipsub topic.
+
+    Notes:
+    - Validator clients SHOULD begin submitting signed proposer preferences at least one epoch
+      before the Gloas fork activates, so the gossip caches are warm at the fork boundary.
```
nflaig: > gossip caches are warm at the fork boundary

what do you mean by that?

apis/validator/proposer_preferences.yaml
nflaig: it looks correct

image

apis/validator/proposer_preferences.yaml
james-prysm: I should probably rename this, just that the caches for preferences are usable at the fork with the required preferences for the proposers submitting at that time

james-prysm: https://github.com/ethereum/beacon-APIs/pull/608/commits/ed61b2b5ec7db1b73df5f59bcf60434615a94bed fixed here

apis/validator/proposer_preferences.yaml
james-prysm: updated https://github.com/ethereum/beacon-APIs/pull/608/commits/6cb794bdbb3aca2ff72cfb55eab14916542ac153

apis/validator/proposer_preferences.yaml
james-prysm: https://github.com/ethereum/beacon-APIs/pull/608/commits/6cb794bdbb3aca2ff72cfb55eab14916542ac153 I updated here.

james-prysm: hmm i didn't think of this edge case, I thought it should fall back to whatever default you provided but if p2p ignores it already then it doesn't matter?

james-prysm: how about this? https://github.com/ethereum/beacon-APIs/pull/608/commits/8b3235eb06ac00f0f1633e589571808b972a99e6

apis/validator/proposer_preferences.yaml
nflaig: > [ed61b2b](https://github.com/ethereum/beacon-APIs/commit/ed61b2b5ec7db1b73df5f59bcf60434615a94bed) fixed here

this is much better, also was thinking if we might wanna highlight that builders can prepare bids sooner right at the fork boundary, but either way, the generic phrasing we have now might already imply that since they also will have their own proposer preferences cache, similar to what we have in the beacon-node

apis/validator/register_validator.yaml
nflaig: @james-prysm I removed this as I find it a bit confusing, see https://github.com/ethereum/beacon-APIs/pull/608/changes/300692f3642d039404b60827732aa6a783e539d9, are you ok with this?

james-prysm: realized i missed pushing for this, but saw you fixed in https://github.com/ethereum/beacon-APIs/pull/608/commits/300692f3642d039404b60827732aa6a783e539d9

apis/validator/proposer_preferences.yaml
james-prysm: https://github.com/ethereum/beacon-APIs/pull/608/commits/a1bf8d215a33b1656f0e71f3aaf27b8071632d83 hows this?

nflaig: > [a1bf8d2](https://github.com/ethereum/beacon-APIs/commit/a1bf8d215a33b1656f0e71f3aaf27b8071632d83) hows this?

this looks good to me :+1:

apis/validator/register_validator.yaml
nflaig: are we fine with my commit? feel free to push anything you had

nflaig (approved):
LGTM, I really like all the notes for implementers, great work, thanks @james-prysm

PR #610, opened 2026-05-21
Add builders endpoints under /eth/v1/beacon/states/{state_id}

potuz:
Add builder endpoints similar to the validator endpoints. This is useful for builder clients to track and resolve their builder indices, status, balances etc when preparing bids.


apis/beacon/states/builder.yaml
```diff
@@ -0,0 +1,60 @@
+get:
+  operationId: "getStateBuilder"
+  summary: "Get builder from state by id"
+  description: "Returns builder specified by state and id or public key along with status and balance."
```
james-prysm, line 4: nit  
```suggestion
  description: "Returns builder specified by state and builder id."
```

apis/beacon/states/builder_balances.yaml
```diff
@@ -0,0 +1,143 @@
...
+    match any known builder, no balance will be returned but this will not cause an error. There are no guarantees for the
+    returned data in terms of ordering; the index is returned for each balance, and can be used to confirm for which inputs a
+    response has been returned.
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+    - name: id
```
james-prysm, line 17: getStateBuilder calls it `builder_id` maybe we should call it that

types/api.yaml
```diff
@@ -58,6 +58,55 @@ ValidatorStatus:
...
+  required: [index, balance, status, builder]
+  properties:
+    index:
+      $ref: './primitive.yaml#/Uint64'
+      description: "Index of builder in the builder registry."
+    balance:
+      $ref: "./primitive.yaml#/Gwei"
+      description: "Current builder balance in gwei."
+    status:
+      $ref: "#/BuilderStatus"
```
nflaig, line 72: I am definitely against introducing a custom type like this with a `status` field, especially because the `status` in case of builders it much simpler compared to validators

apis/beacon/states/builder_balances.yaml
```diff
@@ -0,0 +1,143 @@
+get:
+  operationId: "getStateBuilderBalances"
```
nflaig, line 2: why do we need a separate api for balances? seems overkill

apis/beacon/states/builder_identities.yaml
```diff
@@ -0,0 +1,74 @@
+post:
+  operationId: "postStateBuilderIdentities"
```
nflaig, line 2: I feel like this shouldn't be required, what's the most amount of builders a consumer needs to query? returning the full `Builder` spec object seems fine to me

apis/beacon/states/builder.yaml
```diff
@@ -0,0 +1,60 @@
+get:
+  operationId: "getStateBuilder"
```
nflaig, line 2: why do we need this? I feel like this was already a mistake in the validators api, can't we just have `getStateBuilders`?

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,168 @@
+get:
+  operationId: "getStateBuilders"
```
nflaig, line 2: having both the get and post endpoint is too much, we should decide on whether we need post here, but if we wanna be on the safe side, likely best to just go with post here

nflaig (commented):
I don't think we need to mirror the validator apis 1:1, especially not the validator apis we know have a bad design (eg. using non spec containers).

There is a big difference in how builders work compared to validators and how consumers will use the api.

- we don't expect builders to be nearly as many as validators, so fetching all builders always will likely be fine
- even if we don't wanna fetch all builders, having `getStateBuilders` as defined in this PR would be sufficient as it allows to pass the `pubkey`
- in contrast to validators, there should not be the use case to fetch thousands of records, it will likely just be a few at most, so returning the full `Builder` object should always be fine

There is likely more, but I think we can have a single api here which allows consumers to fetch all builders, or filter them by pubkey (or index) to just get the builders they care about.

apis/beacon/states/builder_balances.yaml
rolfyone: im not against being easily able to get balances but we can probably just retain the 'post', and remove the get...

nflaig: but why does it need to be a separate endpoint? we can just get that by accessing `builder.balance`

this is in contrast to validators, as balances are separatly stored in the state via `state.balances` which is different from `validator.effective_balance` (actual balance vs. effective balance)

types/api.yaml
nflaig: maybe if we can conclude https://github.com/ethereum/beacon-APIs/pull/601 this is fine, but I really like to avoid having another endpoint that cannot support ssz

the argument for the `status` in the response is to have all computation done on the beacon node side which is also valid and simplifies clients consuming the response

nflaig:
I recommend closing this in favor of https://github.com/ethereum/beacon-APIs/pull/614

rolfyone:
@potuz happy with #614 as an alternative? Wondering if we can close this one...

nflaig:
I am closing this in favor of https://github.com/ethereum/beacon-APIs/pull/614, there is too much wrong with this PR, it has way too many apis (copy-pasted from validators) that are not needed for builders, eg. the balances one. We can start with the minimal api proposed in #614 and add more later if there is a use case / consumer that would benefit from it.

PR #614, opened 2026-05-29
Add POST builders query endpoint

rolfyone:


CHANGES.md
```diff
@@ -15,6 +15,7 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#552](https://github.com/ethereum/beacon-APIs/pull/552) `GET /eth/v1/beacon/pool/payload_attestations` added         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#552](https://github.com/ethereum/beacon-APIs/pull/552) `POST /eth/v1/beacon/pool/payload_attestations` added         |                                                  |                                                  |                                                    |                                                 |                                           |
+| `POST /eth/v1/beacon/states/{state_id}/builders` added         |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig: missing pr link

types/api.yaml
```diff
@@ -57,6 +69,16 @@ ValidatorStatus:
...
 
+BuilderStatus:
+  description: |
+    Possible statuses:
+    - **0** - pending
+    - **1** - active
+    - **2** - withdrawn
+  type: integer
+  enum: [0, 1, 2]
+  example: 1
```
nflaig: we should document somewhere how the statuses are derived?

types/api.yaml
```diff
@@ -57,6 +69,16 @@ ValidatorStatus:
   enum: ["pending_initialized", "pending_queued", "active_ongoing", "active_exiting", "active_slashed", "exited_unslashed", "exited_slashed", "withdrawal_possible", "withdrawal_done"]
   example: "active_ongoing"
 
+BuilderStatus:
+  description: |
+    Possible statuses:
+    - **0** - pending
+    - **1** - active
+    - **2** - withdrawn
```
nflaig: so this value is set if `builder.withdrawable_epoch != FAR_FUTURE_EPOCH`? if that's the case we might wanna call this `exited`, this also matches with `initiate_builder_exit`

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,81 @@
+post:
+  operationId: "postStateBuilders"
```
nflaig: I don't like calling this "post" in the operation id but I guess for sake of consistency we can name it like that.. but really it's `getStateBuilders`, imo expressing the HTTP method in the operation id is strange

types/api.yaml
rolfyone: hopefully this is covered now in the descriptions, let me know what your thoughts are

types/api.yaml
```diff
@@ -57,6 +69,17 @@ ValidatorStatus:
...
 
+BuilderStatus:
+  description: |
+    Possible statuses derived from the builder registry entry and `is_active_builder(state, index)`.
+    Evaluate `exited` first; otherwise use `is_active_builder` to distinguish active builders from pending builders.
+    - **0** - pending: `builder.withdrawable_epoch == FAR_FUTURE_EPOCH` and `is_active_builder(state, index)` returns `False`
+    - **1** - active: `is_active_builder(state, index)` returns `True`, i.e. `builder.deposit_epoch < state.finalized_checkpoint.epoch` and `builder.withdrawable_epoch == FAR_FUTURE_EPOCH`
+    - **2** - exited: `builder.withdrawable_epoch != FAR_FUTURE_EPOCH`
+  type: integer
+  enum: [0, 1, 2]
```
nflaig: echoing what @etan-status mentioned here https://github.com/ethereum/beacon-APIs/pull/601#discussion_r3466209477, we might wanna follow the validators api for now and figure out a proper enum solution later, it's definitely not ideal that this api won't support ssz initially but especially for builders it should not be as problematic as there won't be that many compared to validators

CHANGES.md
```diff
@@ -15,6 +15,7 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#552](https://github.com/ethereum/beacon-APIs/pull/552) `GET /eth/v1/beacon/pool/payload_attestations` added         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#552](https://github.com/ethereum/beacon-APIs/pull/552) `POST /eth/v1/beacon/pool/payload_attestations` added         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#614](https://github.com/ethereum/beacon-APIs/pull/614) `POST /eth/v1/beacon/states/{state_id}/builders` added         |                                                  |                                                  |                                                    |                                                 |                                           |
```
chong-he, line 21: A minor comment here, maybe we want to move this down? As I see that the order of the changes is following the PR number

types/api.yaml
rolfyone: It's fairly academic. 
The thing i don't like about strings also is they also bloat the json output, but it does make it easier because you don't need to decode it i guess.
Also given how hard it is to change ssz this may be many hard forks.

rolfyone:
addressed feedback @nflaig , not going to change teku until output is agreed on, as the status output is non trivial if we go that way.

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,87 @@
...
+              type: array
+              uniqueItems: true
+              description: |
+                An array of builder statuses to filter on.
+
+                If the supplied list is empty (i.e. the value is `[]`) or the property is omitted then builders with all statuses will be returned.
+              items:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/BuilderStatus'
+        example:
+          statuses: [1]
```
nflaig: example needs to be updated

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,87 @@
...
+    Returns filterable list of builders with their status and index.
+
+    Information will be returned for all indices or public keys that match known builders. If an index or public key does not
+    match any known builder, no information will be returned but this will not cause an error. There are no guarantees for the
+    returned data in terms of ordering; both the index and public key are returned for each builder, and can be used to confirm
+    for which inputs a response has been returned.
+
+    Returns 400 if the requested state is prior to Gloas.
+
+    Depending on `Accept` header data can be returned either as JSON or as bytes serialized by SSZ.
```
nflaig: this note needs to be removed, no ssz for now 😢

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,87 @@
...
+  tags:
+    - Beacon
+  parameters:
+    - name: state_id
+      in: path
+      $ref: '../../../beacon-node-oapi.yaml#/components/parameters/StateId'
+  requestBody:
+    description: "The lists of builder IDs and statuses to filter on. Either or both may be `null` to signal that no filtering on that attribute is desired. If no body is supplied then all builders will be returned. To return all active builders, omit `ids` and set `statuses` to `[1]`."
```
nflaig: description reference old enum value

rolfyone: ugh sorry

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,88 @@
...
+            statuses:
+              type: array
+              uniqueItems: true
+              description: |
+                An array of builder statuses to filter on.
+
+                If the supplied list is empty (i.e. the value is `[]`) or the property is omitted then builders with all statuses will be returned.
+              items:
+                oneOf:
+                  - $ref: '../../../beacon-node-oapi.yaml#/components/schemas/BuilderStatus'
```
nflaig: why the `oneOf`?

types/gloas/builder.yaml
```diff
@@ -0,0 +1,23 @@
+Gloas:
+  Builder:
+    type: object
+    description: "The [`Builder`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.8/specs/gloas/beacon-chain.md#builder) object from the CL Gloas spec."
```
nflaig:
```suggestion
    description: "The [`Builder`](https://github.com/ethereum/consensus-specs/blob/v1.7.0-alpha.11/specs/gloas/beacon-chain.md#builder) object from the CL Gloas spec."
```

CHANGES.md
nflaig: I think we can keep it here for now, I will need to clean up the changelog a bit anyways, the last merge of #580 made a bit of a mess, I do like if we have it ordered by PR number

types/api.yaml
nflaig: looks good to me

types/api.yaml
nflaig: > may be many hard forks.

enum ssz type doesn't require a hard fork, it will be optional so once we have a well-defined enum ssz type with a spec we can start implementing this already and clients that do have it can support ssz on the validators and builders api since adding it later is backward compatible

apis/beacon/states/builders.yaml
rolfyone: was based on ValidatorStatus format, what should it be? is validatorStatus incorrect?

nflaig: isn't the one `oneOf` just relevant there because we merge 2 enums? I don't think the `oneOf` here does anything

apis/beacon/states/builders.yaml
```diff
@@ -0,0 +1,87 @@
...
+              type: array
+              uniqueItems: true
+              description: |
+                An array of builder statuses to filter on.
+
+                If the supplied list is empty (i.e. the value is `[]`) or the property is omitted then builders with all statuses will be returned.
+              items:
+                $ref: '../../../beacon-node-oapi.yaml#/components/schemas/BuilderStatus'
+        example:
+          statuses: ["active"]
```
nflaig: this shouldn't be needed at all

```suggestion
```

nflaig (approved):
LGTM

potuz (approved):

james-prysm (approved):

PR #619, opened 2026-06-23
Update to handle non canonical ptc duties

eserilev:
Validators should produce payload attestations even for payloads on non canonical branches. The current PTC duties flow can't express this. Duties are fetched per epoch keyed by a single `dependent_root`, but a validator may have to attest on a branch with a different dependent root. Since any non finalized block root is a possible dependent root, the beacon node cannot realistically compute duties across all branches.

This reworks PTC duties to be beacon node driven per slot. Beacon nodes already track connected validators, so for a given slot it returns the PayloadAttestationData for each block it has processed, paired with a subset of the PTC that the validator controls.


michaelsproul:
Eitan and I discussed this on a call and came up with some reasons _not_ to do non-canonical PTC. votes Mostly they relate to the implications of a minority chain existing:

1. Minority chain participants probably can't cast PTC votes on the majority chain, because they probably haven't imported its blocks. If they are attesting and proposing on the minority chain they must think that it's canonical, which is only really possible if the heavier-weighted chain appears invalid to them (or they can't see it at all, due to a partition).
2. Minority chain transactions (and MEV) are vulnerable to being lifted onto the majority chain, so does protecting minority payloads from being reorged on the minority chain itself actually protect minority chain users/builders that much? The only case where transactions on the minority chain are really "safe" is if the minority are correct, and the majority chain is spec-invalid & will be eventually abandoned. In this case, so long as the faulty nodes on the majority chain ALSO view the minority chain as valid (they may not), then they can PTC attest on the minority chain, and these PTC attestations will be protective.

I guess it boils down to: is that niche use-case described in (2) worth the complexity of handling multi-PTC attesting in VCs all the time?

Maybe my argument in (2) is wrong if MEV is probably not transferable between the two chains? I don't know enough MEV specifics to say whether that is a good hunch or not.

CC @potuz

eserilev:
I'm closing this PR. can revisit if we find it necessary

potuz:
I just got the e-mail for this, apologies that I forgot to write here what I left on Discord @michaelsproul. I think the main reason is not about MEV or minority branches and the such, but rather a matter of having aligned incentives. On any branched scenario in which statistically one does not expect a PTC quorum on either branch, proposers are always rationally incentivized to reorg the payload, even if this is not honest behavior. One could think of this as a variation of proposer boost reorg, proposers are indeed incentivized to reorg weak blocks. Perhaps the lack of honest reorgs under higher than the 20% default threshold is enough evidence to show that this won't happen in the wild. 

At any rate I believe this feature is so low priority that it is not worth the effort nor the complexity to maintain it so I fully support clients not implementing it. However, it is not consensus breaking to keep this as part of the honest behavior guideline in case some client does decide to implement this.

nflaig:
> However, it is not consensus breaking to keep this as part of the honest behavior guideline in case some client does decide to implement this.

does that suggest we should have a beacon-api for it so that if clients choose to implement it they don't need to use custom apis? In a sense, you could just support this on the beacon-node, and leave validator clients actually using it optional. I really dislike the idea of implementing this via custom apis if many clients do it, but we could also just skip it for gloas and re-evaluate for heze

PR #621, opened 2026-06-25
Remove `parent_block_number` from `payload_attributes` post-gloas

eserilev:
Remove parent block number post-gloas. Gloas blocks no longer have block number info, and there are edge cases where serving the parent block number becomes unnecessarily complex. EL's should be able to get the parent block number from the parent block hash.


beacon-node-oapi.yaml
```diff
@@ -349,6 +349,10 @@ components:
       $ref: './types/bellatrix/block.yaml#/Bellatrix/BlindedBeaconBlock'
     Bellatrix.SignedBlindedBeaconBlock:
       $ref: './types/bellatrix/block.yaml#/Bellatrix/SignedBlindedBeaconBlock'
+    Bellatrix.PayloadAttributesEventData:
```
ensi321: You are only defining the new types here. 

You probably need to define a `payload_attributes_v2` topic in `eventstream` to achieve what you want

eserilev: i don't think we do v2 for sse events. i was using `light_client_finality_update` as an example, which changes between altair and capella. but maybe im missing something

I think since payload attributes already contains a version field we dont need to introduce a v2

ensi321: Ah then you need to add something that says something about latest fork's `PayloadAttributesEventData` in the `payload_attributes` topic's description

eserilev: [a9c1893](https://github.com/ethereum/beacon-APIs/pull/621/commits/a9c18933e560382fe944f54f86271b452409ea9f)

beacon-node-oapi.yaml
```diff
@@ -349,6 +349,10 @@ components:
...
+      $ref: './types/bellatrix/payload_attributes_event.yaml#/Bellatrix/PayloadAttributesEventData'
+    Gloas.PayloadAttributesEventData:
+      $ref: './types/gloas/payload_attributes_event.yaml#/Gloas/PayloadAttributesEventData'
```
nflaig: bit skeptical about adding these, what's the purpose?

apis/eventstream/index.yaml
nflaig: it's kinda sad that we can't have multiple schemas or examples per event, but in general, I think the example data we have here should represent the latest fork which in this case is gloas, we should update this

apis/eventstream/index.yaml
```diff
@@ -141,7 +141,7 @@ get:
                 - `proposal_slot`: the slot at which a block using these payload attributes may be
                    built.
                 - `parent_block_root`: the beacon block root of the parent block to be built upon.
-                - `parent_block_number`: the execution block number of the parent block.
+                - `parent_block_number`: the execution block number of the parent block. Removed from `gloas` onwards.
```
nflaig, line 144: removing a field is a bit tricky btw, so I hope we don't break anything by doing this, alternative could be to just set 0 or some other arbitrary value.. not sure if that's better though

we solved this by keeping track of this info in fork choice and can query it like this `get_block(parent_block_root, parent_block_hash).execution_payload_number`

so setting `parent_block_number` is not an issue for us

michaelsproul: is there a case in your impl where the block with that block hash has been pruned? i.e. if all blocks since finalisation are empty and the last applied payload is from a block prior to finalisation?

beacon-node-oapi.yaml
eserilev: Wasn't sure how else to express the different variants

apis/eventstream/index.yaml
twoeths: > is there a case in your impl where the block with that block hash has been pruned? i.e. if all blocks since finalisation are empty and the last applied payload is from a block prior to finalisation?

we transfer `execution_payload_number` from one block to its child so there is no issue in this case. 
But when boostrapping the beacon-node we always don't have that info anyway, there was a [TODO](https://github.com/ChainSafe/lodestar/blob/551b8b0a3d22a678ca16a5ab0d5893d27776dba6/packages/beacon-node/src/chain/forkChoice/index.ts#L152)

so it's better to remove if there is no usage at all

eserilev: oh yeah if we checkpoint sync and the non finalized portion is all empty, I don't even know how we'd get the parent block number.

beacon-node-oapi.yaml
nflaig: (1) there isn't a good way to express this, the problem with the types is that they would just be dangling and not connected to the event at all, so when you looked the event via the rendered api explorer there is no connection

(2) the second part is the usefulness for spec testing, for example in lodestar, we use the value after `data:` to check if we emit the right format which isn't great compared to the schema based validation we have for endpoints and it's not fork-aware, you can always only verify the latest fork

I can't think of any way right now to make them useful for (1) or (2), but it might still be fine to point out the difference in this PR and later when people look through the yaml files, also those will be listed in `Schemas` on the explorer, not sure anyone ever looks at these though.

I have no strong opinion to remove them though, so if you prefer to add them that's fine with me.

nflaig: if we wanna keep this, it needs to be moved to gloas types

eserilev: removed

eserilev: thanks, since it doesn't seem useful i went ahead and removed it

nflaig:
@eserilev can we put a bit more meaningful PR title please

eserilev:
updated the PR title

nflaig:
> updated the PR title

`PayloadAttributes` might be confused with the `payload_attributes` field of the event (ie. the one defined in the execution-apis spec), so I renamed slightly

nflaig (approved):

james-prysm (approved):

tersec (approved):

PR #622, opened 2026-06-30
Add SSZ response support for validator_balances endpoints

Zyra-V21:
## What

Adds an `application/octet-stream` (SSZ) response to `GET` and `POST
/eth/v1/beacon/states/{state_id}/validator_balances`, selectable via the
`Accept` header, plus a `406 Not Acceptable` matching the other SSZ-enabled
state endpoints (`validator_identities`, `pending_partial_withdrawals`).

The SSZ payload is `List[ValidatorBalanceResponse, VALIDATOR_REGISTRY_LIMIT]`,
each entry being the container `{ index: uint64, balance: uint64 }`. It is
purely additive: no JSON changes, no new parameters, JSON consumers are
untouched.

## Context

Split out of #601 per the discussion there (cc @nflaig, @etan-status,
@lodekeeper). `validator_balances` has no `status` and no enum, so the SSZ
enum questions raised for the `validators` endpoint do not apply here. The
`validators` side stays at status quo for now; this is the uncontroversial
half.

## Already shipping

This documents behavior clients already serve. Lodestar has returned SSZ on
this endpoint since ChainSafe/lodestar#6749 (v1.20.0, around two years), with
the same `{index, balance}` shape (confirmed by @lodekeeper in #601). Adding
SSZ to an existing endpoint is backward compatible via `Accept` header content
negotiation.

The concrete consumer is analytics/indexing: MigaLabs' goteth snapshots
balances per processed epoch and backfills large historical ranges, where the
JSON encode/parse cost on both ends is the bottleneck (gzip cuts wire size but
not that CPU). SSZ reduces both.

## Validation

`redocly lint ./beacon-node-oapi.yaml` is valid and `swagger-cli bundle`
succeeds, matching the CI.

## Tests

The spec is validated by lint + bundle rather than a test suite, so there is
no test harness to extend here. If you would like anything beyond that (an
example payload, or a client conformance check), say so and I will add it.

Related: #601, #333.

### DIFF
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -67,6 +67,8 @@ get:
           example:
             code: 404
             message: "State not found"
+    "406":
+      $ref: "../../../beacon-node-oapi.yaml#/components/responses/NotAcceptable"
     "414":
       description: "Too many validator IDs"
       content:
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -90,6 +90,8 @@ post:
     match any known validator, no balance will be returned but this will not cause an error. There are no guarantees for the
     returned data in terms of ordering; the index is returned for each balance, and can be used to confirm for which inputs a
     response has been returned.
+
+    Depending on `Accept` header data can be returned either as JSON or as bytes serialized by SSZ.
   tags:
     - Beacon
   parameters:
@@ -128,6 +130,9 @@ post:
                 type: array
                 items:
                   $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorBalanceResponse'
+        application/octet-stream:
+          schema:
+            description: "SSZ serialized `List[Container{index: uint64, balance: uint64}, VALIDATOR_REGISTRY_LIMIT]` bytes. Use Accept header to choose this response type"
     "400":
       description: "Invalid state ID or malformed request"
       content:
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -8,6 +8,8 @@ get:
     match any known validator, no balance will be returned but this will not cause an error. There are no guarantees for the
     returned data in terms of ordering; the index is returned for each balance, and can be used to confirm for which inputs a
     response has been returned.
+
+    Depending on `Accept` header data can be returned either as JSON or as bytes serialized by SSZ.
   tags:
     - Beacon
   parameters:
@@ -44,6 +46,9 @@ get:
                 type: array
                 items:
                   $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorBalanceResponse'
+        application/octet-stream:
+          schema:
+            description: "SSZ serialized `List[Container{index: uint64, balance: uint64}, VALIDATOR_REGISTRY_LIMIT]` bytes. Use Accept header to choose this response type"
     "400":
       description: "Invalid state or validator ID"
       content:
--- a/apis/beacon/states/validator_balances.yaml
+++ b/apis/beacon/states/validator_balances.yaml
@@ -151,5 +151,7 @@ post:
           example:
             code: 404
             message: "State not found"
+    "406":
+      $ref: "../../../beacon-node-oapi.yaml#/components/responses/NotAcceptable"
     "500":
       $ref: '../../../beacon-node-oapi.yaml#/components/responses/InternalError'


lodekeeper (commented):
LGTM from the Lodestar side — and I can confirm this matches what Lodestar already serves, so it's really just documenting existing behavior (which was the point of splitting it out of #601 👍).

- **SSZ schema matches Lodestar exactly.** `List[ValidatorBalanceResponse, VALIDATOR_REGISTRY_LIMIT]` with `ValidatorBalanceResponse = {index: uint64, balance: uint64}` is exactly what Lodestar has returned since [#6749](https://github.com/ChainSafe/lodestar/pull/6749) (v1.20.0) — `List[Container{index: ValidatorIndex, balance: UintNum64}]`, both `uint64`. So indexers already get this wire format from Lodestar today.
- **Additive + backward-compatible** as intended: JSON response unchanged, SSZ opt-in via `Accept`, and the `406 NotAcceptable` on both GET and POST is correct (the `$ref` resolves — `beacon-node-oapi.yaml#/components/responses/NotAcceptable`).
- **Matches house style** — the `application/octet-stream` description follows the sibling state-list endpoints (`pending_consolidations`, `proposer_lookahead`, `pending_partial_withdrawals`); the extra "Each `ValidatorBalanceResponse` is the SSZ container `{index, balance}`" line is a helpful clarification since the JSON schema uses string-encoded uints.

One minor nit (non-blocking): the two new `CHANGES.md` rows are missing the `[#622](https://github.com/ethereum/beacon-APIs/pull/622)` PR-link prefix that every other row in that table carries — worth adding for consistency.

Thanks for splitting this out!

Zyra-V21:
Sure, will fix that soonish

CHANGES.md
```diff
@@ -33,11 +32,8 @@ There are likely to be descriptions etc outside of the list below, but new query
...
-| [#608](https://github.com/ethereum/beacon-APIs/pull/608) `POST /eth/v1/validator/beacon_committee_subscriptions` updated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| `GET /eth/v1/beacon/states/{state_id}/validator_balances` SSZ response added                                                           |                                                  |                                                    |                                                 |                                           |
+| `POST /eth/v1/beacon/states/{state_id}/validator_balances` SSZ response added                                                           |                                                  |                                                    |                                                 |                                           |
```
nflaig: I don't think we need changelog entries for these, we did this before eg. here https://github.com/ethereum/beacon-APIs/pull/527 without adding entries

apis/beacon/states/validator_balances.yaml
```diff
@@ -44,6 +46,12 @@ get:
                 type: array
                 items:
                   $ref: '../../../beacon-node-oapi.yaml#/components/schemas/ValidatorBalanceResponse'
+        application/octet-stream:
+          schema:
+            description: |
+              SSZ serialized `List[ValidatorBalanceResponse, VALIDATOR_REGISTRY_LIMIT]` bytes.
+
+              Each `ValidatorBalanceResponse` is the SSZ container `{ index: uint64, balance: uint64 }`. Use `Accept` header to choose this response type.
```
nflaig: what do you think about the following

```suggestion
            description: "SSZ serialized `List[Container{index: uint64, balance: uint64}, VALIDATOR_REGISTRY_LIMIT]` bytes. Use Accept header to choose this response type"
```

CHANGES.md
lodekeeper: Good point — I'll withdraw that nit. I'd flagged the missing `[#622]` link on these rows, but if we're following #527 and skipping changelog entries for this, that's moot — better to just drop the two `CHANGES.md` rows than add the link.

apis/beacon/states/validator_balances.yaml
Zyra-V21: thanks both! applied the suggestion (single-line description with List[Container{index: uint64, balance: uint64}, VALIDATOR_REGISTRY_LIMIT]) on both get and post, and dropped the changelog entries per #527. also rebased onto master to clear the conflict from #608, so it's just the validator_balances.yaml change now. lint passes !

lodekeeper: Verified — both GET and POST now use the single-line `List[Container{index: uint64, balance: uint64}, VALIDATOR_REGISTRY_LIMIT]` description, the CHANGES.md entries are dropped, and after the rebase the diff is just `validator_balances.yaml`. Still matches exactly what Lodestar serves. LGTM — thanks for the quick turnaround! 👍

nflaig (approved):
LGTM, thanks @Zyra-V21 for working on this, it's long overdue to add this to the spec, we've already implemented ssz response for this endpoint ~2 years ago.

james-prysm (approved):

rolfyone (approved):
LGTM

Issue #519, opened 2025-04-01
`produceBlock` and `publishBlock`: Remove blobs and KZG proofs


nalepae:
### Current Situation

Currently, when a validator proposes a block, the workflow is as follows:

1. The validator client requests a (possibly blinded) block from the beacon node using [produceBlockV3](https://ethereum.github.io/beacon-APIs/#/Validator/produceBlockV3). The response's `data` field contains `block`, `kzg_proofs`, and `blobs`. **Note:** The `kzg_proofs` and `blobs` fields may be empty if the returned block is blinded.
2. The validator client signs the **block** and returns `signed_block`, `kzg_proofs`, and `blobs` via [publishBlockV2](https://ethereum.github.io/beacon-APIs/#/Beacon/publishBlockV2).
3. The beacon node builds blob/data column sidecars and then publishes the signed block and sidecars over the P2P network.

In this workflow, the validator client does not use `blobs` and `kzg_proofs` directly. It fetches them from the beacon node in step 1 and sends them back in step 2.

#### Pros of the Current Approach
- **Stateless Design:** The validator client can be connected to multiple beacon nodes (e.g., via a load balancer). It can send `produceBlockV3` to one beacon node and `publishBlockV2` to another.
- **Simplicity:** The beacon node receives all the necessary data within `publishBlockV2`, allowing it to directly publish the block and sidecars on the P2P network.

#### Cons of the Current Approach
- Some data (`kzg_proofs` and especially `blobs`) is redundantly transferred twice—once from the beacon node to the validator client, and again from the validator client back to the beacon node. This redundancy is manageable with a low blob count but becomes increasingly significant as blob sizes grow, particularly if the beacon node and validator client run on separate servers.

### Proposal

The general idea is to eliminate the redundant transmission of `kzg_proofs` and `blobs` in both `produceBlock` and `publishBlock`. This can be achieved in two ways:

#### **Proposal 1:** Modify Existing Endpoints
- Retain `produceBlockV3` and `publishBlockV2`.
- Modify the implementation of `produceBlockV3` in the beacon node so that it no longer includes `kzg_proofs` and `blobs`.
- No changes are required on the validator client side.

#### **Proposal 2:** Introduce New API Versions
- Introduce `produceBlockV4` and `publishBlockV3`, which remove unnecessary data fields.

Instead of returning:

```json
{
  "version": "electra",
  "execution_payload_blinded": false,
  "execution_payload_value": "12345",
[...]

PR #624, opened 2026-07-08
remove BlindedExecutionPayloadEnvelope

nflaig:
While looking at the implementation of https://github.com/ethereum/beacon-APIs/pull/580 I realized using a `BlindedExecutionPayloadEnvelope` adds quite some complexity and I don't think it's warranted to add it to the spec
- (1) requires a second signing route for the same signature, validator clients, remote signers and DVT
    middleware would all have to support this
- (2) might be implementation specifc but at least for us having a object without a slot adds some complexity, initially when we discussed adding `BlindedExecutionPayloadEnvelope`, the slot was still a top-level field but we moved it to `payload.slot_number` so it's not part of the blinded object anymore
- (3) on top of that, it requires to maintain a new type and adds more complexity to caching and reconstruction on the beacon node side during publishing, which adds more failure modes to a hot path, not ideal

There is probably more, I stopped the implementation when I realized using a blinded object was a bad idea, as it also just optimizes not having to send `transactions` and `block_access_list` between VC<>BN, everything else in the envelope is a few hundred bytes. While in theory, this seems like the ideal way to do it from a performance perspective (you could even just send the root), I don't think it makes sense as it's rather a micro optimization compared to not sending around the blobs and proofs, which this PR still retains, and previous concerns from https://github.com/ethereum/beacon-APIs/issues/519 are still being addressed in gloas.

**Note:** This PR includes several smaller follow-ups to #580 that are opinionated, please take a look at my comments below with further rationale, I also just polished a few descriptions here and there, shouldn't be too controversial


eserilev (approved):
Seems to be a nice simplification. Blobs were the real bottleneck here, blinding the payload was probably overkill

tersec (dismissed):

potuz (approved):
Haven't read anything but I do approve the notion of avoid sending the blob data but keeping the payload there to avoid the existence of a blinded object. Blob data is already logically separated in our code base and it makes sense to keep it so here. But blinding the payload is unnecessary extra complexity.

apis/beacon/execution_payload/envelope_post.yaml
```diff
@@ -56,39 +54,38 @@ post:
       name: Eth-Consensus-Version
       description: "The active consensus version to which the execution payload envelope being submitted belongs."
     - in: header
-      name: Eth-Execution-Payload-Blinded
+      name: Eth-Blob-Data-Included
```
nflaig, line 57: we can nit pick the header name, that was the best name I could come up with

apis/beacon/execution_payload/envelope_post.yaml
```diff
@@ -100,3 +97,5 @@ post:
       $ref: "../../../beacon-node-oapi.yaml#/components/responses/UnsupportedMediaType"
     "500":
       $ref: "../../../beacon-node-oapi.yaml#/components/responses/InternalError"
+    "503":
```
nflaig, line 100: not really related, but this PR has some follow ups for https://github.com/ethereum/beacon-APIs/pull/580, so might as well include  that, at least in our implementation we return `503` if node is syncing, and I couldn't come up with a good reason not to do that, we do the same in `publishBlockV2` which has the 503 documented on the spec side

https://github.com/ethereum/beacon-APIs/blob/4f6a0a757374ad6617ac57c138353436aedd601f/apis/beacon/blocks/blocks.v2.yaml#L82-L83

apis/validator/block.v4.yaml
```diff
@@ -42,12 +43,12 @@ get:
...
     - name: include_payload
       in: query
-      required: false
+      required: true
       description: |
         Controls whether the execution payload envelope and blobs are included in the response
         when self-building (using local execution payload).
 
-        When `true` (default), the response includes the full block contents: beacon block,
+        When `true`, the response includes the full block contents: beacon block,
```
nflaig, line 51: two reasons for changing this
- (1) I do not like optional parameters in general
- (2) I want to de-opinionate the spec about what should be the default

eg. for (2), in lodestar we look at how many beacon node(s) are connected and if it's > 1, then we default to stateless, otherwise stateful, and allow to override via separate cli flag. All of this is configured on the vc side though, so making this parameter optional would mean the bn would have to decide if omitted, but that's not easily possible since the bn has no good view on whether or not other bns are connected to the vc that is requesting to prodcue a block

beacon-node-oapi.yaml
```diff
@@ -543,11 +539,7 @@ components:
       schema:
         $ref: '#/components/schemas/ConsensusVersion'
     Eth-Execution-Payload-Blinded:
-      description: |
-        Indicates whether the execution payload (or envelope) is in blinded form. Sent in
-        responses so clients can deserialize returned JSON or SSZ data to the correct object,
-        and sent in requests on endpoints that accept either blinded or full forms so the
-        server can select the correct request schema.
+      description: Required in response so client can deserialize returned json or ssz data to the correct object.
```
nflaig, line 542: this just reverts back to pre https://github.com/ethereum/beacon-APIs/pull/580 description

qu0b (approved):

KaloyanTanev (approved):

eserilev (approved):

hangleang (approved):

apis/beacon/execution_payload/envelope_post.yaml
james-prysm: Eth-Contains-Blob-Data ? Eth-Includes-Blob-Data? Eth-With-Blob-Data?

james-prysm (approved):
added a small list of suggestions for renames mostly nits

apis/beacon/execution_payload/envelope_post.yaml
nflaig: we do have `Eth-Execution-Payload-Included` already, so I do prefer the `-Included` suffix to at least keep consistency between those two headers, but appreciate the suggestions

apis/beacon/execution_payload/envelope_post.yaml
eth2353: What if the beacon node _mistakenly_ thinks it's syncing, like we saw happen during the Holesky incident non-finality when very few blocks were being produced? Should the BN refuse a perfectly valid request to publish a payload envelope?

eth2353 (approved):
Simplifying this seems like a good idea to me too, the main worry was about passing around the much larger blob data which is still addressed 👍

apis/beacon/execution_payload/envelope_post.yaml
nflaig: this is a bit tricky, the problem is that while the node is syncing it might not be able to validate the payload envelope easily, or have a bad view of the network state since it's not subscribed to gossip usually, so equivocation checks may not be possible

but I see what you are saying, this has been mostly addressed on the beacon node side to be able to configure this, eg. there is a flag in lodestar with which I can force it to build on very old heads. there is a trade-off there, but those scenario are being considered as part of the hardening efforts already

curious about other implementations and whether or not they return 503 currently

shane-moore (approved):

Issue #620, opened 2026-06-24
Handling of bid selection


potuz:
Currently when the VC requests a beacon block to sign from the BN, it passes information that helps the BN picking a payload, for example `builder_boost_factor`. After Gloas, the VC will still need to pass information that helps the BN pick a bid instead of a payload. But this information can now be much more granular.

## Configuration vs endpoint

Clients for the most coded their own way of getting these flags from the node operator. This is something that has been a pain point for institutional node operators and staking pools and has been a constant request even from homestakers and educators, we should try to at the very least come up with a list of configuration flags that can be standardized across clients, and then some parameters that must be accepted and dealt with by the beacon node (for example the above mentioned `builder_boost_factor`).  For the most part, these flags are passed in the validator client and they can be tweaked per validator key, that is, clients currently support chooseing a different `builder_boost_factor` per different validator key they track.  The VC then has a map `validator_key -> builder_boost_factor` as its configuration. And when requesting a block to sign sends the correct `builder_boost_factor` to the beacon node.  Clients are free to add a flag to default this value, that is if I do not specify a value that overrides the default boost factor, the map is self-filled with the default. 

## Per builder flag. 

With Gloas, validator clients get to adjust bid selection on a per-builder basis. That is, the VC would now have a map `validator_key -> (map builder_id -> builder_boost_factor)`. This is the main crux of the configuration problem, since the bid selection is made by the beacon node, but the configuration is at the validator client instance, the beacon node **has to receive** a map in the endpoint. Instead of getting a `builder_boost_factor` it needs to receive something equivalent to the information `builder_id -> builder_boost_factor` so that it can apply the correct factor to each builder. 

## Standard parameters. 

Since we now can be more opinionated and nuanced on bid selection, I believe the endpoint thus needs to accept a map `builder_id -> BuilderPreferences` where `BuilderPreferences` is an object that collect the many different flags that are applicable on a per-builder basis. Examples of flags that I believe every node must handle are

- `builder_boost_factor` a parameter that would make this builder more or less likely to be accepted. 
- `max_trusted_bid` either discard any bid from a builder that offers an off-protocol payment higher than `X` or simply treat the bid value as if it were `X` disregarding if it's offering more. 
- `min_bid` only accept bids from this builder if they offer at least this much. 

There are many other possible flags that can be added as optional, and some that can be constructed simply from the above list that is the most skeletal I can think of that need to be implemented by all beacon  nodes. 

[...]

PR #625, opened 2026-07-08
Add a POST produceBlockV4

potuz:
This PR adds a POST version of produceBlockV4 in which the validator sends a list of `BuilderPreferences` objects. These object specify a per-builder list of preferences that helps the beacon node chose the right payload bid contained in the block to be produced. This solves issue #620.


apis/validator/block.v4.yaml
```diff
@@ -155,3 +155,145 @@ get:
       $ref: '../../beacon-node-oapi.yaml#/components/responses/InternalError'
     "503":
       $ref: '../../beacon-node-oapi.yaml#/components/responses/CurrentlySyncing'
+post:
```
nflaig: there shouldn't be 2 apis, if we want to pass `BuilderPreferences` around when producing blocks I would rather modify the existing `produceBlockV4` and make it an optional parameter

we could also hex encode it and pass as query param? that way we could keep the `GET`, I don't expect builder preferences to be massive objects (but using `POST` is likely safer)

potuz: yes, as stated in Discord, only opening this one to get the ball rolling, if something like this is taken I'll delete the get version and just keep this post one.

eth2353:
I was unable to make ACDC earlier today but here is some feedback from my different points of view.

___

**Vero validator client maintainer:**

All per-key configurability in Vero is handled through the Keymanager API which is standardized across all CL/VC clients. This is the only way Vero allows node operators to specify certain things, like different fee recipient per validator key. I hate the idea of adding support for some kind of YAML file (that hasn't even been standardized) instead of doing this in a more standard way using the Keymanager API. We should seriously stop with these CL-client-specific YAML files, we have the standard Keymanager API for a reason.

Setting these various per-key per-builder preferences via the Keymanager API presents some challenges but I'd still prefer solving those over the YAML files.

Switching CL/VC clients should be easy and different YAML configurations make that unnecessarily hard, whereas the Keymanager API makes it trivial.

___

**Staking-as-a-service provider / node operator:**

We don't expect to use the advanced functionality described in #620 and I also expect 99% of other node operators not to.

In practice, I expect the vast majority of node operators will be running a much simpler configuration similar to one of these:

A) trustless P2P bids only (no CLI flags, this could be the default clients ship with)
B) trustless P2P bids + specified builder URLs (`--builder-urls=http://...` , similar to how relays URLs are provided today)
C) specified builder URLs only - I expect entities that censor transactions may want to use this kind of config (`--builder-urls=...` + perhaps something like `--disable-p2p-bids` ?). I don't think they'd be using a blocklist of builders for this purpose since it could easily be circumvented by non-censoring builders.

Will these 3 configurations actually cover 99% of the node operators? I don't know. But if they do, we should really avoid overcomplicating things.

___

I think we should consider making the 1% of super-advanced node operators that _MAY_ want to run a ton of different builder configurations on different subsets of their keys to simply run a corresponding number of validator client instances with different configurations. If they need more than a few, that might require a bit of automation, but so would populating a YAML file, so I don't think that makes things significantly more difficult for them. Whereas supporting these per-key per-builder configuration settings does add a considerable amount of maintenance burden/tech debt across all CL/VC codebases.

Hope this makes sense, ePBS itself is honestly quite complex on its own already so I may have overlooked something.

JasonVranek:
Agree with @eth2353 on the Keymanager API point, and tried to implement here ethereum/keymanager-APIs/pull/87 with those specs, per-key block-production preferences (fee recipient, gas limit, graffiti, builder list and constraints) as one atomic document per key. The schema follows the BuilderEntry shape from OffchainLabs/prysm#17124 which I think captures the real use cases well (e.g. the optional proxy endpoint, with the builder URL staying the signed identity), and adopts the preference semantics from this PR, including builder_boost_factor and its reserved values. 

I understand this PR is for VC to BN but since the Keymanager API sits upstream (user/tool to VC) it needs to carry a superset of what the beacon api will use. So standardizing both makes sense + agreed we dont want to have to support N client-specific YAML dialects.

ofc the big open question is which knobs are truly per-builder, and which can be per-key only? e.g., max_execution_payment makes sense as per-builder but probably min_bid is more suitable as per-key. In the Keymanager PR I just defaulted to maximal flexibility for now.

shane-moore:
Cross-repository sync note from the SSV side: I raised a `RequestAuthV1` construction and signing issue in [builder-specs #165](https://github.com/ethereum/builder-specs/pull/165#issuecomment-4982724911). Since this PR copies that type, could we make the following corresponding changes here?

I think builder-specs should remain normative for construction and signing. This PR only needs to preserve the resulting signed object across the Validator API boundary:

1. In `types/gloas/request_auth.yaml`, specify that `data` contains the canonical ASCII bytes of the builder identity URL. These are the `ByteList` contents used by the signed SSZ object and are represented as hex in JSON. Also specify that `slot` is the proposal slot the authorization applies to, not the slot when it is signed or sent.
2. In `types/builder_preferences.yaml`, require the beacon node to forward `signed_request_auth` without changing its decoded `message` or `signature`. In particular, it must not canonicalize or reconstruct `message.data`.

potuz:
> I think builder-specs should remain normative for construction and signing. This PR only needs to preserve the resulting signed object across the Validator API boundary:

Agree, will update (if this PR is the one adopted) when the builder API is stabilized, but I think the Auth data message should not be specified to be the URL but rather arbitrary data that can be negotiated off-band with the builder. Some builders would want to see here an URL, some others something more generic... there is no need to specify this data to be anything in particular.

JasonVranek:
> [...] I think the Auth data message should not be specified to be the URL but rather arbitrary data that can be negotiated off-band with the builder. Some builders would want to see here an URL, some others something more generic... there is no need to specify this data to be anything in particular.

Agreed with @shane-moore the canonicalizing should happen as the first step, not after a RequestAuthV1 was made. Also agree with @potuz that opaque auth data bytes is more expressive / future proof. An alt proposal is to drop all notions of "canonicalization" and just use auth data bytes as-is. 

Cross referencing to the ethereum/keymanager-APIs/pull/87 PR, a builder entry could look like this, which is unambiguous for VC/BN/ DVT setups etc.

```json
{
  "url": "https://builder-a.example.com",
  "auth_data": "0x123123123....", // new
  "pubkey": "0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a",
  "proxy": "http://side-car:9001",
  "max_execution_payment": "250000000",
  "min_bid": "10000000",
  "builder_boost_factor": "100"
}
```

The wrinkle is that before, a proxy could extract the builder url from the auth data to route. Since the SignedRequestAuth wouldn't be required to carry this anymore, either:

1. the type could include the destination builder url (ugly)
2. submitBuilderPreferences and getExecutionPayloadBid callers add an optional `Eth-Builder-Url`​ header. Note since `auth_data` is bilaterally agreed upon, an unsigned routing header is safe since a mis-routed request fails bc of the wrong `auth_data`.

wdyt about 2?

shane-moore:
From the SSV/DVT side, option 2 works. The same builder-provided auth_data for a validator-builder relationship would need to be configured across all operators so they construct the same RequestAuthV1 signing root.

The builder URL and proxy destination can remain unsigned routing metadata. If Eth-Builder-Url misroutes the request, the receiving builder should reject it because the exact signed auth_data does not match what it negotiated. So I do not think SSV requires the URL to be included in RequestAuthV1.

JasonVranek:
Both PRs updated following this convention:
- https://github.com/ethereum/keymanager-APIs/pull/87/commits/9f66e70effaf6225a5006d7bbf7edb10590ffafa
- https://github.com/ethereum/builder-specs/pull/165/commits/5078eabc63cf092007bb933ae16209087c053eba

michaelsproul:
> A) trustless P2P bids only (no CLI flags, this could be the default clients ship with)
B) trustless P2P bids + specified builder URLs (--builder-urls=http://... , similar to how relays URLs are provided today)
C) specified builder URLs only - I expect entities that censor transactions may want to use this kind of config (--builder-urls=... + perhaps something like --disable-p2p-bids ?). I don't think they'd be using a blocklist of builders for this purpose since it could easily be circumvented by non-censoring builders.

> Will these 3 configurations actually cover 99% of the node operators? I don't know. But if they do, we should really avoid overcomplicating things.

Don't we need a POST for use-cases B and C? We need to include the `SignedRequestAuthV1`, even if no other builder preferences are configured.

This thread on Eth R&D: https://discord.com/channels/595666850260713488/874767108809031740/1522112101714362448

iurii-ssv:
Following up on @shane-moore's [earlier note](https://github.com/ethereum/beacon-APIs/pull/625#issuecomment-4982915296) from the SSV side — we reviewed the convention this thread converged on, now implemented in [builder-specs #165](https://github.com/ethereum/builder-specs/pull/165) (opaque pre-agreed `data` signed exactly as serialized, unsigned `Eth-Builder-Url` routing, byte-preserving forwarding): LGTM, implementable by SSV without protocol headaches — and of the two models discussed, we prefer this one over deriving `data` from a canonicalized builder URL.

The reason for that preference: reconstructing a threshold signature requires byte-identical messages from every operator either way, so URL canonicalization only helps if every validator implementation and every builder normalize URLs bit-identically — a silent cross-implementation failure mode. Opaque bytes signed as-is turn it into plain config discipline: distribute the exact per-builder `data` bytes to all operators, done.

With `data` a pure function of static config and `slot` the proposal slot, all operators in an SSV cluster can independently construct the identical `RequestAuthV1` at duty-discovery time (thanks to proposer lookahead), threshold-sign it ahead of the slot, and attach the reconstructed signature at proposal time. For us this is one new signing duty, same shape as validator registrations today but without the timestamp-determinism workarounds.

Putting on record the properties this relies on, so future revisions keep them:

1. `RequestAuthV1` contains no per-signer or time-of-signing fields (no pubkey, timestamp, or nonce) — N operators must be able to sign the identical message independently.
2. Signing and verification operate on the message exactly as serialized, under the pinned `DOMAIN_REQUEST_AUTH`; no hop (VC, BN, proxy, builder) transforms `message.data`.
3. The beacon node forwards `signed_request_auth` byte-for-byte unchanged — 👍 to mirroring the forwarding MUST from builder-specs #165 in `types/builder_preferences.yaml` when this PR syncs, as discussed above.
4. The empty-`url` default entry ("one config applied to all validators") remains a fully supported mode — SSV clusters have no per-validator channel for builder preferences, so cluster-level defaults are how we'd use this API.

One suggestion: it would help to RECOMMEND a default value for `data` for the case where nothing was agreed out of band — the simplest being the builder's URL exactly as advertised. In practice most proposers will just copy builder URLs from a public list without ever talking to the builder (like relay lists today), and they need to know what bytes to sign. With a spec-suggested default, that plug-and-play case keeps working out of the box, while any builder and proposer that do have a relationship can still agree on whatever custom value they like.

One question: for the ahead-of-time channel (`submitBuilderPreferences`, which gates `max_execution_payment`) — is the expectation that VCs call builders directly in the epoch prior to proposing, possibly via the `Eth-Builder-Url` proxy pattern, or is a BN-mediated endpoint planned? This PR covers the just-in-time attach at `produceBlockV4` only. Either works for us; we'd just like to plan the integration.

types/builder_preferences.yaml
```diff
@@ -0,0 +1,63 @@
+BuilderPreferences:
```
chong-he, line 1: I wonder should we put this file under the directory `types/gloas/builder_preferences.yaml` rather than `types/builder_preferences.yaml`? As `BuilderPreferences` is a Gloas thing

chong-he (commented):
Left a small comment below. 

On a side note, this is about replacing the current `GET` to `POST` endpoint. Maybe we can rename the title a bit? Something like "replace `GET` with `POST`"

JasonVranek:
> 4\. The empty-`url` default entry ("one config applied to all validators")

https://github.com/ethereum/builder-specs/pull/165/commits/31712daba1e3f6ccc2abd774904d6556f57ba6c0 adds clarity that the builder URL SHOULD be used as the default interpretation of the auth `data` bytes

types/builder_preferences.yaml
```diff
@@ -0,0 +1,63 @@
...
+      allOf:
+        - $ref: "./gloas/request_auth.yaml#/Gloas/SignedRequestAuthV1"
+        - description: |
+            Optional `SignedRequestAuthV1` object, as defined in the
+            [Builder API](https://ethereum.github.io/builder-specs/) specification, that the beacon
+            node MUST forward to the builder when requesting a bid so that the builder can
+            authenticate the request. If omitted, requests to this builder are sent without
+            authentication.
+    url:
+      type: string
```
ethDreamer, line 16: Assuming this `BuilderPreferences` here is the same as the [`BuilderPreferencesV1`](https://github.com/ethereum/builder-specs/blob/25aaa5c39b1e7e9af15e198f7e56b927dfdb7862/specs/gloas/validator.md#builderpreferencesv1) that is tentatively defined in the current builder-spec.. if we make this change, how are we supposed to ssz serialize / deserialize a `BuilderPreferences` object which now contains a  field of type `string`?

types/builder_preferences.yaml
```diff
@@ -0,0 +1,63 @@
...
+    signed_request_auth:
+      allOf:
+        - $ref: "./gloas/request_auth.yaml#/Gloas/SignedRequestAuthV1"
+        - description: |
+            Optional `SignedRequestAuthV1` object, as defined in the
+            [Builder API](https://ethereum.github.io/builder-specs/) specification, that the beacon
+            node MUST forward to the builder when requesting a bid so that the builder can
+            authenticate the request. If omitted, requests to this builder are sent without
+            authentication.
+    url:
```
nflaig, line 15: this PR deviates quite a lot from what you mentioned in https://github.com/ethereum/beacon-APIs/issues/620

so first of all the `BuilderPreferences` as defined here do not support p2p builders at all, the `builder_id` concept is missing from this PR. Can you explain why we no longer wanna support this?

apis/validator/block.v4.yaml
```diff
@@ -7,6 +7,11 @@ get:
   description: |
     Requests a beacon node to produce a valid block, which can then be signed by a validator.
 
+    The validator supplies in the request body a list of `BuilderPreferences`, one for each
+    builder that will be contacted, which the beacon node MUST apply when requesting bids from
+    external builders. The `builder_boost_factor` is expressed per builder in the request body
+    rather than as a query parameter.
```
nflaig, line 13: > The `builder_boost_factor` is expressed per builder

today we use this parameter as a global flag (applied to all builders) to let the operator decide their block production strategy

eg. you can set `builder_boost_factor=0` which basically says "always pick local unless the EL fails to produce the payload", or `builder_boost_factor=90` which says "pick local payload if builder payload(s) are only marginally more valuable"

having `builder_boost_factor` and also `min_bid` expressed per builder works with directly connected builders via api, but how does that work with p2p builders?

I think it would be good to have a way to globally apply builder boost factor and min bid


in https://github.com/ethereum/beacon-APIs/issues/620 there is the following statement

> The default BuilderPreferences can be sent explicitly as a default to the endpoint or just be assigned to the builder_id being the max uint.

this seems to suggest a way to do this? I don't see it expressed in this PR though

nflaig (commented):
@potuz this PR doesn't solve https://github.com/ethereum/beacon-APIs/issues/620, see my comments (also cc @JasonVranek I left the comments here as this is easier to review than your PR, I think the points apply to both PRs though)

apis/validator/block.v4.yaml
JasonVranek: Definitely an oversight on my side since I based 630 on 625. 

To match the current conventions in beacon 630 / builder 165 / keymanager 88, p2p bids would be configurable per-key, per-builder like: "validator X accepts p2p bids from builder Y only above this min_bid, weighted by this builder_boost_factor." I'm personally not sure that's worth it. p2p bids are entirely trustless payments so a Gwei from one p2p builder is the same as a Gwei from another, and there's no trust dimension left to express per builder. The thing you might genuinely want to differentiate on is censorship but you can't read that off a bid. If the node operator's goal is to "blacklist" specific p2p builders, the builders could always get around by rotating keys. Also most builders worth configuring individually would likely also be reachable over the builder API, where the full knob set already applies.

I'm more partial to per-key p2p preferences. The pragmatic version is to bring back something like the `builder_boost_factor` query param on produceBlockV4 and treat these as per-key p2p knobs. Then either VCs hold them in their own config as a global, or we extend setBuilders [here](https://github.com/ethereum/keymanager-APIs/pull/88) to carry them per key.

Curious what others think.

nflaig: > "validator X accepts p2p bids from builder Y only above this min_bid, weighted by this builder_boost_factor." I'm personally not sure that's worth it

yes I don't think it's worth it but neither is that worth it on the builder-api, this is not used today and will not be useful after gloas, builder boost factor and also min-bid is very hard to configure even just for local vs. builder payload, having this per builder doesn't seem very useful, but I rather implement that at this point and trying to debate it. But I don't see why we don't wanna support the full spectrum, all that seems to be required is add a `builder_pubkey` field, and make `url` optional and if configured the beacon-node reaches out to the builder via api too + enforcing same policy to the bids of that builder on p2p

JasonVranek: Following [this thread](https://github.com/ethereum/keymanager-APIs/pull/88#discussion_r3653782598), I updated #630 and [#88](https://github.com/ethereum/keymanager-APIs/pull/88) to offer the same config knobs for p2p bids. tl;dr is including `builder_pubkey` and excluding `url` means your `BuilderEntry` supplies `min_bid` and `builder_boost_factor` constraints on the p2p bids coming from `builder_pubkey` so the BN can apply them when determining the winning bid. Keymanager then needed to change to allow defaults to be configured.

nflaig: was thinking about this more and getting more convinced about your point of not needing per-`BuilderEntry` for p2p builders, maybe that also keeps the configuration simpler for users if we only allow global `builder_boost_factor` and `min_bid` to affect p2p bids while for directly connected builders we keep the per-builder configs. So since you made changes to support global parameters now, we might be fine with those only and can avoid overloading the usage of `builder_pubkey`.

one of the reasons is also the following p2p rule
> [IGNORE] this bid is the highest value bid seen for the tuple (bid.slot, bid.parent_block_hash, bid.parent_block_root)

since `builder_boost_factor` and `min_bid` are local only, it would be kinda random based on how bids are propagated, so even if you boost a certain builder's bid by 100x, it might not even reach your node due to the rule above, so it might be better from a user perspective to see p2p bids as essentially a single (the highest value) bid as coming from a single builder (not sure that framing makes sense).

The use cases outlined by potuz in https://github.com/ethereum/beacon-APIs/issues/620 sound pretty complete to me already, so if we can cover those which I think we can now since we added the global parameters, then we are kinda getting close to the final design we want.

image

@JasonVranek maybe you can double check this

JasonVranek: If I remove the per-builder p2p configs in favor for just the global it would definitely simplify. Overloading `builder_pubkey` is confusing / error prone. Regarding this list, *almost* everything would be covered. 

The one exception is by dropping per-builder p2p configs, you can't blacklist a specific **p2p** builder by pubkey via **config**. But like you said the p2p rule might make this moot anyways. 

I'll make the simplification across the PRs today (I think the keymanager interface can stay as-is since I think it's important for the API to be able to configure the global p2p defaults).

nflaig: > you can't blacklist a specific p2p builder by pubkey via config.

@potuz is this something we need/want? I feel like this isn't that relevant

potuz: I don't see why we should have a separate object for p2p builders and direct-connection builders. In fact I kinda don't like enshrining this separation in the beacon API spec (I guess it's fine in the builder API one). There are reasons to want to have a per-builder config on the beacon API even if they are being applied to p2p builders. The reason being that specialized types of builders become viable in Gloas. Imagine an application that settles on-chain, doesn't care about latency and has purely private order flow (a chess-like game). This application may want to eventually submit full blocks on the P2P stack with a decent bid. At times of low congestion they may actually win since their POF would be superior to the regular builders. I hope many different dapps that don't have strong pressure for short inclusion may want to move to such a system and can advertise a p2p builder. 

In such a case some validators may want to boost some of these apps. This is just  one example of that type of builders. Another one would be a client team having an altruistic builder, purely trusted not to front-run and sending cheap bids and returning tips to participants. Some people may want to boost these types of  known altruistic builders setting the "floor" on the auction. 

At any rate, I dot not see any benefit in not allowing this on the API itself. A client that wants to use a global default just needs to fill all values with the same and then add a canary "default entry"  that has the same values.

nflaig: @potuz  yeah I am going back and forth on this whether it's useful, do you have good design in mind how we can express that over the api? my idea was just to overload the use of `builder_pubkey` is pubkey the right one to use, or should it be by index for p2p builders? This PR currently doesn't support per-p2p-builder preferences, the one that @JasonVranek didn't support it either, then I brought it up there and current state should support it. I am just not sure right now what is the clean design for users to configure on the keymanager and also to support it over the wire

JasonVranek: #630 does currently support per-builder p2p bidding configs. Might be easiest to understand the complexity/footguns via this [implementer guide](https://github.com/JasonVranek/epbs-config-flow/pull/1). 

In other words, as-is, 630 checks off anything in the list at the cost of some extra complexity. I'm happy to edit it if the complexity isn't justified. 

Personally I feel like today's profile of builders / node operators / apps likely won't use these knobs, but maybe something like what @potuz described could be desirable in the future. So I'd lean towards keeping our options open, especially now when it's arguably the least friction time to add this. That being said, I'm not opinionated on p2p configs and this seems more like a client dev workload question.

nflaig:
- superseded by https://github.com/ethereum/beacon-APIs/pull/630

PR #630, opened 2026-07-21
Add gloas builder flow: produceBlockV4, builder preferences, block forwarding

JasonVranek:
The Beacon API side of the Gloas builder flow, aligned with [ethereum/builder-specs#165](https://github.com/ethereum/builder-specs/pull/165) and [ethereum/keymanager-APIs#88](https://github.com/ethereum/keymanager-APIs/pull/88) so the three specs describe one proposer flow. Supersedes #625.

- `produceBlockV4` takes a required `BuilderConfig` body: a `builders` list with one `BuilderEntry` per builder-API bid request, plus a top-level `min_bid` and `builder_boost_factor` that apply to p2p bids. A bid is valued at `value + min(execution_payment, max_execution_payment)`, surviving bids are weighted by their `builder_boost_factor` and compared without division, and the local build's value is weighted by `100`.
- `submitBuilderPreferences` (`POST /eth/v1/validator/builder_preferences`) lets the validator client push `BuilderPreferencesEntry` objects, batched across its proposers, so the beacon node can submit each builder's `max_execution_payment` cap ahead of the slot.
- Block forwarding: `produceBlockV4` returns `Eth-Builder-Url` when a builder-API bid wins; `publishBlockV2` echoes it so the beacon node forwards the signed block to that builder via `submitSignedBeaconBlock`, letting it learn the result without waiting on gossip.
- Both request bodies support JSON and SSZ and require `Eth-Consensus-Version`; builder failures are handled per entry so one unusable builder does not fail the block-production request or prevent other preference submissions.

Reference implementation: [ChainSafe/lodestar#9832](https://github.com/ChainSafe/lodestar/pull/9832)


CHANGES.md
```diff
@@ -38,6 +38,7 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#608](https://github.com/ethereum/beacon-APIs/pull/608) `POST /eth/v1/validator/register_validator` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#608](https://github.com/ethereum/beacon-APIs/pull/608) `POST /eth/v1/validator/beacon_committee_subscriptions` updated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#630](https://github.com/ethereum/beacon-APIs/pull/630) `GET /eth/v4/validator/blocks/{slot}` changed to `POST /eth/v4/validator/blocks/{slot}`         |                                                  |                                                  |                                                    |                                                 |                                           |
```
nflaig: I would rather edit the existing changelog entry, ie. change `GET` to `POST` and add the PR to the list, similar to how I did it for #624

apis/validator/block.v4.yaml
```diff
@@ -127,9 +148,9 @@ get:
                 type: boolean
                 description: |
                   Indicates whether the execution payload envelope is included in the response.
-                  When `true`, the `data` field contains a `BlockContents` object with the
-                  beacon block, execution payload envelope, KZG proofs, and blobs. When `false`,
-                  the `data` field contains only a `BeaconBlock`.
+                  When `true`, the `data` field contains a `BlockContents` object with the beacon
+                  block, execution payload envelope, blobs, and KZG proofs. When `false`, the `data`
+                  field contains only a `BeaconBlock`.
```
nflaig: unintended diff?

types/gloas/builder_preference_entry.yaml
```diff
@@ -0,0 +1,47 @@
...
+          builder here unless `proxy` is set, and sends this value as the `Eth-Builder-Url` header
+          when routing through a proxy.
+        example: "https://builder.example.com"
+      proxy:
+        type: string
+        format: uri
+        description: |
+          An optional proxy URL. When set, the beacon node sends the request to this URL instead of
+          to `url`, and sets the `Eth-Builder-Url` header to `url` so the proxy can forward the
+          request to the intended builder.
```
nflaig: can you explain what's the use case of this, what's the purpose of the proxy in this setup and why can't the CL call the builder directly?

apis/validator/block.v4.yaml
```diff
@@ -45,59 +56,46 @@ get:
...
-        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+  requestBody:
+    required: false
+    description: |
+      An optional list of `BuilderEntry` objects, one per external builder to request a bid from,
+      added to the beacon node's local build and any p2p bid when selecting a payload. Each entry
+      names a concrete builder `url`, and no two entries may share the same `url`. For each entry the
+      beacon node contacts the builder at its `proxy` if set, otherwise at its `url`; forwards the
+      entry's `auth` unchanged; and applies the entry's `max_execution_payment`, `min_bid`, and
+      `builder_boost_factor` when comparing its bid. Malformed entries are ignored rather than failing
```
james-prysm: this says ignore malformed entries but submit builder preferences has a hard 400 error, maybe those should be ignored too instead?

apis/validator/block.v4.yaml
```diff
@@ -108,6 +106,29 @@ get:
           $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Consensus-Block-Value'
         Eth-Execution-Payload-Included:
           $ref: '../../beacon-node-oapi.yaml#/components/headers/Eth-Execution-Payload-Included'
+        Eth-Builder-Url:
+          description: |
+            For a block built on an external builder's bid, the `url` of the winning builder. The
```
james-prysm: should this be clarified that it's won via the builder api bid channel and not that an external builder won ? 

or not that it's present only when builder api is used

apis/validator/block.v4.yaml
JasonVranek: submitBuilderPreferences is now explicitly best-effort/non-atomic to match produceBlockV4

apis/validator/block.v4.yaml
JasonVranek: updated in f8d2943

types/gloas/builder_preference_entry.yaml
JasonVranek: updated the language in f8d2943

> An optional proxy URL for reaching this builder. Some deployments place a sidecar between
          the beacon node and its builders; when `proxy` is set the beacon node sends the request to
          this URL and sets the `Eth-Builder-Url` header to `url` so the sidecar can forward it to the
          intended builder. When `proxy` is unset the beacon node contacts the builder directly at
          `url`.

nflaig: that doesn't really answer my question, what I wanna know is what does the `proxy` do in addition to the CL client in gloas when it comes to the PBS pipeline and why should be enshrine the concept of such a `proxy` into the protocol?

JasonVranek: the plan post-gloas is for the BN to do muxing and bid selection. If we assume those are the only roles the builder-facing layer will ever need, the proxy is unnecessary, but I don't think that assumption holds. PBS itself came from out-of-protocol experimentation, not bc the protocol predicted it.

The BN exposes a fixed bid selection policy using min_bid, max_execution_payment, builder_boost_factor (at least in this PR). Out-of-protocol services may have different policy reqs beyond that set, e.g. (acknowledging my bias) commit-boost lets a proposer make commitments, so the winning bid must also be commitment-honoring, which the BN knobs can't express but a proxy in the request path can. 

These out-of-protocol experiments have a chicken and egg problem. To eventually become enshrined they have to run in the wild first, and if the only way to do that is forking a CL, it either kills the experimentation or incentivize forked clients. I'm advocating to keep the experimentation zone open (and ofc optional)

apis/beacon/blocks/blocks.v2.yaml
```diff
@@ -43,6 +43,30 @@ post:
       required: true
       name: Eth-Consensus-Version
       description: "The active consensus version to which the block being submitted belongs."
+    - name: Eth-Builder-Url
+      in: header
+      required: false
+      description: |
```
james-prysm, line 49: it's required false but maybe we should specify that we must echo this and eth-builder-proxy when available if returned by produceBlockv4 just to be explicit

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,82 @@
...
+              Optional builder BLS public key. When set, binds trusted execution payments to this
+              builder's on-chain key: bids not signed by it MUST NOT be accepted.
+      max_execution_payment:
+        allOf:
+          - $ref: "../primitive.yaml#/Gwei"
+          - description: |
+              Maximum trusted execution-layer payment, in Gwei, accepted from this builder. Bids
+              promising a trusted payment above this value MUST be rejected.
+            example: "1000000000"
+      min_bid:
```
james-prysm, line 55: min_bid and builder_boost_factor are defined per entry, but they feed a single global decision (which bid wins the slot). How have you interpreted for this

 Per-builder reading: each entry's values judge only that builder's bid. 
Proposer-level reading: they express one overall policy — but then entries can disagree and a collapse rule is needed (first? max?). 

hese pick different blocks from identical configs. 
```
┌───────────────────────────────────────────────┬───────────────────────────┬───────────────────────────┬────────────────────┐
│                Interpretation                 │         A's test          │         B's test          │       Winner       │
├───────────────────────────────────────────────┼───────────────────────────┼───────────────────────────┼────────────────────┤
│ Per-builder                                   │ 90 × 100% = 90 < 100, out │ 60 × 200% = 120 > 100, in │ B (a 60-value bid) │
├───────────────────────────────────────────────┼───────────────────────────┼───────────────────────────┼────────────────────┤
│ Proposer-level, collapse = first entry (100%) │ 90 < 100, out             │ 60 < 100, out             │ local              │
├───────────────────────────────────────────────┼───────────────────────────┼───────────────────────────┼────────────────────┤
│ Proposer-level, collapse = max (200%)         │ 180 > 100, in             │ 120 > 100, in             │ A (best raw bid)   │
└───────────────────────────────────────────────┴───────────────────────────┴───────────────────────────┴────────────────────┘
```

apis/validator/builder_preferences.yaml
```diff
@@ -0,0 +1,60 @@
...
+      application/json:
+        schema:
+          type: array
+          items:
+            $ref: "../../beacon-node-oapi.yaml#/components/schemas/Gloas.BuilderPreferenceEntry"
+  responses:
+    "200":
+      description: "The beacon node submitted the preferences to every builder."
+    "400":
+      description: "Errors with one or more preference submissions; the failing entries are identified by index. Well-formed entries are still submitted."
```
james-prysm: we should also say the duplicate thing here right?  or two entries sharing the same `url`

types/gloas/builder_entry.yaml
JasonVranek: my understanding was the per-builder reading. the [keymanager API describes the normative resolution order](https://github.com/ethereum/keymanager-APIs/pull/87/changes#diff-51f12a7e02d30ac4f311c36883ee2e7c9b64335e9ed7cc16f9a681542f64c79bR8) from BuilderEntry → BuilderConfig → default_config → client default per field so by the time we're at the bid selection process, the BN should have received all of the per-builder resolved preferences. So there shouldn't be a notion of a "global" min_bid at that point, but earlier upstream the VC could have effectively resolved a global min_bid from its default_config. still the comparisons should all be per-builder. in other words the 'collapse rule' isn't needed bc it should have collapsed upstream at the VC upon keymanager config



if that makes sense I can make that more explicit in the spec

james-prysm: sounds good interested what other cls think

apis/validator/block.v4.yaml
```diff
@@ -45,59 +56,48 @@ get:
...
+      beacon node contacts the builder at its `proxy` if set, otherwise at its `url`; forwards the
+      entry's `auth` unchanged; and applies the entry's `max_execution_payment`, `min_bid`, and
+      `builder_boost_factor` to that builder's own bid only. Each builder's bid is evaluated
+      independently under its own entry, and the highest-value candidate across those, the local
+      build, and any p2p bid wins; there is no cross-entry policy to reconcile. Malformed entries are
+      ignored rather than failing the request, so a single bad entry never blocks block production.
+      When the body is omitted or empty, or every entry is ignored, the beacon node contacts no
+      external builders and chooses between its local build and any p2p bid.
+    content:
+      application/json:
```
james-prysm, line 98: wait we should also allow ssz

types/gloas/builder_preference_entry.yaml
nflaig: > PBS itself came from out-of-protocol experimentation, not bc the protocol predicted it.

well I don't think mev-boost should even exist as a sidecar, CL clients are already 90% there in terms of functionality since the merge, all we needed to add is multiplexing and bid selection

> commit-boost lets a proposer make commitments

doesn't commit-boost need access to the bls keys of the proposer in that case anyways? so it could sign the auth request itself

but anyways, my concerns are addressed since we got rid of the concept of a `proxy`, while allowing such a "proxy" (whether that means relay or sidecar) to exist and work if operators wanna use it

apis/validator/block.v4.yaml
```diff
@@ -45,59 +57,57 @@ get:
...
-        of the `builder_boost_factor` do not trigger overflows or errors. If this parameter is
-        provided and no builder bid is available, the beacon node MUST respond with a block
-        committing to the local execution node payload, which the caller can choose to reject if
-        it wishes. If the value is provided but out of range for a 64-bit unsigned integer, then
-        an error response with status code 400 MUST be returned.
-      schema:
-        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+      The body MAY be encoded as JSON (`Content-Type: application/json`) or SSZ
+      (`Content-Type: application/octet-stream`); the `BuilderEntry` list is not fork-versioned, so
+      no `Eth-Consensus-Version` header is required.
```
nflaig: so we never plan to change `BuilderEntry` in the future?

also please make sure this somewhat follows how we document stuff in the beacon-api

eg. "MAY be encoded as JSON (`Content-Type: application/json`) or SSZ (`Content-Type: application/octet-stream`)", that seems completely unnecessary to point out, the spec already documents that and the `content:` section is enough

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,37 @@
...
+    returns only the `BeaconBlock`. The `Eth-Execution-Payload-Included` header and
+    `execution_payload_included` field report which form was returned. For a block built on an
+    external builder's bid the beacon node has no payload to return, and the `Eth-Builder-Url`
+    response header identifies the winning builder for the validator to echo when publishing.
+
+    The two outcomes publish differently and never overlap. For an external builder's bid, the
+    validator publishes the signed block via `publishBlockV2`, echoing the `Eth-Builder-Url` header,
+    and the winning builder releases the execution payload envelope. For a self-built block, the
+    validator publishes the signed block via `publishBlockV2` and the execution payload envelope
+    via `publishExecutionPayloadEnvelope`.
```
nflaig: this is quite verbose, the flow should be somewhat intuitive from a spec perspective, without having to overexplain it, really need to try and simplify the large text sections in this PR

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,37 @@
...
+    `max_execution_payment` (ceiling on the trusted execution payment). `builder_boost_factor` leans
+    the choice toward the local payload or a given builder without disabling either source.
 
-    When self-building (local execution payload), the response includes the full block contents
-    (beacon block, execution payload envelope, blobs, and KZG proofs) if `include_payload` is
-    set to `true`, otherwise only the `BeaconBlock` is returned.
-    When using an external builder bid, only the `BeaconBlock` is returned as the beacon node
-    does not have access to the builder's execution payload.
+    Post-Gloas, proposers submit execution payload bids rather than full execution payloads, so
+    there is no longer a blinded or unblinded distinction; builders release the payload later. This
```
nflaig: why does this mention anything about blinded or unblinded? sure it doesn't exist anymore but produce block v4 is post-gloas only

what's the intend behind pointing this out explicitly here?

apis/beacon/blocks/blocks.v2.yaml
```diff
@@ -43,6 +43,22 @@ post:
       required: true
       name: Eth-Consensus-Version
       description: "The active consensus version to which the block being submitted belongs."
+    - name: Eth-Builder-Url
```
nflaig, line 46: do we really need this header? if I understand correctly the sole purpose of this is so that the beacon node knows where to submit it via `submitSignedBeaconBlock` but the beacon node should know this anyways, since it got the bid from that builder via api

the only case where the beacon node doesn't know this is if it's a different node (eg. in a multi-node setup) if the primary node that sourced the bid from the builder via api went offline during the proposal flow but this seems like an edge case and even then, publishing only via p2p is a fine fallback behavior that is sufficient anyways most of the time

that said, we can keep the header if people wanna have it, just questionable to me if it's useful enough

types/gloas/builder_preference_entry.yaml
```diff
@@ -0,0 +1,50 @@
...
+      bid request, one entry per preference submission. Each entry names the builder `url` to
+      contact. For each entry the beacon node makes one `submitBuilderPreferences` builder-API
+      call: it contacts the builder at its `url`, forwards `auth` unchanged, and submits
+      `max_execution_payment`.
+
+      Unlike the block-production `BuilderEntry`, this carries only what a builder is allowed to
+      see. The proposer's private bid-filtering strategy (`min_bid` and `builder_boost_factor`) is
+      never sent to a builder; the beacon node applies those locally when selecting a bid.
+
+      SSZ request encoding. The `application/octet-stream` request body of the beacon-node
```
nflaig: this is standard in the beacon-api, repeating this over and over is not necessary, please try to follow existing conventions, this will also help trim these large text sections a bit and make it easier to review

apis/validator/block.v4.yaml
```diff
@@ -45,59 +57,57 @@ get:
...
+      the beacon node should make, added to the beacon node's local build and any p2p bid when
+      selecting a bid. Each entry names the builder `url` to contact; multiple entries MAY share a
+      `url`, and the beacon node makes one `getExecutionPayloadBid` call per entry. For each entry
+      the beacon node contacts the builder at its `url`; forwards the entry's `auth` unchanged; and
+      applies the entry's `max_execution_payment`, `min_bid`, and `builder_boost_factor` to that
+      builder's own bid only. Each builder's bid is evaluated independently under its own entry, and
+      the highest-value candidate across those, the local build, and any p2p bid wins; there is no
+      cross-entry policy to reconcile. Malformed entries are ignored rather than failing the
+      request, so a single bad entry never blocks block production. When the body is omitted or
+      empty, or every entry is ignored, the beacon node contacts no external builders and chooses
```
nflaig: > there is no cross-entry policy to reconcile

what does this mean?

> Malformed entries are ignored rather than failing the request

what does malformed mean in that case, I am guessing the format should still be per schema, especially in case of ssz that must be the case otherwise it fails deserialization

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,37 @@
...
   summary: "Produce a new block, without signature."
   description: |
-    Requests a beacon node to produce a valid block, which can then be signed by a validator.
+    Requests a beacon node to produce a valid block, which the validator then signs.
 
-    Post-Gloas, proposers submit execution payload bids rather than full execution payloads,
-    so there is no longer a concept of blinded or unblinded blocks. Builders release the
-    payload later. This endpoint is specific to the post-Gloas forks and is not backwards compatible
-    with previous forks.
+    The beacon node always builds a local payload and considers any p2p bid. The validator
```
nflaig: > The beacon node always builds a local payload and considers any p2p bid.

hmm I don't think this is the default behavior I want in lodestar, we have a option called `executiononly` on the validator client and I intend to respect that

I don't think the spec should be opinionated about this, there are valid reasons to not consider p2p bids, eg. a user that just prefers to build local, or for censorship resistance reasons

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,37 @@
...
+    optionally supplies a list of `BuilderEntry` objects in the request body, one per builder-API
+    `getExecutionPayloadBid` call to make; the beacon node adds those bids to the comparison and
+    returns the highest-value candidate. It biases each builder's bid by the entry's
+    `builder_boost_factor` and rejects a bid that falls outside the entry's `min_bid` (floor) or
+    `max_execution_payment` (ceiling on the trusted execution payment). `builder_boost_factor` leans
```
nflaig: isn't that documented in the builder-specs already? `max_execution_payment` shouldn't be documented here

apis/validator/block.v4.yaml
nflaig: ah realized we had this note before, hmm I don't really like it, but out of scope for this PR in that case

types/gloas/request_auth.yaml
```diff
@@ -0,0 +1,32 @@
+Gloas:
+  RequestAuthV1:
+    type: object
+    description: "The `RequestAuthV1` object from the [Builder API](https://ethereum.github.io/builder-specs/) Gloas spec."
```
nflaig: usually there is a proper reference to the object itself, at least when it comes to the consensus specs, I guess we kinda need a release of builder specs, could use master branch for now

apis/beacon/blocks/blocks.v2.yaml
```diff
@@ -5,16 +5,16 @@ post:
...
+    Instructs the beacon node to broadcast a newly signed beacon block to the beacon network, to be
+    included in the beacon chain. A success response (20x) indicates that the block passed gossip
+    validation and was successfully broadcast onto the network. The beacon node is also expected to
+    integrate the block into the state, but may broadcast it before doing so, so as to aid timely
+    delivery of the block. Should the block fail full validation, a separate success response code
+    (202) is used to indicate that the block was successfully broadcast but failed integration. For
+    Deneb/Electra/Fulu, this additionally instructs the beacon node to broadcast all given blobs.
+    For Gloas and later, blobs are broadcast as part of the `ExecutionPayloadEnvelope` and are not
+    submitted with the block. The broadcast behaviour may be adjusted via the `broadcast_validation`
+    query parameter.
```
nflaig: what's the diff here, it seems like just reformatting?

dictionary.dic
nflaig, line 1: why is this binary updated?

nflaig (commented):
did another quick pass, I feel like mostly this needs to be trimmed a lot, if details regarding bid selection are already part of the builder-specs there is no need to repeat that here, I don't really think the beacon-api should care about this at all, the builder-specs is not part of the required specs, so we should avoid leaking so much off protocol details into the spec here, especially when it comes to configs like `max_execution_payment`, min bid and boost factor we can consider keeping documented, similar to how we had it for `builder_boost_factor` query param

types/gloas/builder_preference_entry.yaml
```diff
@@ -0,0 +1,50 @@
...
+      `submitBuilderPreferences` endpoint is the SSZ serialization of
+      `List[BuilderPreferenceEntryV1, MAX_BUILDER_ENTRIES]`, using the constants defined on
+      `BuilderEntry` and the `SignedRequestAuthV1` container from the [Builder
+      API](https://ethereum.github.io/builder-specs/) Gloas spec:
+
+          class BuilderPreferenceEntryV1(Container):
+              url: ByteList[MAX_BUILDER_URL_SIZE]  # UTF-8 bytes of `url`, exactly as in JSON
+              auth: SignedRequestAuthV1
+              max_execution_payment: Gwei
```
nflaig: there is no need to re-define containers here, ideally we wanna add a reference to the spec container

apis/validator/builder_preferences.yaml
```diff
@@ -0,0 +1,67 @@
...
+      schema:
+        $ref: "../../types/primitive.yaml#/Pubkey"
+  tags:
+    - Validator
+  requestBody:
+    required: true
+    description: |
+      An array of `BuilderPreferenceEntry` objects, one per preference submission the beacon node
+      should make. Each entry names the builder `url` to contact; multiple entries MAY share a
+      `url`, and the beacon node makes one submission per entry. Inclusion is best-effort: the
```
nflaig: if the beacon node anyways makes one submission per entry, why do we need to submit on request per pubkey to the beacon node? we could add the validator pubkey to each entry and more closely align this api with `POST /eth/v1/validator/proposer_preferences`, so max items would be `64 * SLOTS_PER_EPOCH * 2`

might be missing something, would need to implement this to know what's easier, but seems simpler on paper if the validator client only has to send a single request

nflaig: even [submitBuilderPreferences](https://ethereum.github.io/builder-specs/?urls.primaryName=dev#/Builder/submitBuilderPreferences) should probably be reconsidered to allow an array and add pubkey to request body per entry. I know it's gonna be unlikely that this happens often on mainnet but implementation wise this seems simpler

apis/beacon/blocks/blocks.v2.yaml
JasonVranek: this was solely for the multi-node support, not opinionated on it but it seemed like DVT teams wanted. Will leave for now and address most of the fat trimming comments

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,70 @@
+Gloas:
+  BuilderEntry:
+    type: object
+    description: |
+      A per-builder input the validator client supplies on a block-production request, one entry per
+      builder-API bid request the beacon node should make. `auth` is the
+      [`SignedRequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#signedrequestauthv1) object from the Builder API Gloas spec.
```
nflaig: the previous `builder_boost_factor` query param also applied to p2p bids, how does this work using the `BuilderEntry` mechanism?

we should definitely have a way to apply `builder_boost_factor` (and `min_bid`) globally and at the same time it would be good if `BuilderEntry` can also be applied to p2p builders either by index or pubkey instead of url

beacon-node-oapi.yaml
```diff
@@ -566,3 +576,12 @@ components:
       required: true
       schema:
         type: boolean
+    Eth-Execution-Payload-Source:
+      description: |
+        Where this block's execution payload comes from: `local` when the beacon node built it,
+        `builder` when a builder-API bid won, and `p2p` when a p2p bid won. Required in response so
+        client can determine whether it must publish the execution payload envelope itself.
```
nflaig: > Required in response so client can determine whether it must publish the execution payload envelope itself.

this shouldn't be required, you can check `bid.builder_index === BUILDER_INDEX_SELF_BUILD` to figure out if it's self-build and whether or not you need to publish the execution payload envelope yourself

is there another egde case I am missing? generally for the validator client it shouldn't matter from where the builder bid is sourced

aside that, I kinda like the idea of this header, it's useful metadata, so I am not really opposed to adding it, could be a separate pr though, I don't really see how it's related to this one

JasonVranek: will remove from this PR

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,81 @@
...
+        allOf:
+          - $ref: "./request_auth.yaml#/Gloas/SignedRequestAuthV1"
+      builder_pubkey:
+        description: "The builder's BLS public key. On an entry with a `url` it filters the response: a builder-API bid not signed by it MUST NOT be accepted. On an entry without one it identifies the builder whose p2p bids this entry applies to."
+        allOf:
+          - $ref: "../primitive.yaml#/Pubkey"
+      max_execution_payment:
+        allOf:
+          - $ref: "../primitive.yaml#/Gwei"
+          - description: "Maximum trusted execution-layer payment, in Gwei, accepted from this builder. A bid promising more MUST be rejected."
```
ethDreamer: Another option for valuing a bid is simply:
```
total value = bid.value + min(bid.execution_payment, builder_entry.max_execution_payment)
```

So we wouldn't have to reject bids that offer more. This is entirely possible if the builder simply didn't receive our `BuilderPreferences`.

nflaig: this should be documented in the builder-specs though? I personally don't think the beacon-api should be aware of how to use `max_execution_payment` at all other than it being a field that exists and that it is passing around, and sure, the description is also fine to explain the field, but the rest should be part of builder-specs

we can keep in-protocol bid selection values, like min-bid or builder boost factor documented here, but `max_execution_payment` seems entirely off-protocol to me

JasonVranek: > this should be documented in the builder-specs though?

it's also described there but the BN needs to ultimately decide the winning bid. That means enforcing the `max_execution_payment` constraint on builder api bids. Otherwise it would need to be enforced at something like a sidecar level or the constraint itself is moot.

nflaig: > it's also described there but the BN needs to ultimately decide the winning bid

beacon nodes implement the builders-specs since The Merge, or I guess most of it, all the apis are implemented at least otherwise you can't even talk with mev-boost or other sidecars today. the point is that how we select off-protocol bids doesn't need to be defined in the `beacon-APIs` spec, that doesn't have any implication though on whether or not the beacon node does something, if what you are suggesting is true, we would have to move everything from the builder-specs that is not related to builders into this repo or the consensus-specs, but it's kept separately and this is by intend, the builder-specs are optional, while beacon-apis is not from a client implementers perspective

JasonVranek: I'll trim the `max_execution_payment` description to just what the field is and leave its selection semantics to the builder-specs

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,113 @@
...
+      `auth` is the [`SignedRequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#signedrequestauthv1) object from the Builder API Gloas spec.
+
+      A `BuilderEntry` never reaches a builder, so it has no Builder API definition. Its SSZ
+      container and constants are:
+
+          MAX_BUILDER_ENTRIES = 64
+          MAX_BUILDER_URL_SIZE = 2048
+
+          class BuilderEntryV1(Container):
+              url: ByteList[MAX_BUILDER_URL_SIZE]  # UTF-8 bytes of `url`, exactly as in JSON
```
nflaig, line 22: so in json the url would be hex-encoded? I guess this doesn't need to be human-readable so should be fine, I do think we should make sure to support ssz for everything that is transmitted over the wire from the validator --> beacon and from beacon --> builder

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,113 @@
...
+      min_bid:
+        allOf:
+          - $ref: "../primitive.yaml#/Gwei"
+          - description: "Minimum total payment, in Gwei, accepted from a bid that matches no entry, which in practice is a p2p bid. A bid whose `value` plus `execution_payment` is below this MUST be rejected."
+            example: "10000000"
+      builder_boost_factor:
+        allOf:
+          - $ref: "../primitive.yaml#/Uint64"
+          - description: "Percentage multiplier applied to a bid that matches no entry, exactly as a `BuilderEntry`'s `builder_boost_factor` is applied to a bid from that entry."
+            example: "100"
```
nflaig, line 109: so it should only be `min_bid` and `builder_boost_factor` that are applicable as global parameters. what I am wondering if it's in that case best to keep them here, or add them as query parameters, technically both works, and it might be nice to keep everything related to bid selection in one place, so we keep the query params for operational stuff only

apis/beacon/blocks/blocks.v2.yaml
nflaig: I do not feel strongly about removing it, also I think when I commented there was in addition a `Eth-Proxy-Url` header which is now gone, so this single header, if useful for DVT teams, might be worth keeping

types/gloas/builder_entry.yaml
JasonVranek: > Another option for valuing a bid is simply:
> 
> ```
> total value = bid.value + min(bid.execution_payment, builder_entry.max_execution_payment)
> ```
> 
> So we wouldn't have to reject bids that offer more. This is entirely possible if the builder simply didn't receive our `BuilderPreferences`.

Note for this, the logic/prose was removed from the beacon api. but the builder spec wording was relaxed to accomodate for the case where the builder failed to receive the preferences:

> Without stored preferences the builder MAY serve a bid with any `execution_payment`; the proposer's locally configured per-builder limits are the backstop: the proposer discards any bid that exceeds them.

apis/validator/builder_preferences.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+    This API is applicable from the Gloas fork onwards.
+  parameters:
+    - name: pubkey
+      in: path
+      required: true
+      description: |
+        The BLS public key of the proposing validator whose preferences these are. The beacon node
+        passes it as the `proposer_pubkey` of each builder-API preference submission, and each
+        entry's `auth` is signed by this key.
```
nflaig: I left this comment before but can we please fix this api, I don't see a good reason why we would do a single call per pubkey. I don't believe there is a single beacon api that works like this, we generally do batching cc @rolfyone @james-prysm @tersec or anyone else that could have an opinion on this too

@JasonVranek can you please do the same on the builder-api, or check why we went with that api design, and also double check with builders

apis/validator/builder_preferences.yaml
```diff
@@ -0,0 +1,56 @@
...
+        $ref: "../../types/primitive.yaml#/Pubkey"
+  tags:
+    - Validator
+  requestBody:
+    required: true
+    description: |
+      The `BuilderPreferenceEntry` objects to submit. Inclusion is best-effort: the beacon node
+      submits each entry independently and skips any entry it cannot submit rather than failing the
+      whole request. Entries that were not submitted are reported by index in a 400 response.
+    content:
```
nflaig, line 30: this api is missing fork versioning, wouldn't it be good to keep the door open to easily update `BuilderPreferenceEntryV1` in heze without having to bump the api? it's pretty painful to introduce a new api version outside of a hard fork, so might as well version via fork, especially a api like this

how to deal with this on the builder-api side can be different, but I would just add a fork version to leave the door open to add a new container in heze, there isn't much harm to add this metadata

apis/validator/builder_preferences.yaml
JasonVranek: since gloas gossips proposer preferences instead of flooding relays with validator registrations, and builder preferences are only sent for the proposers in the lookahead, this is realistically only an optimization for operators large enough to have >1 proposer in the lookahead, and even then it just saves a couple of http calls outside the hot path.

I'm happy to update the beacon API to batch and match existing conventions (will start on that now). I'd push back on changing the builder spec though. The BN already has to fan out per-builder to request bids, so batching preferences there doesn't save it any mechanism it isn't already building. also it makes error handing more confusing at VC since there'd be two levels of partial failures (some entries rejected at the BN, then some at each builder)

nflaig: >  this is realistically only an optimization for operators large enough to have >1 proposer in the lookahead

this is not about a performance optimization, it's about having a good api design, yes it's unlikely on mainnet that this is gonna be >1 but the validator client still needs to support this scenario, but this opens up several implementation details, do you send the request in sequence or parallel? if in sequence, if there is a http error, should you stop sending or keep going?

> The BN already has to fan out per-builder to request bids

yes this is already bad enough, why do we want to make it per-builder per-pubkey?

> also it makes error handing more confusing at VC since there'd be two levels of partial failures (some entries rejected at the BN, then some at each builder)

imo the error handling is strictly better, we already have an established pattern for this, see `IndexedErrorMessage` in the beacon-api spec, we use this for consensus critical apis like [submitPoolAttestationsV2](https://ethereum.github.io/beacon-APIs/#/Beacon/submitPoolAttestationsV2) and it works perfectly there

> some entries rejected at the BN

I need to read up on this still, you might have documented that already, but what are rejections/error that can happen on the beacon node side? I would expect the beacon node to act as more of a proxy here, and then report back the failures/errors it gets from the builder requests, and only then send the response 200/400 back to the validator client, othwerise you have 0 visiblity from the vc side on what's going on

but generally, I don't really see what kind of error can happen here, the individual fields are defined in the schema and there shouldn't be many ways to produce a invalid request?

the failure cause I can see is a 401 if the authentication fails, but would you rather have one 401 from the builder or a 401 per-pubkey? I can see the argument here if pubkeys sign over different `auth_data`? maybe that is an argument to keep one request per pubkey on builder-specs?

but considering we already send batches to builders via `registerValidator` and it's working since ~4 years I don't see why that is a problem

I would probably just go with what builders/relays prefer here, they have to implement the server-side of this, I wouldn't care to much to add a for-loop on the beacon node side if they much prefer per-pubkey

nflaig: > the failure cause I can see is a 401 if the authentication fails, but would you rather have one 401 from the builder or a 401 per-pubkey? I can see the argument here if pubkeys sign over different auth_data? maybe that is an argument to keep one request per pubkey on builder-specs?

answering my own question, I think on the beacon node --> builder request side it makes sense to send per-pubkey due to 401 http errors, added more details here  https://github.com/ethereum/builder-specs/pull/165

JasonVranek: sounds good, in the process of drafting:
- VC -> BN batching for preferences
- dropping *_V1 across the PRs in favor of fork-versioned types

note a byproduct of VC->BN batching is moving `pubkey` from the VC->BN path to the request body, so to simplify/share types I'm mirroring the change on the builder spec. So a `BuilderPreferencesRequest` gets a `pubkey` field and the `submitBuilderPreferences` builder api endpoint drops the `proposer_pubkey` path param. This is also a step towards enabling BN->builder batching but intentionally holding off on that change for now

nflaig: > note a byproduct of VC->BN batching is moving pubkey from the VC->BN path to the request body,

I don't understand this, aren't the containers different already? why does how we submit it from vc --> bn have any impact on bn --> builder?

I had codex quickly give me the shape I am thinking of right now

image

image

image


doesn't this work nicely on both ends?

apis/validator/builder_preferences.yaml
nflaig: > dropping *_V1 across the PRs in favor of fork-versioned types

@JasonVranek let's discussion the versioning here, so from my experience, using a `V1` suffix isn't useful, it implies you need a new api version to change it? or how else does this work?

in my opinion and this has worked quite nicely so far is to use the `Eth-Consensus-Version` header, this allows pretty seamless updates to the containers across forks to meet new requirements, and notably, this doesn't prevent you from doing a v2 mid-fork, although, like I said, it's pretty hard to roll this out mid-fork, rolling out just ssz which is supposed to be backward compatible was a huge pain... so really even if you do a v2 api, you can mostly just rely on it in the next fork once all parties are forced to update their software

apis/validator/builder_preferences.yaml
nflaig: although it might be nice to match the containers more, like why is `max_execution_payment` a top-level field on the beacon api and nested inside `preferences.max_execution_payment` on the builder-api?

JasonVranek: this works. I was considering reusing the `BuilderEntry` struct supplied in `produceBlockV4` in the preferences endpoint to avoid defining a new intermediate type like `BuilderPreferenceEntry`. Idea was from the `BuilderEntry` you could create the `BuilderPreferencesRequest` the builder api needs. Also the builder api doesn't allow you to send things like `min_bid` so we structurally don't have to worry about leaking the private proposer config values. But bc of the ssz not allowing optional fields it actually turned out a bit messier/uglier than just defining the intermediate type. 

> why does how we submit it from vc --> bn have any impact on bn --> builder

yea not going to end up doing this. Will keep `BuilderPreferencesRequest` as is (except for the fork versioning change)

apis/validator/builder_preferences.yaml
JasonVranek: > in my opinion and this has worked quite nicely so far is to use the Eth-Consensus-Version header

yes I believe I'm doing this exactly, just to verify:
- `BuilderPreferencesRequestV1` -> `BuilderPreferencesRequest`
- `Eth-Consensus-Version` optional -> required

apis/validator/builder_preferences.yaml
nflaig: I think it's fine to keep containers separate, maybe even desired. also the screenshots above, we probably wanna rename `pubkey` to `validator_pubkey` to avoid potential confusing with `builder_pubkey`

JasonVranek: > we probably wanna rename pubkey to validator_pubkey to avoid potential confusing with builder_pubkey 

going with `proposer_pubkey` to match the builder api's path params

nflaig: > going with proposer_pubkey to match the builder api's path params

I would need to double check, it's sounds bit more unfamiliar to me than `validator_pubkey` but for the purpose of differentiating it works fine, so no strong preference

apis/validator/builder_preferences.yaml
nflaig: yes sounds good to me, having `Eth-Consensus-Version` optional anywhere should be considered a bug

apis/validator/builder_preferences.yaml
JasonVranek: The changes have been pushed across the 3 PRs:
- bb1e790 adds batched submitBuilderPreferences beacon api calls
- https://github.com/ethereum/builder-specs/pull/165/commits/36cefe4aa82b562e1cfdde2be0dba7b5b8b93254 adopts fork-versioned types
- https://github.com/ethereum/keymanager-APIs/pull/88/commits/655d731c222657c25dd2719d030481b3d07eb061 is a minor diff to reference the fork-versioned type
- this is the version of the [implementer guide](https://github.com/JasonVranek/epbs-config-flow/blob/360cbfdc8aaabebf3c58c3a09d7fe2590a7ae47f/builder-config-flow.md) pinned to these commits

types/gloas/builder_entry.yaml
ethDreamer: There's an important nuance here..

  > by the time we're at the bid selection process, the BN should have received all of the per-builder resolved preferences … the comparisons should all be per-builder         

  That holds for bids the BN pulls directly via `getExecutionPayloadBid` — each arrives together with its entry, so the BN can rank by the per-builder *boosted* value. But I don't think it holds for p2p (gossip) bids, and it's the table's headline example that breaks.

  Gossip bids are pre-filtered by the [consensus-specs gossip rule](https://github.com/ethereum/consensus-specs/blob/master/specs/gloas/p2p-interface.md#new-execution_payload_bid)

  > _[IGNORE]_ this bid is the highest value bid seen for the tuple `(bid.slot, bid.parent_block_hash, bid.parent_block_root)`

  So a node retains only the single highest-`value` bid per tuple; lower bids are `IGNORE`d at validation and never cached. Now re-read the per-builder example assuming A and B are gossip bids for the same tuple:

  |           | A    | B          |
  | --------- | ---: | ---------: |
  | raw value | 90   | 60         |
  | boost     | 100% | 200%       |
  | boosted   | 90   | 120 → wins |

  The table says B (the 60-value bid) wins. But under the gossip rule, B was `IGNORE`d the moment A (value 90) was seen for that tuple — it's never in the cache, so per-builder boost (which is only known later at proposal time) can't resurrect it. In other words, `builder_boost_factor` and `min_bid` can't reorder among gossip bids, because the gossip layer has already collapsed them to one before selection runs.

potuz: Yes the boost is kinda irrelevant for p2p bids since they are filtered by nodes that don't know the proposer's value.

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,113 @@
...
+        maxItems: 64  # MAX_BUILDER_ENTRIES
+        description: "The builders to request bids from, one `BuilderEntry` each. Empty means request none, so only p2p bids are considered."
+        items:
+          $ref: "#/Gloas/BuilderEntry"
+      min_bid:
+        allOf:
+          - $ref: "../primitive.yaml#/Gwei"
+          - description: "Minimum total payment, in Gwei, accepted from a bid that matches no entry, which in practice is a p2p bid. A bid whose `value` plus `execution_payment` is below this MUST be rejected."
+            example: "10000000"
+      builder_boost_factor:
```
nflaig, line 100: same nit here, I would prefer to have global fields at the top of the container

apis/validator/block.v4.yaml
```diff
@@ -45,59 +65,47 @@ get:
...
+        `GET /eth/v1/validator/execution_payload_envelopes/{slot}/{beacon_block_root}`, signs it, and
+        publishes it via `POST /eth/v1/beacon/execution_payload_envelopes`, and the producing beacon
+        node attaches its cached blobs and KZG proofs before broadcasting. This sends fewer bytes but
+        ties publishing to that one beacon node (stateful operation).
 
-        This parameter only affects self-building scenarios. When using an external builder's bid,
-        only the beacon block is returned regardless of this parameter (the beacon node does not
-        have access to the builder's execution payload).
+        This parameter affects the self-built case only. When a bid wins, the beacon node does not
+        hold that payload, so it returns only the beacon block regardless of this parameter.
```
nflaig: what exactly changed here?

> This parameter affects the self-built case only. When a bid wins, the beacon node does not
        hold that payload, so it returns only the beacon block regardless of this parameter.

this note seems useful

but also previously there was
> included in the response when self-building

so if block includes a builder bid, this flag is ignored anyways

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,113 @@
...
+          - $ref: "../primitive.yaml#/Gwei"
+          - description: "Minimum total payment, in Gwei, accepted from this builder. A bid whose `value` plus `execution_payment` is below this MUST be rejected."
+            example: "10000000"
+      builder_boost_factor:
+        allOf:
+          - $ref: "../primitive.yaml#/Uint64"
+          - description: |
+              Percentage multiplier applied to this builder's bid once it has passed the
+              `min_bid` check, giving
+              `builder_boost_factor * (builder_payload_value // 100)`, where
```
nflaig: hmm can you please revisit this, I did some cleanup here https://github.com/ethereum/beacon-APIs/pull/624/changes#diff-0bfdd4898edd7ab2e4cdefa66264cd4d0c4467ab5ff6cff00a3a8296b4274cbc, now in the pr we revert that, `builder_payload_value` is just the wrong term imo, can you please make sure this is simiarl to what we have on `master` branch, of course with the caveat that it's per-builder here so wording might be different

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,35 @@
...
   description: |
-    Requests a beacon node to produce a valid block, which can then be signed by a validator.
+    Requests a beacon node to produce a valid block, which the validator then signs.
 
-    Post-Gloas, proposers submit execution payload bids rather than full execution payloads,
-    so there is no longer a concept of blinded or unblinded blocks. Builders release the
-    payload later. This endpoint is specific to the post-Gloas forks and is not backwards compatible
-    with previous forks.
+    The beacon node always builds a local payload and MAY consider a p2p bid, so a block is returned
+    even when no bid is available. The validator client optionally supplies a `BuilderConfig` in the
```
eserilev: nit: there are several places where we mention bid (i.e. here "no bid is available"). in the local build case a bid is still constructed. It might be useful instead to specify builder bid, for example 

```suggestion
    even when no builder bid is available. The validator client optionally supplies a `BuilderConfig` in the
```

there are multiple places in the PR where "bid" is assumed to be a builder bid. it might make things a bit clearer if we just explicitly say builder bid in those cases

nflaig: not disagreeing with adding "builder" here as suggested but referring to local payload as "local bid" or similar is really confusing, during block proposal you either use a bid from a builder or you commit to a local payload, the fact that the `ExecutionPayloadBid` is overloaded and used for both is mostly spec limitation, ideally this should be something like `Union[ExecutionPayloadBid, ExecutionPayloadCommitment]` where the commitment type used when self-building drops anything that is unnecessary for local payloads like `builder_index`, `fee_recipient` and `value`

eserilev: yeah i totally agree with not using the term local bid

apis/validator/block.v4.yaml
```diff
@@ -58,46 +68,30 @@ get:
...
-        Servers should use saturating arithmetic or another technique to ensure that large values
-        of the `builder_boost_factor` do not trigger overflows or errors. If this parameter is
-        provided and no builder bid is available, the beacon node MUST respond with a block
-        committing to the local execution node payload, which the caller can choose to reject if
-        it wishes. If the value is provided but out of range for a 64-bit unsigned integer, then
-        an error response with status code 400 MUST be returned.
-      schema:
-        $ref: "../../beacon-node-oapi.yaml#/components/schemas/Uint64"
+  requestBody:
+    required: false
```
nflaig: while this request body is not required, we should make the `Eth-Consensus-Version` header required, it's completely missing from the request right now

arguably it's only needed if `BuilderConfig` is passed in the request body, but I would make the header always required

also worth considering making passing `BuilderConfig` since it also applies to p2p bids, I don't see a good reason to omit this?

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,107 @@
...
+          class BuilderEntry(Container):
+              url: ByteList[MAX_BUILDER_URL_SIZE]  # UTF-8 bytes of `url`, exactly as in JSON
+              auth: SignedRequestAuth
+              builder_pubkey: BLSPubkey
+              max_execution_payment: Gwei
+              min_bid: Gwei
+              builder_boost_factor: uint64
+
+      SSZ cannot express absence, so an unset `builder_pubkey` is all-zero, which is not a valid
+      BLS public key. A zero-length `url` is equivalent to omitting it, so a body containing one is
```
nflaig: url is required now? so the comment here seems outdated

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,107 @@
...
+        format: uri
+        maxLength: 2048  # MAX_BUILDER_URL_SIZE
+        description: "Where this entry's bid request is sent."
+        example: "https://builder.example.com"
+      auth:
+        description: "Authenticates this entry's bid request to the builder. The beacon node MUST forward `message` and `signature` byte-for-byte unchanged."
+        allOf:
+          - $ref: "./request_auth.yaml#/Gloas/SignedRequestAuth"
+      builder_pubkey:
+        description: "The builder's BLS public key. When present it filters the response: a builder-API bid not signed by it MUST NOT be accepted."
```
nflaig: this should be required in json too so that ssz/json has same shape

JasonVranek:
Batch update across the three PRs, summarizing the recent review threads. Beacon-API changes are in eb752a1 through 71c7315, [ethereum/keymanager-APIs#88](https://github.com/ethereum/keymanager-APIs/issues/88) (d5b54fc through be0f46e), [ethereum/builder-specs#165](https://github.com/ethereum/builder-specs/issues/165) (9a58e07 through a5552d6).

- The `BuilderConfig` body and the `Eth-Consensus-Version` header are now required in `produceBlockV4`
- `builder_pubkey` is now `builder_pubkeys`, a `List[BLSPubkey, MAX_BUILDER_PUBKEYS]`. Empty list is the old behavior. Fixes the optional SSZ field ugliness and improves expressibility.
- A zero-length auth data is now invalid across all three repos, closing the last SSZ absence ambiguity
- The `enabled` flag is removed from keymanager api
- Builder preferences explicitly stored per proposer per `auth.message.slot` in builder spec
- Eth-Consensus-Version is required on every request that carries a body; absent or unrecognized is a 400. 
- Added the missing 415 on `submitBuilderPreferences`
- The implementer guide ([JasonVranek/epbs-config-flow#1](https://github.com/JasonVranek/epbs-config-flow/issues/1)) is updated to match.

cc @nflaig @james-prysm

apis/validator/block.v4.yaml
```diff
@@ -1,25 +1,37 @@
...
 
-    When self-building (local execution payload), the response includes the full block contents
-    (beacon block, execution payload envelope, blobs, and KZG proofs) if `include_payload` is
-    set to `true`, otherwise only the `BeaconBlock` is returned.
-    When using an external builder bid, only the `BeaconBlock` is returned as the beacon node
-    does not have access to the builder's execution payload.
+    A builder bid, over the builder API or p2p, is rejected if its total payment falls below the
+    `min_bid` that applies to it. Every surviving bid is then boosted by the
+    `builder_boost_factor` that applies to it, and the
+    highest boosted bid competes with the local build in Gwei. The local build wins a tie.
```
nflaig, line 21: maybe this can be clarified on the spec more but at least on the implementation side need to be careful with Gwei vs. Wei

today (pre-gloas), we use Wei almost everywhere
- `execution_payload_value` Wei
- `consensus_block_value` Wei
- `blockValue` returned from EL Wei

but now in gloas we use Gwei for most
- `bid.value` Gwei
- `bid.execution_payment` Gwei
- `min_bid` Gwei

so double checked this in lodestar we do the unit conversion correctly for p2p bids already to compare vs. local payload value but this is something to easily get wrong, should find it during testing though

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,115 @@
...
+    type: object
+    description: |
+      A per-builder input the validator client supplies on a block-production request.
+
+      Each entry is a bid request sent to its `url`, authenticated by its `auth`, which the
+      builder requires on every request. No two entries may share both their `url` and their
+      `auth`'s `data`, and one request is made per entry, so several MAY share a `url` with
+      different `data`.
+
+      `auth` is the [`SignedRequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#signedrequestauthv1) object from the Builder API Gloas spec.
```
nflaig: this seems stale, we renamed `SignedRequestAuthV1` to `SignedRequestAuth`, the commit hash is outdated

types/gloas/builder_preferences_entry.yaml
```diff
@@ -0,0 +1,42 @@
...
+    description: |
+      A per-builder preference the validator client asks the beacon node to submit ahead of the bid
+      request, one entry per builder per proposer. A single submission batches these across the
+      validator client's proposers, so each entry names the proposer it belongs to in
+      `proposer_pubkey`. It is a dedicated type, distinct from the block-production `BuilderEntry`.
+
+      The entry pairs the beacon node's routing with the payload the builder receives, so like
+      `BuilderEntry` it has no Builder API definition. Its SSZ container uses that type's constants
+      and the
+      [`SignedRequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#signedrequestauthv1) object from the Builder API Gloas spec:
```
nflaig: stale

types/gloas/request_auth.yaml
```diff
@@ -0,0 +1,32 @@
+Gloas:
+  RequestAuth:
+    type: object
+    description: "The [`RequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#requestauthv1) object from the Builder API Gloas spec."
```
nflaig: stale

types/gloas/request_auth.yaml
```diff
@@ -0,0 +1,32 @@
...
+        example: "0x1234567890abcdef"
+        pattern: "^0x(?:[a-fA-F0-9]{2}){1,4096}$"
+      slot:
+        allOf:
+          - $ref: "../primitive.yaml#/Uint64"
+          - description: "The proposal slot this request is authorized for."
+
+  SignedRequestAuth:
+    type: object
+    description: "The [`SignedRequestAuthV1`](https://github.com/ethereum/builder-specs/blob/78a5546d9d8253beabf7db8baf988a58abdec87f/specs/gloas/validator.md#signedrequestauthv1) object from the Builder API Gloas spec."
```
nflaig: stale

nflaig (commented):
there's a lot of reference to builder-specs which are now outdated, later, we should probably replace the commit hashes with a tag once we have a release

types/gloas/request_auth.yaml
JasonVranek: made all 4 refs pin to master. note it's fork scoped to `..../types/gloas/*` so can potentially get away with not pinning to a specific commit hash.

nflaig: will pin these later to a tag once we have builder-specs release

apis/beacon/blocks/blocks.v2.yaml
eth2353: +1 questioning the usefulness of this header, both here and in the produceBlockV4 response - can someone explain how this would be used?

eth2353: +1 questioning the usefulness of this header, both here and in the produceBlockV4 response. I'd personally prefer removing it unless there is a good reason to keep it.

nflaig: after implementing it, the header makes it easier on the submission side on the beacon node, avoids having to implement a lookup cache to "remember" to which builder to submit the bid to, it also acts as a signal whether the beacon node should forward the beacon block to a builder

dawidn19847-pixel (changes requested):
[](url)

StefanBratanov (approved):

james-prysm (dismissed):
ship it

KaloyanTanev (approved):
All good on Obol's side, as long as a beacon node accepts a submission of a signed block with a builder URL different than what it provided as an unsigned payload earlier. And of course, utilises this new URL it has received upon submission.

types/gloas/request_auth.yaml
```diff
@@ -0,0 +1,32 @@
+Gloas:
+  RequestAuth:
```
nflaig: same renaming from https://github.com/ethereum/builder-specs/pull/165#discussion_r3836969633 would apply here as well

I have https://github.com/ethereum/keymanager-APIs/pull/92 which would cover the keymanager api (just prose text there)

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,116 @@
...
+              favors the local payload and one above `100` favors this builder.
+
+              Servers must support the following values, which encode common preferences:
+
+              * `builder_boost_factor=0`: prefer the local payload unless an error makes it unviable.
+              * `builder_boost_factor=100`: profit maximization mode; choose whichever pays more.
+              * `builder_boost_factor=2**64 - 1`: prefer this builder unless an error or health check
+                makes it unviable.
+
+              Servers should use saturating arithmetic so that large values do not overflow.
```
nflaig: if we wanna get rid of the division to preserve full precision per comment from @ethDreamer on discord

```suggestion
              Percentage multiplier applied to this builder's bid once it has passed the
              `min_bid` check. Its weighted value is
              `builder_boost_factor * builder_bid_value`, where
              `builder_bid_value` is the bid's `value` plus `min(execution_payment,
              max_execution_payment)`. Compare bids by their weighted values without division.
              When comparing against the local payload, convert both to the same unit and weight
              the local value by `100`; the local payload wins an exact tie.

              Servers must support the following values, which encode common preferences:

              * `builder_boost_factor=0`: prefer the local payload unless an error makes it unviable.
              * `builder_boost_factor=100`: profit maximization mode; choose whichever pays more.
              * `builder_boost_factor=2**64 - 1`: prefer this builder unless an error or health check
                makes it unviable.

              Servers should compute weighted values without overflow or loss of precision.
```

apis/beacon/blocks/blocks.v2.yaml
```diff
@@ -43,6 +43,20 @@ post:
...
+    - name: Eth-Builder-Url
+      in: header
+      required: false
+      description: |
+        The `url` of the winning builder, as returned by `POST /eth/v4/validator/blocks/{slot}`.
+        When that endpoint returns this header the validator client MUST echo it here, so the beacon
+        node can forward the signed block to the same builder and the builder can release the
+        payload without waiting for gossip. The echo is what lets a beacon node that did not serve
+        the block-production request still forward it. Omitted for a self-built block or a block won
+        by a p2p bid.
```
nflaig, line 55: this can be implementation specific but we should probably not blindly call any url that is provided here, otherwise this creates a potential SSRF path

this isn't that critical since generally the beacon api is not to be exposed publicly but there are providers like rescue node or in general rpc nodes that could be vulnerable to this if not handled correctly

cc @jshufro might be relevant for you

wordlist.txt
```diff
@@ -47,3 +47,11 @@ produceBlockV
...
 CGC
 RLP
+auth
+RequestAuth
+RequestAuthV
+SignedRequestAuth
+SignedRequestAuthV
+UTF
+uri
+publishBlockV
```
nflaig:
```suggestion
  UTF
  uri
  ```

File: wordlist.txt [hunk 45]
```diff
@@ -47,3 +47,5 @@ produceBlockV
 stateful
 CGC
 RLP
+ UTF
+ uri
```
nflaig:
```suggestion
UTF
uri
```

apis/beacon/blocks/blocks.v2.yaml
```diff
@@ -43,6 +43,26 @@ post:
...
+
+        The URL is untrusted input and a server-side request forgery (SSRF) risk: the beacon
```
nflaig, line 57: as a side note, lodestar will only forward urls that have previously been authenticated by a proposer, I expect even in DVT setups that either `produceBlockV4` or `submitBuilderPreferences` has authenticated the url beforehand on all beacon nodes

types/gloas/builder_preferences_entry.yaml
```diff
@@ -0,0 +1,42 @@
...
+    description: |
+      A per-builder preference the validator client asks the beacon node to submit ahead of the bid
+      request, one entry per builder per proposer. A single submission batches these across the
+      validator client's proposers, so each entry names the proposer it belongs to in
+      `proposer_pubkey`. It is a dedicated type, distinct from the block-production `BuilderEntry`.
+
+      The entry pairs the beacon node's routing with the payload the builder receives, so like
+      `BuilderEntry` it has no Builder API definition. Its SSZ container uses that type's constants
+      and the
+      [`SignedBuilderRequestAuth`](https://github.com/ethereum/builder-specs/blob/main/specs/gloas/validator.md#signedbuilderrequestauth) object from the Builder API Gloas spec:
```
nflaig: we have the reference already in the object


```suggestion
      and the `SignedBuilderRequestAuth` object from the Builder API Gloas spec:
```

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,116 @@
...
+    type: object
+    description: |
+      A per-builder input the validator client supplies on a block-production request.
+
+      Each entry is a bid request sent to its `url`, authenticated by its `auth`, which the
+      builder requires on every request. No two entries may share both their `url` and their
+      `auth`'s `data`, and one request is made per entry, so several MAY share a `url` with
+      different `data`.
+
+      `auth` is the [`SignedBuilderRequestAuth`](https://github.com/ethereum/builder-specs/blob/main/specs/gloas/validator.md#signedbuilderrequestauth) object from the Builder API Gloas spec.
```
nflaig:
```suggestion
      `auth` is the `SignedBuilderRequestAuth` object from the Builder API Gloas spec.
```

types/gloas/builder_entry.yaml
```diff
@@ -0,0 +1,116 @@
...
+          - $ref: "../primitive.yaml#/Uint64"
+          - description: |
+              Percentage multiplier applied to this builder's bid once it has passed the
+              `min_bid` check. Its weighted value is
+              `builder_boost_factor * builder_bid_value`, where
+              `builder_bid_value` is the bid's `value` plus `min(execution_payment,
+              max_execution_payment)`, saturating at `2**64 - 1`, so the weighted value fits in
+              128 bits. Bids are compared by weighted value; against the local payload, the local
+              value is weighted by `100`, so a factor below `100` favors the local payload and one
+              above `100` favors this builder.
```
nflaig:
```suggestion
              Percentage multiplier applied to this builder's bid once it has passed the `min_bid`
              check. Its weighted value is `builder_boost_factor * builder_bid_value`, where
              `builder_bid_value` is the bid's `value` plus `min(execution_payment, max_execution_payment)`.
              The sum saturates at `2**64 - 1`, so the weighted value fits in 128 bits.

              Bids are compared by weighted value. Against the local payload, the local value is
              weighted by `100`, so a factor below `100` favors the local payload and one above
              `100` favors this builder.
```

nflaig (approved):
LGTM

Bettdatsh210-boop (commented):
Approve

PR #631, opened 2026-07-25
Add execution_payload_value to produceBlockV4

nflaig:
`produceBlockV3` returns `execution_payload_value` but `produceBlockV4` dropped it, which seems like an oversight/regression.

There are validator clients (like vero/vouch) that query multiple beacon nodes and compare `consensus_block_value` + `execution_payload_value` to pick the most valuable block. For self-builds the value is otherwise unavailable since the bid value is zero.

This PR re-adds the field and the `Eth-Execution-Payload-Value` response header to restore parity with v3.


apis/validator/block.v4.yaml
```diff
@@ -123,6 +125,12 @@ get:
                 type: string
                 example: "12345"
                 description: "Consensus rewards for this block in Wei"
+              execution_payload_value:
+                type: string
+                example: "12345"
+                description: |
+                  Execution payload value in Wei. The value of the local execution payload when
+                  self-building, or the bid value when committing to a builder bid.
```
eth2353: Using the term bid value here leads me to think it may be equal to `bid.value`. Whereas I would expect it to equal `bid.value + bid.execution_payment`.

nflaig: this seems fine to me, I leave it up to implementations to interpret what "bid value" means, if a client doesn't allow trusted bids it's simply the `bid.value`

eth2353:
```suggestion
                  Execution payload value in Wei. The value of the local execution payload when
                  self-building, or the total value of the bid when committing to a builder bid.
```

I'd suggest rewording it slightly then, since the current wording makes it very easy to interpret this simply as `bid.value` which is wrong in case there is an `execution_payment` present in the bid as well.

eth2353 (approved):

rolfyone (approved):

eserilev (approved):
lgtm

james-prysm (approved):

PR #632, opened 2026-08-06
Add a plugin to make SSZ request body description visible

syjn99:
This is sort of hacky way to render a description for SSZ request body via Swagger UI. This covers all 9 octet-stream request bodies at once:
- `POST /eth/v2/beacon/blinded_blocks`
- `POST /eth/v2/beacon/blocks`
- `POST /eth/v1/beacon/execution_payload_bids`
- `POST /eth/v1/beacon/execution_payload_envelopes`
- `POST /eth/v2/beacon/pool/attestations`
- `POST /eth/v1/beacon/pool/payload_attestations`
- `POST /eth/v2/validator/aggregate_and_proofs`
- `POST /eth/v1/validator/register_validator`
- `POST /eth/v1/validator/proposer_preferences`

Example:

image

Today it just shows like

```
Example values are not available for application/octet-stream media types.
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -44,6 +44,30 @@
     <script src="./dist/swagger-ui-bundle.js"> </script>
     <script src="./dist/swagger-ui-standalone-preset.js"> </script>
     <script>
+    // A plugin to render the schema description for binary media types in the RequestBody component.
+    // rather than showing "Example values are not available for <type> media types."
+    const RequestBodySchemaDescription = () => ({
+      wrapComponents: {
+        RequestBody: (Original, system) => (props) => {
+          const React = system.React
+          const contentType = props.contentType || ""
+          const isBinaryMediaType = ["application/octet-stream", "image/", "audio/", "video/"]
+            .some((mediaType) => contentType.startsWith(mediaType))
+          const description = isBinaryMediaType && props.requestBody
+            ? props.requestBody.getIn(["content", contentType, "schema", "description"])
+            : null
+          if (!description) {
+            return React.createElement(Original, props)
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -44,6 +44,30 @@
...
     <script src="./dist/swagger-ui-standalone-preset.js"> </script>
     <script>
+    // A plugin to render the schema description for binary media types in the RequestBody component.
+    // rather than showing "Example values are not available for <type> media types."
+    const RequestBodySchemaDescription = () => ({
+      wrapComponents: {
+        RequestBody: (Original, system) => (props) => {
+          const React = system.React
+          const contentType = props.contentType || ""
+          const isBinaryMediaType = ["application/octet-stream", "image/", "audio/", "video/"]
```
nflaig: `"image/", "audio/", "video/"` why do we need these?

nflaig (commented):
thanks @syjn99 this is great, did you make sure this doesn't break anything else, that would be my worry when adding such a plugin

index.html
syjn99: Simplified in https://github.com/ethereum/beacon-APIs/pull/632/commits/13b98c803667956025fcefce387c8494feb49cde

syjn99:
> thanks @syjn99 this is great, did you make sure this doesn't break anything else, that would be my worry when adding such a plugin

Yes I believe it is a hacky way but this should not affect other parts of UI as it additionally renders one more component when `RequestBody` is present.

nflaig (approved):
LGTM

PR #633, opened 2026-08-07
Specify SSZ framing for `GET /eth/v1/debug/beacon/data_column_sidecars/{block_id}`

syjn99:
`GET /eth/v1/debug/beacon/data_column_sidecars/{block_id}` declares its SSZ response as:

> SSZ serialized `DataColumnSidecars` bytes.

`DataColumnSidecars` is not an SSZ type, so it might be confusing for encoding the response. Here's a research conducted by Claude:

| Client | Framing | Source |
|---|---|---|
| Lighthouse | `List` | `beacon_node/http_api/src/lib.rs` → `data_columns.as_ssz_bytes()` on `DataColumnSidecarList = Vec<Arc<DataColumnSidecar<E>>>` (offset-framed for variable-size `T`) |
| Nimbus | `List` | `beacon_chain/rpc/rest_debug_api.nim` → `handleDataSidecarRequest[..., List[fulu.DataColumnSidecar, NUMBER_OF_COLUMNS]]`, `sszResponse` on the `List` |
| Lodestar | `List` | `packages/api/src/beacon/routes/debug.ts` → `sszTypesFor(fork).DataColumnSidecars`, i.e. `new ListCompositeType(DataColumnSidecar, NUMBER_OF_COLUMNS)` |
| Grandine | `List` | `http_api/src/standard.rs::debug_beacon_data_column_sidecars` → `DynamicList<Arc<DataColumnSidecar<P>>>`, whose `SszWrite` is `shared::write_list` |
| Teku | concatenation | `.../v1/debug/GetDataColumnSidecars.java` → `data.forEach(s -> s.sszSerialize(out))` |
| Prysm | concatenation | `beacon-chain/rpc/eth/debug/handlers.go::buildDataColumnSidecarsSSZResponse` → loops `MarshalSSZ` and appends |

I can tell Prysm is just concatenating it, so about to fix this.

This PR clarifies and specifies that we are gonna use SSZ List for the response, so the clients can prepend offset bytes before marshalling the items.

### DIFF
--- a/apis/debug/data_column_sidecars.yaml
+++ b/apis/debug/data_column_sidecars.yaml
@@ -50,7 +50,7 @@ get:
                   - $ref: "../../beacon-node-oapi.yaml#/components/schemas/Fulu.DataColumnSidecars"
         application/octet-stream:
           schema:
-            description: "SSZ serialized `DataColumnSidecars` bytes. Use Accept header to choose this response type"
+            description: "SSZ serialized `List[DataColumnSidecar, NUMBER_OF_COLUMNS]` bytes. Use Accept header to choose this response type"
     "400":
       description: "The block ID supplied could not be parsed"
       content:


nflaig:
> DataColumnSidecars is not an SSZ type, so it might be confusing for encoding the response.

there is the following container defined in the spec
```python
class DataColumnSidecars(List[DataColumnSidecar, compute_max_request_data_column_sidecars()]):
    """
    Data column sidecars returned in a ``DataColumnSidecarsByRange`` or
    ``DataColumnSidecarsByRoot`` response.
    """
```

where `compute_max_request_data_column_sidecars` is

```python
def compute_max_request_data_column_sidecars() -> Uint64:
    """
    Return the maximum number of data column sidecars in a single request.
    """
    return Uint64(MAX_REQUEST_BLOCKS_DENEB * NUMBER_OF_COLUMNS)
```

but this type is used by `DataColumnSidecarsByRoot` and `DataColumnSidecarsByRange`, the list limit `MAX_REQUEST_BLOCKS_DENEB * NUMBER_OF_COLUMNS` is not correct for the api response

we do have the schema defined here

https://github.com/ethereum/beacon-APIs/blob/95b9faf52f19fc3b071676ee089c9197475f66b7/types/fulu/data_column_sidecar.yaml#L2-L7

but I do think the change in this PR makes sense to avoid potential confusing about the response type

nflaig (approved):
LGTM, thanks @syjn99

PR #634, opened 2026-08-16
Add `execution_requests_root` to `execution_payload_bid` example

eth2353:
Looks like this was missed in #607


nflaig:
yes, there is much more missing, I have a huge diff locally, just haven't gotten around to push changes

nflaig (approved):

PR #637, opened 2026-08-25
Clarify `Eth-Builder-Url` response-header description

markolazic01:
Clarification of the `Eth-Builder-Url` description. Tightens "echoes" to "MUST echo", in accordance with `blocks.v2.yaml`.


nflaig (approved):
LGTM

PR #638, opened 2026-08-31
post-gloas `payload_attributes` event fields

markolazic01:
## Description
Introduces `safe_block_hash` and `finalized_block_hash` to `payload_attributes`, from Gloas onwards.

### Rationale
This event exists so an external process can call `engine_forkchoiceUpdated`, but that call requires a full `ForkchoiceStateV1`(`headBlockHash`, `safeBlockHash`, `finalizedBlockHash`) and the event currently only provides the head, `parent_block_hash`, which maps to `headBlockHash`.

The safe and finalized execution hashes therefore complete the `ForkchoiceStateV1` object that an external builder passes to `engine_forkchoiceUpdated`. The builder is a second `forkchoiceUpdated` writer on the same execution client as the beacon node; inability to retrieve these fields can result in passing different safe and finalized hashes than the node, which would make the execution client's view flip between the two callers.


markolazic01:
Merged to resolve a conflict

PR #640, opened 2026-09-08
Allow "gloas" version on all applicable endpoints

eth2353:
Vero's Beacon API compliance tests revealed the `"gloas"` (fork) `version` value is not allowed on some Beacon API endpoints where it should be allowed -> this PR fixes that.


eth2353:
PR requires changes to account for EIP-7688 in Gloas -> converting to draft for now

Issue #599, opened 2026-04-27
New event to signal to builder that bid was accepted?


nflaig:
Based on some discussion I had with @rolfyone we might need an event for gloas that allows a CL client that "acts as a builder" to know whether the bid was accepted and included in a block by a proposer without having to subscribe to gossip, or alternatively listen to `block` event and call `getBlockV2` to inspect the contents. A simple fix would be to include the builder index in the `block` event but that requires adding a field, which should be no issue but need to see if people are happy with this.

nflaig:
as discussed with @bharath-123, we might want to extend the existing `block` event to also include the `block_hash` (as committed in the bid) and `builder_index` to make it useful for builders to know whether or not their bid was included by a proposer

rolfyone:
thoughts on this now that we're close to implementation complete?

nflaig:
> thoughts on this now that we're close to implementation complete?

I have a EPF project that is working on `lodestar builder` pretty sure such a event is useful, @pk910 is buildoor still doing that hacky approach of fetching the full block via api?

in general, there could be a bunch of apis that are useful to make a builder work more cleanly, or even have it as a small sidecar, the only way to figure this out properly is to write the software, so I will wait with this until we see what `lodestar builder` needs

I am still debating whether we just wanna extend the `block` even or a separate `bid_included` (or similar) that specifically is emitted if a bid is observed in a beacon block, and does not emit for bids only observed over gossip.

krisoshea-eth:
Hi there, I am working on implementing the interim Builder observation path in Lodestar using the standard `block` event followed by `getBlockV2` for the event's block root.

The flow is sufficient for correctness, but it exposes three API costs that may help resolve this issue:

1. The `block` event contains the beacon block root, slot, and `execution_optimistic`. It does not contain the fork version, selected Builder index, or execution block hash. An external Builder therefore performs one `getBlockV2` request for every newly observed imported post-Gloas block merely to learn whether its bid was selected.
2. Lodestar emits the event after state transition and fork-choice import. Its root lookup checks fork choice for presence, then serves the block from the seen-block input cache or the database; the cache is pruned only after the database write. We therefore do not expect a 404 window on Lodestar. The Beacon API does not specify that event delivery must follow retrievability, and another client may choose a different emission point. A BN-agnostic Builder still needs bounded retry for an event-before-block window.
[...]

PR #641, opened 2026-09-13
Add builder bid identity to the block event

krisoshea-eth:
Related to #599.

This proposes adding `builder_index` and `block_hash` to the existing `block` event from Gloas onwards. Both fields would be required, including for self-builds using `BUILDER_INDEX_SELF_BUILD`. Pre-Gloas events and the existing fields are unchanged.

The Builder currently fetches each imported block to find out whether its bid was included. These fields would let it skip unrelated blocks. If it needs to verify the complete signed bid, it still fetches the block and compares it with its local record.

The trigger remains successful block import, including valid non-head blocks, without waiting for the execution payload envelope. Inclusion is not a guarantee that the block is canonical or a decision to reveal.

## Alternative

A separate lightweight `bid_included` event would leave `block` unchanged and carry `slot`, `block_root`, `block_hash` and `builder_index`, only for externally built blocks. The [alternative patch](https://github.com/krisoshea-eth/lodestar-eip-7732-builder-docs/blob/96d11b9/docs/spec-01/bid-included.patch) and [short comparison](https://github.com/krisoshea-eth/lodestar-eip-7732-builder-docs/blob/96d11b9/docs/spec-01/README.md) are available alongside this proposal.

I'm opening the extended `block` option first to start this discussion, not because the choice is settled. Feedback on the existing topic versus a dedicated event, required self-build fields and client compatibility would be useful. These are alternatives, not two changes intended to merge together.

The relevant Lodestar context is [the block-fetch consumer](https://github.com/ChainSafe/lodestar/pull/9931) and [Marko's extended-event PoC](https://github.com/ChainSafe/lodestar/pull/9854). The latter is implementation exploration, not evidence that this exact contract is supported.

## Validation

Based on Beacon APIs master `ef98d512c03c8ca6b9d7cbdc45b9293ec2b24722`.

- Redocly 1.19.0 lint and swagger-cli 4.0.4 bundling passed for this patch and the alternative separately.
- Additional checks passed for existing event examples, exact new field sets, quoted uint64 values, hashes, the self-build sentinel, topic names and bundled examples.
- `git diff --check` passed.

This does not establish runtime interoperability. The changelog's client-support cells are intentionally empty.


CHANGES.md
```diff
@@ -8,6 +8,7 @@ There are likely to be descriptions etc outside of the list below, but new query
 
 | Endpoint                                                                                                            | [Lighthouse](https://github.com/sigp/lighthouse) | [Lodestar](https://github.com/ChainSafe/lodestar) | [Nimbus](https://github.com/status-im/nimbus-eth2) | [Prysm](https://github.com/prysmaticlabs/prysm) | [Teku](https://github.com/ConsenSys/teku) |
 |---------------------------------------------------------------------------------------------------------------------|--------------------------------------------------|---------------------------------------------------|----------------------------------------------------|-------------------------------------------------|-------------------------------------------|
+| [#641](https://github.com/ethereum/beacon-APIs/pull/641) `block EVENT` includes `builder_index` and `block_hash` from Gloas | | | | | |
```
chong-he: Usually the changes is added at the bottom of the list

apis/eventstream/index.yaml
```diff
@@ -66,10 +66,28 @@ get:
...
+
+                Before Gloas, `builder_index` and `block_hash` are absent. From Gloas onwards, both fields are required and contain
+                `signed_execution_payload_bid.message.builder_index` and `signed_execution_payload_bid.message.block_hash` from the imported block.
+                `builder_index` is a quoted decimal uint64. `block_hash` is the `0x`-prefixed 32-byte execution block hash, not the beacon block root in `block`.
+                Self-built blocks also include both fields, with `builder_index` set to `BUILDER_INDEX_SELF_BUILD` (`"18446744073709551615"`).
+                The existing `slot`, `block`, and `execution_optimistic` fields are unchanged.
               value: |
                 event: block
                 data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false}
+            block_gloas:
```
chong-he: This is adding a new event `block_gloas` in the example list, which does not look like the intention of the PR? From the description which is about adding 2 additional fields to the current event `block`

apis/eventstream/index.yaml
```diff
@@ -66,10 +66,28 @@ get:
...
+              description: A Gloas or later block containing an external Builder's bid was successfully imported.
+              value: |
+                event: block
+                data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false, "builder_index":"42", "block_hash":"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76"}
+            block_gloas_self_build:
```
chong-he: Similar comment as above. Maybe the intention is not to add a new examples in the list, but to provide them in the description itself?

eth2353:
> A Builder needs to know when an imported beacon block includes its bid.

To reveal the payload, right? Wouldn't it be more useful to emit a separate event that instructs the builder to reveal the payload? Basically, when a certain attestation threshold is reached for the block?

(Exact % up for discussion but we'd probably want to default to 40% and perhaps make that threshold configurable).

apis/eventstream/index.yaml
krisoshea-eth: Yes I meant those as example labels, not new events. Both examples still use `event: block`, and the topic list is unchanged. I've moved the Gloas examples into the `block` description so they don't look like separate event types.

apis/eventstream/index.yaml
krisoshea-eth: Moved the self-build example there too. Both Gloas examples now sit under `block`.

CHANGES.md
krisoshea-eth: Thank you, I moved it to the bottom of the Development Version table.

krisoshea-eth:
> > A Builder needs to know when an imported beacon block includes its bid.
> 
> To reveal the payload, right? Wouldn't it be more useful to emit a separate event that instructs the builder to reveal the payload? Basically, when a certain attestation threshold is reached for the block?
> 
> (Exact % up for discussion but we'd probably want to default to 40% and perhaps make that threshold configurable).

I think I get what you mean but please correct me if I'm wrong. You're essentially suggesting that the BN watches the attestation support and tells the Builder when to reveal, rather than just reporting that its bid was included.

Would you expect that to completely replace the inclusion notification for a sidecar Builder? Also, what would the 40% be measured against, and what should happen if that threshold isn't reached before the payload deadline?

eth2353:
> Would you expect that to completely replace the inclusion notification for a sidecar Builder?

Possibly, yes! I am a fan of keeping things minimal, so before considering the specific changes in this PR, I wanted to see if these changes are sufficient for the builder usecase (I'm guessing not fully sufficient?) -> and if not, if there is a way to get "enough" using just one SSE event for this usecase, in which case this specific change would probably be unnecessary.

> Also, what would the 40% be measured against, and what should happen if that threshold isn't reached before the payload deadline?

This one I'm personally less confident on answering. Maybe we can discuss this with more knowledgeable people in the epbs/payload-builders channel?

If the threshold isn't reached before the deadline, I feel like the "payload_should_be_revelead" SSE event could indeed not be fired.

krisoshea-eth:
> > Would you expect that to completely replace the inclusion notification for a sidecar Builder?
> 
> Possibly, yes! I am a fan of keeping things minimal, so before considering the specific changes in this PR, I wanted to see if these changes are sufficient for the builder usecase (I'm guessing not fully sufficient?) -> and if not, if there is a way to get "enough" using just one SSE event for this usecase, in which case this specific change would probably be unnecessary.
> 
> > Also, what would the 40% be measured against, and what should happen if that threshold isn't reached before the payload deadline?
> 
> This one I'm personally less confident on answering. Maybe we can discuss this with more knowledgeable people in the epbs/payload-builders channel?
> 
> If the threshold isn't reached before the deadline, I feel like the "payload_should_be_revelead" SSE event could indeed not be fired.

Ah I get you, thanks for explaining that. On the sufficiency point, the two fields cover the one thing a sidecar builder can't get from the stream today, which is whether an imported block committed to its bid. Right now that costs a block fetch per imported block, inside the window between import and `PAYLOAD_DUE_BPS`. They aren't a reveal decision on their own and I don't think they should be. The honest builder rule in the spec is to reveal when it has seen the block and the block is the timely head of its chain, and to withhold otherwise, and the stream already carries the inputs for that (`head_v2` with `payload_status`, plus the builder's own clock).

On a reveal trigger keyed to attestation weight, I think the idea is sound and it fits why the slot is laid out the way it is i.e. attestations at 25% and the payload at 50%. I think it might be better as a separate event rather than a replacement, for three reasons that are checkable in the spec:

1. Fork choice ignores attestations for the current slot (`validate_on_attestation` requires `get_current_slot(store) >= attestation.data.slot + 1`), so a node can't read a weight for the block off its store during the slot. It would need a new real-time tally of gossip attestations.
2. That tally only covers the subnets the node subscribes to until aggregates arrive, and aggregates are due at 50%, the same moment as the payload. The same threshold would mean different things on different nodes.
3. If the event silently doesn't fire, a builder relying on it doesn't reveal. If the block then reaches the payment quorum on chain (60% of the slot's balance in same-slot attestations), the builder pays anyway. Not revealing into a block that becomes canonical is the most expensive outcome a builder has, so I'd rather the default be to reveal on a timely head and treat weight as an additional input.

On the denominator, the spec's own quorum is total active balance divided by `SLOTS_PER_EPOCH`, so I'd measure any threshold against that. The numerator is the problem, per point 2.

If the concern is touching `block` at all, the `bid_included` alternative linked in the description is a single new event that fires only for externally built blocks and leaves `block` unchanged, so it may be the more minimal shape. I'm happy with either approach really; I opened the extended `block` version first because it mirrors what `execution_payload` already carries.

I'll raise the reveal-trigger question in the epbs/payload-builders channel and link it here if that's ok.

CHANGES.md
```diff
@@ -39,6 +39,7 @@ There are likely to be descriptions etc outside of the list below, but new query
...
 | [#608](https://github.com/ethereum/beacon-APIs/pull/608) `POST /eth/v1/validator/register_validator` deprecated         |                                                  |                                                  |                                                    |                                                 |                                           |
 | [#608](https://github.com/ethereum/beacon-APIs/pull/608) `POST /eth/v1/validator/beacon_committee_subscriptions` updated         |                                                  |                                                  |                                                    |                                                 |                                           |
+| [#641](https://github.com/ethereum/beacon-APIs/pull/641) `block EVENT` includes `builder_index` and `block_hash` from Gloas | | | | | |
```
nflaig: formatting is off here compared to existing columns, we also don't need to add details here, the PR reference should be sufficient

apis/eventstream/index.yaml
```diff
@@ -66,7 +66,29 @@ get:
...
+                ```
+
+                Self-build:
+                ```text
+                event: block
+                data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false, "builder_index":"18446744073709551615", "block_hash":"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76"}
+                ```
               value: |
                 event: block
                 data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false}
```
nflaig: should always reflect the latest fork in the example

apis/eventstream/index.yaml
```diff
@@ -66,7 +66,29 @@ get:
...
+                ```text
+                event: block
+                data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false, "builder_index":"42", "block_hash":"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76"}
+                ```
+
+                Self-build:
+                ```text
+                event: block
+                data: {"slot":"10", "block":"0x9a2fefd2fdb57f74993c7780ea5b9030d2897b615b89f808011ca5aebed54eaf", "execution_optimistic": false, "builder_index":"18446744073709551615", "block_hash":"0x76262e91970d375a19bfe8a867288d7b9cde43c8635f598d93d39d041706fc76"}
+                ```
```
nflaig: imo this is way too much, this change is trivial, we are adding two fields which are established in the consensus-specs already on how they work or to what value they should be set

I prefer keeping it really simple, 1 sentence that says that gloas adds the two new fields should be sufficient

also see https://github.com/ethereum/beacon-APIs/pull/583 and https://github.com/ethereum/beacon-APIs/pull/621 for sse event updates

this one is even easier becasue we are just adding fields, so this should be trivial for clients and consumers to handle

nflaig (commented):
extending `block` event seems the simplest to me and clients/consumers should have no issues with this, we did it multiple times in the past (eg. `attestation` event) and it was quite trivial

PR #642, opened 2026-09-16
Derive default auth data from the builder hostname

nflaig:
Aligns the `data` default with https://github.com/ethereum/builder-specs/pull/168, the builder's hostname instead of the URL bytes verbatim.


JasonVranek (approved):

jimmygchen (approved):

rolfyone (approved):
LGTM

PR #643, opened 2026-09-17
Remove `GET /eth/v1/validator/execution_payload_bids/{slot}/{builder_index}`

### LINKED ISSUE #595: Move endpoint to get unsigned bid to builder namespace

markolazic01:
### Description

This endpoint returns a beacon node constructed unsigned `ExecutionPayloadBid` for the caller to sign. Per new builder design, it no longer has a consumer:
 
* Builders that can produce a payload already have execution client access and construct the bid themselves.
* Self build bids use an empty signature and are assembled during block production, so the validator client never fetches one.

This is why it is being removed rather than moved to `builder/`.

Builder section inside `validator-flow.md` has been updated accordingly.

Closes #595


nflaig (approved):
LGTM, thanks @markolazic01, this hasn't been implemented by clients yet and there is no known consumer for it and no one is planning to utilize this api in the future, builders can construct the unsigned bid outside the beacon node.

